PR #295, opened 2026-02-27
feat: allow users to retrain models and keep track of parent models

MarkRedeman:
# Pull Request

This PR allows users to retrain their models as well as set a custom `max_steps` instead of the hardcoded 10.000 steps.
Still a bit of a draft - requires some more testing. I initially started working on the retrain feature as I was testing inference on XPU.

## Type of Change

- [x] ✨ `feat` - New feature

## Screenshots

image


MarkRedeman:
Good point! I will add it and then hide the batch size + max steps settings under a "Show advanced settings" option. Later once our whole training pipeline matures more we can adopt the same advanced training settings UX that we have in Geti.

application/ui/src/routes/models/model-table.component.tsx
```diff
@@ -62,6 +74,7 @@ export const ModelRow = ({
                         <MoreMenu />
                     </ActionButton>
                     <Menu onAction={onAction}>
+                        <Item key='retrain'>Retrain</Item>
```
RHeckerIntel: Not sure if I like the name retrain. Perhaps fine tune or add more steps?

PR #299, opened 2026-03-02
refactor: improve application logging

MarkRedeman:
# Pull Request

This PR sets up a better default logging configuration based on [Anomalib Studio's settings](https://github.com/open-edge-platform/anomalib/tree/main/application/backend/src/core/logging).
It also updates old code so that we always log via loguru.

Later we can extend this feature by adding new logs api endpoints which expose the logs as SSE responses, which the UI can then expose by a similar modal as [Anomalib Studio does](https://github.com/open-edge-platform/anomalib/pull/3259).

## Type of Change

### DIFF
--- a/application/backend/src/core/logging/__init__.py
+++ b/application/backend/src/core/logging/__init__.py
@@ -1,4 +1,4 @@
-# Copyright (C) 2025 Intel Corporation
+# Copyright (C) 2026 Intel Corporation
 # SPDX-License-Identifier: Apache-2.0
 
 from .log_config import LogConfig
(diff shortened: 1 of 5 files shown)


application/backend/src/core/logging/__init__.py
```diff
@@ -0,0 +1,7 @@
+# Copyright (C) 2025 Intel Corporation
```
maxxgx:
```suggestion
# Copyright (C) 2026 Intel Corporation
```

MarkRedeman: I guess that no one has really bothered with it so far. I will look into adding a CI step that enforces it via our formatters.

PR #305, opened 2026-03-03
ci: add Alembic migration conflict detection to prevent multiple heads

MarkRedeman:
Add a CI job, pre-commit hook, and shell script that assert exactly one Alembic migration head exists. This prevents merging PRs that introduce conflicting down_revisions, which would break 'alembic upgrade head'. Includes MIGRATIONS.md with resolution steps and best practices.

## Type of Change

### DIFF
--- a/application/backend/MIGRATIONS.md
+++ b/application/backend/MIGRATIONS.md
@@ -2,6 +2,13 @@
 
 This document covers how database migrations work in this project, how conflicts arise, and how to prevent and resolve them.
 
+Tl/dr: Run the script below to check if you have migration conflicts.
+
+```bash
+cd application/backend
+bash scripts/check_alembic_heads.sh
+```
+
 ## Overview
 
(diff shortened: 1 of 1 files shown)


application/backend/MIGRATIONS.md
```diff
@@ -0,0 +1,151 @@
...
+ERROR: Multiple head revisions are present; please specify a specific target.
+```
+
+This is called the **"multiple heads"** problem.
+
+## Automated Protection
+
+### CI Check (GitHub Actions)
+
+Every pull request runs a `migration-check` job in the Backend workflow (`.github/workflows/backend.yml`). This job:
```
AlexanderBarabanov: There is no `migration-check` job in `.github/workflows/backend.yml` - could you please clarify logic?

MarkRedeman: Thanks! I forgot to remove that part after I removed it in the previous commit.
Initially I added it as a separate workflow, but after the action failed I realized that having it in the `pre-commit` check would be faster.

PR #306, opened 2026-03-03
refactor: decouple `physicalai.inference` from `torch`

samet-akcay:
## Summary

This PR decouples the `physicalai.inference` from `torch` to make `InferenceModel` a pure numpy interface. The inference path now operates entirely with numpy arrays and removed torch as a runtime dependency for inference workloads. All adapters (including `TorchAdapter`) now share a consistent input contract: `dict[str, np.ndarray]` in, `dict[str, np.ndarray]` out.

## Motivation

The `physicalai.inference` module currently imports torch transitively through various paths, which:
- Increases container image size for inference-only deployments
- Prevents using lightweight inference runtimes without torch
- Blocks splitting InferenceModel into a standalone package

### Core Inference Model
- `InferenceModel.select_action` now accepts `dict[str, np.ndarray]` (observation dict) and returns `np.ndarray` (actions)
- Removed all torch tensor operations from the inference path
- Model now operates purely with numpy arrays end-to-end
- `_prepare_inputs` now filters observation dicts to only the adapter's expected input names, preventing runtime errors from ONNX Runtime / OpenVINO when extra keys are passed

### TorchAdapter — Unified Input Contract
- `TorchAdapter.predict()` now accepts `dict[str, np.ndarray]` (same as all other adapters) instead of `dict[str, Observation]`
- Internally converts numpy inputs to `Observation` via `Observation.from_dict(inputs).to_torch(device)` before calling the policy
- `input_names` now returns granular keys (`["state", "images"]`) derived from `policy.model.sample_input` instead of the opaque `["observation"]` from `extra_export_args`
- This ensures `_prepare_inputs` filtering works correctly for the torch backend

### Inference Worker
- Rewrote `inference_worker.py` to build observations using numpy instead of torch tensors
- No torch imports in the inference execution path

### DIFF
--- a/.github/workflows/backend.yml
+++ b/.github/workflows/backend.yml
@@ -53,9 +53,14 @@ jobs:
         working-directory: application/backend
         run: |
           uv lock --check
-          uv sync --frozen --extra cpu
+          uv sync --frozen --extra cpu --extra tests
 
       - name: Check source code with mypy
         working-directory: application/backend
         run: |
           uv run mypy . --config-file=pyproject.toml
+
+      - name: Run python tests
+        working-directory: application/backend
+        run: |
+          uv run pytest tests
--- a/application/backend/pyproject.toml
+++ b/application/backend/pyproject.toml
@@ -45,6 +45,10 @@ xpu = [
     "pytorch-triton-xpu ; sys_platform == 'linux' or sys_platform == 'win32'",
 ]
 
+tests = [
+    "pytest"
+]
+
 # --- PyTorch index configuration ---
 # torch is a transitive dependency (via lightning, lerobot) that defaults to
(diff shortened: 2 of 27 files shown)


library/src/physicalai/inference/adapters/onnx.py
```diff
@@ -109,6 +109,22 @@ def _get_providers(self) -> list[str]:
...
+        Returns:
+            'cuda' if CUDAExecutionProvider is available, else 'cpu'
+        """
+        try:
+            import onnxruntime as ort  # noqa: PLC0415
+        except ImportError:
+            return "cpu"
+
+        available = ort.get_available_providers()
+        if "CUDAExecutionProvider" in available:
```
sovrasov: The behavior differs from the other adapters: should we always pick GPU if available in other adapters too?

PR #331, opened 2026-03-10
docs: release process documentation

AlexanderBarabanov:
# Pull Request

This PR adds release process documentation.

## Type of Change

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -30,7 +30,7 @@
 The `physical-ai-studio` repository releases the following binary artifacts:
 
 - `library` - Python package (`physicalai-train`) published to PyPI;
-- `application` - container images for different architectures - `physical-ai-studio-xpu`, `physical-ai-studio-cuda`, and `physical-ai-studio-cpu` - pushed to `ghcr.io`.
+- `application` - container images for different architectures pushed to `ghcr.io`.
 
 The following container image names are used:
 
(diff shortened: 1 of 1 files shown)


ivanzati (approved):
Thank you, Alexander!

RELEASE.md
```diff
@@ -0,0 +1,232 @@
...
+### Tagging binary patch releases
+
+For patch release `patch` > `0`.
+
+| Artifact to be released         | Tag format                            | Example |
+| ------------------------------- | ------------------------------------- | ------- |
+| `library` (Python package)      | `<lib-major>.<lib-minor>.<lib-patch>` | `0.1.2` |
+| `application` (container image) | `<app-major>.<app-minor>.<app-patch>` | `0.1.2` |
+
+### `VERSION` files format
```
MarkRedeman: What is the use case of this file if we are using `lib/v0.1.2` and `app/v0.1.2` as tag names for releases, don't they provide enough distinction?

MarkRedeman (commented):
I mainly have some clarification questions, but nothing big to block the pr.

My main criticism of the release plan is that our application + library releases feels a bit awkward, even more so now that we will extract the inference framework to a separate repo/package.
The other thing is that I'd prefer to avoid release branches and instead tag new patches directly on `main` when needed.

AlexanderBarabanov:
This proposal will be reworked based on release-please automation.

PR #343, opened 2026-03-13
feat: enable streaming video encoding

MarkRedeman:
This PR enables [streaming video encoding](https://huggingface.co/docs/lerobot/streaming_video_encoding) when recording new episodes. (from https://github.com/huggingface/lerobot/pull/2974)

I had to make a small tweak to get the `auto` mode working for my system due to having an older `h264_nvenc` driver (required: 13.0, found 12.2). There might be a better approach to this but with this extra check storing new episodes is already significantly faster than before: it now takes less than a second while before it could take 20 seconds or more.

## Type of Change

- [x] ✨ `feat` - New feature

### DIFF
--- a/application/backend/src/internal_datasets/lerobot/lerobot_dataset.py
+++ b/application/backend/src/internal_datasets/lerobot/lerobot_dataset.py
@@ -16,6 +16,7 @@ from lerobot.utils.constants import ACTION, OBS_STR
 from loguru import logger
 
 from internal_datasets.dataset_client import DatasetClient
+from internal_datasets.lerobot.streaming_encoding_settings import StreamingEncodingSettings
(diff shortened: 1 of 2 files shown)


application/backend/src/schemas/teleoperation.py
```diff
@@ -1,14 +1,74 @@
...
+from fractions import Fraction
+
+from pydantic import BaseModel, Field
 
 from .dataset import Dataset
 from .environment import EnvironmentWithRelations
 from .model import Model
 
 
+class StreamingEncodingSettings(BaseModel):
```
samet-akcay: the rest of the pydantic classes have `Config` suffix, while this one has `Settings`. Is this intentional?

Daankrol (dismissed):

PR #353, opened 2026-03-17
feat: VLA export

sovrasov:
# Pull Request

## Description

- `Export` is refactored to consider export of tokenizers: a tokenizer file now is located in the root export directory, alongside with the main model
- Preprocessing and export parameters of particular models were updated to provide compatibility with the new export
- Introduced classes for `Model` and corresponding `ExportModel` mixin to hold enforce declaring general and export-specific properties required for the framework to be funtional

## Type of Change

- [x] ✨ `feat` - New feature

## Examples

```python
from physicalai.data import LeRobotDataModule

if __name__ == "__main__":
    l_dm = LeRobotDataModule(repo_id="lerobot/pusht", train_batch_size=1)
    policy = Pi05(dataset_stats=l_dm.train_dataset.stats).eval()
    policy.to_openvino(f"pi05_ov")
```

## Current coverage

|            | ACT | pi05 | SmolVLA |
|------------|-----|------|---------|
| export ir  |  x   |      |         |
| onnx       |   x  |  x    |  x       |
| openvino   |  x   |  x    |  x       |

`x` here means that a corresponding file can be generated, without any reference to validation of such a file

## Limitations

- pi05 export is dove via onnx, because OV fails to export some of the PaliGemma nn modules
- pi05 export consumes ~40GB of RAM when OV tries to process onnx model
- pi05 tokenizer can not be exported to both ov and onnx for now, which means `hf_tokenizers` would be a dep for inference module
- onnx export for VLA policies takes a noticeable amount of time (up to ~5 min)

### DIFF
--- a/application/backend/src/internal_datasets/dataset_client.py
+++ b/application/backend/src/internal_datasets/dataset_client.py
@@ -10,8 +10,12 @@ if TYPE_CHECKING:
 
 class DatasetClient(ABC):
     type: str
-    exists_on_disk: bool = False
     has_episodes: bool = False
+
+    @property
+    @abstractmethod
+    def exists_on_disk(self) -> bool:
+        """Whether the dataset exists on disk."""
 
     @abstractmethod
     def prepare_for_writing(self) -> None:
--- a/application/backend/src/internal_datasets/lerobot/lerobot_dataset.py
+++ b/application/backend/src/internal_datasets/lerobot/lerobot_dataset.py
@@ -16,6 +16,7 @@ from lerobot.utils.constants import ACTION, OBS_STR
 from loguru import logger
 
 from internal_datasets.dataset_client import DatasetClient
+from internal_datasets.lerobot.streaming_encoding_settings import StreamingEncodingSettings
 from internal_datasets.mutations.recording_mutation import RecordingMutation
 from schemas import Episode, EpisodeInfo, EpisodeVideo
 from settings import get_settings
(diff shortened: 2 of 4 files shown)


library/src/physicalai/policies/act/model.py
```diff
@@ -40,7 +41,7 @@
 log = logging.getLogger(__name__)
 
 
-class ACT(nn.Module, FromConfig):
+class ACT(ExportModel, Model):
```
samet-akcay: do we have to drop FromConfig? Is it intentional or by mistake

library/src/physicalai/policies/act/policy.py
```diff
@@ -17,7 +17,7 @@
 from physicalai.train.utils import reformat_dataset_to_match_policy
 
 
-class ACT(Export, Policy):
+class ACT(ExportPolicy, Policy):
```
samet-akcay: now that I see how it is used, it looks a bit weird to have both ExportPolicy and Policy together.

How about the following: we could have both Policy and ExportPolicy as parent classes, not mixins.

```python
class Policy(L.LightningModule, ABC):
    ...

class ExportablePolicy(Policy, ABC):
    ...
```

Then if a policy is exportable we could have the following:

```python
class ACT(ExportablePolicy):
    ...
```

and a non-exportable one would simply be:

```python
class SomeResearchPolicy(Policy):
    ...
```

And the same goes for models?

```python
class Model(nn.Module, ABC):
    ...

class ExportableModel(Model, ABC):
    @property
    @abstractmethod
    def sample_input(self) -> dict[str, torch.Tensor]:
        ...
```

PR #394, opened 2026-03-26
refactor: introduce per export backend params data types

sovrasov:
# Pull Request

## Description

- Introduce per export backend parameters data type
- Torch inference adapter now supports all torch policies even without export parameters declared. For missing parameters case, the default ones are used. As a result, there's no need to use export mixin for research torch-only models.

## Type of Change

## Examples

Now:
```python
extra_args["openvino"] = OpenVINOExportParameters(
            outputs=["action"],
            compress_to_fp16=False,
            export_tokenizer=True,
            exporter_kwargs={},
            preprocessing_type="smolvla",
        )
```
before:
```python
extra_args["openvino"] = {
            "output": ["action"],
            "compress_to_fp16": False,
            "export_tokenizer": True,
            "exporter_kwargs": {},
            "preprocessing_type": "smolvla",
        }
```

### DIFF
--- a/library/tests/unit/inference/test_adapters.py
+++ b/library/tests/unit/inference/test_adapters.py
@@ -11,6 +11,7 @@ import pytest
 import torch
 
 from physicalai.data.observation import Observation
+from physicalai.export.backends import TorchExportParameters
 from physicalai.export.mixin_policy import ExportBackend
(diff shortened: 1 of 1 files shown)


eugene123tw (approved):
Thanks Vlad!

PR #406, opened 2026-03-27
feat: enable pyrefly and setup a baseline for library

MarkRedeman:
This PR sets up pyrefly prek so that we run it in the CI for our library code. Due to the amount of warnings pyrefly showed I opted to setup a baseline first an consider fixing some of the warnings later.

I've included 462497d1c8fdffd6747896e7c2352b7308d13254 in this PR (and reverted it) as an example for the fixes that I was able to quickly fix by asking OpenCode to fix it. However I'm not confident that these fixes don't contain any regressions, hence I reverted it.

To run it and optionally update the baseline use,

```bash
# Run Pyrefly with the configured baseline
uv run pyrefly check -c pyproject.toml

# Regenerate/update the baseline after intentional typing changes
uv run pyrefly check -c pyproject.toml --baseline="pyrefly-baseline.json" --update-baseline
```

## Type of Change

- [x] ✨ `feat` - New feature
- [x] 🔧 `chore` - Maintenance


maxxgx (dismissed):

AlexanderBarabanov (dismissed):

maxxgx (approved):
Looks good!

AlexanderBarabanov (approved):

PR #420, opened 2026-03-31
chore: bump lerobot to 0.5.1 and fix breaking import changes

samet-akcay:
## Summary

- Bump `lerobot` from `0.4.4` to `0.5.1` across library and application


RHeckerIntel (dismissed):
Approval From pov of backend it seems that most things work.

There might be an inefficiency when loading a dataset where by default it might start the image writers etc, even if the user has no intention to write to the dataset. This issue might be already here. I'll investigate this but at least for this PR it works.

samet-akcay:
> Approval From pov of backend it seems that most things work.
> 
> There might be an inefficiency when loading a dataset where by default it might start the image writers etc, even if the user has no intention to write to the dataset. This issue might be already here. I'll investigate this but at least for this PR it works.

Thanks @RHeckerIntel

RHeckerIntel (approved):

sovrasov (approved):

eugene123tw (approved):

PR #437, opened 2026-04-02
feat: OV inference for pi05

sovrasov:
# Pull Request

## Description

- Adds e2e inference in OV for pi05 model
- Introduces numpy pi05 image/text prompt preprocessor 
- no random noise in pi05 flow matching by default

## Type of Change

- [x] ✨ `feat` - New feature

## Limitations

We're using hf tokenizers here, and gemma tokenizer is gated. One of the ways to overcome this, is to use OV tokenizers


eugene123tw (approved):
Thanks!

PR #443, opened 2026-04-07
feat: enable auto-reload server for development

MarkRedeman:
This PR allows us to run `./run.sh --development` so that the server automatically reloads when making code changes to the server code.
One caveat with this is that currently the server is a bit slow to startup. PR #433 will resolve part of this, but there are more improvements we can make here: I have one branch where I added some aggressive lazy loading of imports and reduced the startup time to 1s-2s. However I need to spend some more time on testing that branch to make sure there are no regressions.

One other thing to take note about is that when using this `--dev` mode we will also restart our training workers. This means that if you have running training jobs and then reload the server this will cancel any training jobs.
This is something we will need to spend a bit more time on. I think it would be better if we could make sure that we don't reload training workers.

## Type of Change

- [x] ✨ `feat` - New feature

### DIFF
--- a/application/backend/run.sh
+++ b/application/backend/run.sh
@@ -12,6 +12,7 @@ set -euo pipefail
 # Usage:
 #   SEED_DB=true ./run.sh       # Seed database before launching server
 #   ./run.sh                    # Run server without seeding
+#   ./run.sh --dev              # Force dev mode (hot reload enabled)
 #
 # Environment variables:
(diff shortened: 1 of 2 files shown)


MarkRedeman:
At the moment enabling auto-reload is not a great experience due to our slow server start up. We should first improve this (by improving our import orders, adding lazy loading and perhaps separating any torch logic from the fastapi process) before continuing with this PR.
I will close this for now.

PR #456, opened 2026-04-09
refactor(robot): update robot interface in framework

maxxgx:
# Pull Request

## Description

Update robot protocol interface to align with robot client. Main changes:
- Added members:
  -  `joint_names` property
  -  `is_connected()` 
- Introduced `RobotObservation` protocol. Robot implementation must define the observation as a class instead of returning generic dicts. This allows for stricter type checking and hopefully easier developer experience in the backend.

`enable_torque` and `disable_torque` methods are not added because it doesn't generalize accross robots SDKs, probably due to safety concerns (the trossen implementation does not allow it). 

Next step is to integrate this SO101 implementation into the backend.

## Type of Change

- [ ] ✨ `feat` - New feature
- [ ] 🐞 `fix` - Bug fix
- [ ] 🧪 `test` - Tests
- [ ] 🔧 `chore` - Maintenance

## Breaking Changes

---

### DIFF
--- a/physicalai/examples/so101/README.md
+++ b/physicalai/examples/so101/README.md
@@ -23,6 +23,9 @@ Use `--offset 0.15` (radians) for larger movements or `--delay 1.0` for more tim
 ## Finding your port
 
 ```bash
-ls /dev/tty.usb*        # macOS
+ls /dev/cu.usb*          # macOS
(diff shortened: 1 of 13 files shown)


MarkRedeman (approved):
Thanks Max! The interface makes sense to me and should make it easier for us to have type safe implementations.

PR #457, opened 2026-04-09
feat(smolvla): add camera rename_map, empty_cameras, and name validation

Daankrol:
## Summary

SmolVLA pretrained models expect standardized camera names (`camera1`=top-down, `camera2`=wrist, `camera3`=side) but user datasets recorded via the application use arbitrary names (e.g. `top`, `wrist`). This mismatch causes silent training degradation — the model processes images in wrong positions without any error or warning.

- **`rename_map`**: Remaps dataset camera keys to pretrained model expectations (e.g. `top`→`camera1`, `wrist`→`camera2`). Applied in preprocessor before image processing and in stats before normalization buffer creation.
- **`empty_cameras`**: Pads missing camera slots with `-1` filled placeholder images and zero masks, matching the Pi0.5 pattern. Allows training with fewer cameras than the pretrained model expects.
- **Camera name validation**: Warns on first batch when dataset camera names don't overlap with expected pretrained names, guiding users to configure `rename_map`.

## Usage

```python
model = SmolVLA(
    rename_map=(("top", "camera1"), ("wrist", "camera2")),
    empty_cameras=1,  # 2 real + 1 placeholder = 3 total
    load_vlm_weights=True,
)
```

Or via CLI:
```bash
physicalai fit --config configs/physicalai/smolvla.yaml \
  --model.rename_map '[[top,camera1],[wrist,camera2]]' \
  --model.empty_cameras 1 \
  --model.load_vlm_weights true
```

## Changes

- `config.py`: Add `rename_map: tuple[tuple[str, str], ...] | None` field
- `preprocessor.py`: Add `_apply_rename_map()`, `_validate_camera_names()`, `rename_stats()`, `SMOLVLA_EXPECTED_CAMERA_NAMES` constant. Refactor `_preprocess_images()` to return lists for empty camera stacking. Add module docstring documenting camera position convention.
- `policy.py`: Thread `rename_map` and `empty_cameras` through `__init__` → config → `_initialize_model`. Apply `rename_stats()` before passing stats to `SmolVLAModel`.
- `model.py`: Accept `empty_cameras` param, account for it in image count validation.
- `test_smolvla.py`: 34 new tests (rename, stats renaming, validation, empty cameras).

## Testing

- 62/62 unit tests pass
- No new Pyright errors

## References

- [SmolVLA camera standardization](https://huggingface.co/blog/smolvla#standardizing-camera-views)
- [SmolVLA paper Section 3.2](https://arxiv.org/abs/2506.01844)
- [LeRobot rename_map docs](https://huggingface.co/docs/lerobot/rename_map)


library/src/physicalai/policies/smolvla/model.py
```diff
@@ -380,10 +383,14 @@ def _preprocess_batch(self, batch: dict[str, torch.Tensor]) -> dict[str, torch.T
...
         all_keys = [key for key in self._dataset_stats if self._dataset_stats[key]["type"] == FeatureType.VISUAL.value]
+        expected_count = len(all_keys) + self._empty_cameras
 
-        if len(all_keys) != batch[IMAGES].shape[0]:
-            msg = f"Some of the image features are missing from the batch. \
-                    (batch: {batch.keys()}) (image_features:{all_keys})"
+        if expected_count != batch[IMAGES].shape[0]:
+            msg = (
+                f"Image count mismatch: expected {expected_count} "
+                f"({len(all_keys)} from stats + {self._empty_cameras} empty), "
```
RHeckerIntel, line 333: Why leave this open as is? I thought smolvla expects 3 cameras. If there are fewer than 3, then we add empty cameras tensors. If there are more we dont need to add them. Why not just hardcode the minimum camera size to 3 and pad accordingly with empty frames?

RHeckerIntel (commented):
Love it.

library/src/physicalai/policies/smolvla/model.py
sovrasov: That only makes sense if we commit to hardocde `lerobot/smolvla_base` as the only source of weights, which even lerobot don't do. Weights loading PR is not here so far, and no experiments have been conducted. Every extra camera significantly affects inference speed

RHeckerIntel: Ah right, ok, I assumed one base indeed. Ok. All good.

library/src/physicalai/policies/smolvla/policy.py
```diff
@@ -209,10 +213,15 @@ def _initialize_model(
             env_action_dim: Environment action dimension.
             dataset_stats: Dataset normalization statistics.
         """
-        from .preprocessor import make_smolvla_preprocessors  # noqa: PLC0415
+        from .preprocessor import make_smolvla_preprocessors, rename_stats  # noqa: PLC0415
+
+        rename_map = dict(self.config.rename_map) if self.config.rename_map else None
+        if rename_map and dataset_stats is not None:
+            dataset_stats = rename_stats(dataset_stats, rename_map)
```
sovrasov: Are images statistics actually used by the preprocessor?

library/src/physicalai/policies/smolvla/preprocessor.py
```diff
@@ -118,17 +169,37 @@ def forward(self, batch: dict[str, Any]) -> dict[str, torch.Tensor]:
...
-        batch = self._preprocess_images(batch)
+        images, img_masks = self._preprocess_images(batch)
+
+        # Append empty cameras as -1-filled placeholder images with zero masks.
+        # This allows training/inference with fewer cameras than the pretrained
+        # model expects (e.g. 2 cameras when the model was trained with 3).
+        if self.empty_cameras > 0 and len(images) > 0:
+            for _ in range(self.empty_cameras):
+                images.append(torch.ones_like(images[-1]) * -1)
+                img_masks.append(torch.zeros_like(img_masks[-1]))
```
sovrasov, line 189: One important limitation is that we can pad only tail cameras, but it's allright so far

library/src/physicalai/policies/smolvla/preprocessor.py
```diff
@@ -118,17 +169,37 @@ def forward(self, batch: dict[str, Any]) -> dict[str, torch.Tensor]:
             A dictionary containing the processed batch with added 'tokenized_prompt'
             and 'tokenized_prompt_mask' tensors, after applying state-action normalization.
         """
+        if self.rename_map:
+            batch = self._apply_rename_map(batch)
```
sovrasov: I don't think we need that much hustle renaming things here and there. The policy itself doesn't consume camera names, it just expects the images to arrive in the right order. 
The input images order is defined by `image_keys = Observation.get_flattened_keys(batch, IMAGES)`
Can we just rearrange this list with respect to order of cameras in the mapping?

library/src/physicalai/policies/smolvla/policy.py
Daankrol: No. image normalization is hardcoded (img * 2.0 - 1.0), and FeatureNormalizeTransform explicitly skips visual features with continue in the factory loop. rename_stats for image keys only serves metadata: expected_camera_names for validation, sample_input key construction, and error messages in _preprocess_batch.

library/src/physicalai/policies/smolvla/preprocessor.py
Daankrol: Agreed, I have only seen the usecase where last camera is missing. Lets keep like this until we see other usescases.

library/src/physicalai/policies/smolvla/preprocessor.py
Daankrol: I see where you are coming from. I wanted to focus on the renaming only but we can just remap this before passing to the model.

I have refactored to a hybrid approach: removed _apply_rename_map() (no more batch key mutation), added _reorder_image_keys() that sorts raw keys by mapped target name for correct camera-to-slot assignment, kept rename_stats() for stats/checkpoint/export metadata, and updated _validate_camera_names() to validate mapped logical names. Note: ResizeSmolVLA (exported inference) has no equivalent ordering logic

library/src/physicalai/policies/smolvla/preprocessor.py
```diff
@@ -118,38 +170,88 @@ def forward(self, batch: dict[str, Any]) -> dict[str, torch.Tensor]:
...
+        Args:
+            batch_img_keys: List of flattened image keys (e.g. ["images.top", "images.wrist"]).
+
+        Returns:
+            Reordered list of image keys sorted by their mapped target names.
+        """
+        if not self.rename_map:
+            return batch_img_keys
+
+        rename_map = self.rename_map
```
sovrasov, line 218: This line is unnecessary, `self.rename_map` is used only once later

library/src/physicalai/policies/smolvla/policy.py
sovrasov: Do you use stats in validation or in sample_input key construction?

Daankrol: Yes, both. SmolVLAModel.sample_input() (model.py L275-284) iterates self._dataset_stats and uses image feature names as keys, so the stat names must match the model's expected camera slots after renaming. It's also used in the factory to derive expected_camera_names for _validate_camera_names().

library/src/physicalai/policies/smolvla/preprocessor.py
```diff
@@ -355,20 +556,31 @@ def make_smolvla_preprocessors(
...
     if stats is not None:
+        image_stat_names: set[str] = set()
         for key, stat in stats.items():
             if ACTION in key:
                 feature_type = FeatureType.ACTION
             elif STATE in key:
                 feature_type = FeatureType.STATE
             else:
+                stat_name = str(stat.get("name", key))
+                base = stat_name.rsplit(".", maxsplit=1)[-1] if "." in stat_name else stat_name
```
sovrasov: This looks like a lot of magic for one line. Is it possible to clean up this?

library/src/physicalai/policies/smolvla/preprocessor.py
```diff
@@ -118,38 +170,87 @@ def forward(self, batch: dict[str, Any]) -> dict[str, torch.Tensor]:
...
+
+        # Append empty cameras as -1-filled placeholder images with zero masks.
+        # This allows training/inference with fewer cameras than the pretrained
+        # model expects (e.g. 2 cameras when the model was trained with 3).
+        if self.empty_cameras > 0 and len(images) > 0:
+            for _ in range(self.empty_cameras):
+                images.append(torch.ones_like(images[-1]) * -1)
+                img_masks.append(torch.zeros_like(img_masks[-1]))
+
+        if images:
```
sovrasov, line 191: Why don't move this and related empty cameras processing back to `_preprocess_images()`? That will maintain high-level flow of the `forward` method

library/src/physicalai/policies/smolvla/preprocessor.py
alfieroddan: Hopefully resolved.

Daankrol:
Closing in favor of https://github.com/open-edge-platform/physical-ai-studio/pull/699

PR #459, opened 2026-04-09
feat: add eval-loss validation via episode-level train/val split

### LINKED ISSUE #458: [Bug]: pi05 multi-gpu training fails on pusht
### Description

Lightning checks if there are unused parameters, but it can be bypassed if picking a corresponding training strategy option

### Steps to Reproduce

```python
from physicalai.train import Trainer
from physicalai.data import LeRobotDataModule
from physicalai.inference import InferenceModel
from physicalai.gyms import PushTGym

from torchvision.transforms import v2

import copy
import torch
import numpy as np

if __name__ == "__main__":

    torch.manual_seed(42)
    np.random.seed(42)
    data_path = "lerobot/pusht"

    dataset_name = data_path.split("/")[-1]
    model_name = f"pi05_{dataset_name}"
    experiment_name = f"pi05_{dataset_name}"

    l_dm = LeRobotDataModule(repo_id=data_path, train_batch_size=2)

    policy = Pi05()

    trainer = Trainer(
        max_steps=100,
        accelerator="gpu",
        devices=2,
        experiment_name=experiment_name,
    )
    trainer.fit(policy, l_dm)
```

### Expected Behavior

Training continues

### Actual Behavior

Training fails at the first iteration

### Component

Library (VLA framework)

### Environment

```markdown

```

### Logs / Error Output

```shell
[rank0]:   File "/home/user/geti-action/library/.venv/lib/python3.12/site-packages/torch/nn/parallel/distributed.py", line 1551, in _pre_forward
[rank0]:     if torch.is_grad_enabled() and self.reducer._rebuild_buckets():
[rank0]:                                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[rank0]: RuntimeError: It looks like your LightningModule has parameters that were not used in producing the loss returned by training_step. If this is intentional, you must enable the detection of unused parameters in DDP, either by setting the string value `strategy='ddp_find_unused_parameters_true'` or by setting the flag in the strategy with `strategy=DDPStrategy(find_unused_parameters=True)`.
sys:1: DeprecationWarning: builtin type swigvarlink has no __module__ attribute
sys:1: DeprecationWarning: builtin type swigvarlink has no __module__ attribute
```

### Additional Context

_No response_

eugene123tw:
# Pull Request

## Description

Adds **eval-loss validation** so training runs can monitor `val/loss` on a held-out episode split — useful when no gym environment is available for evaluation. A configurable fraction of episodes (`val_split`, default 10%) is randomly held out before training begins. Each validation epoch runs a forward pass on these episodes and logs `val/loss`, which `ModelCheckpoint` can monitor for early stopping or best-model saving.

For flow-matching models (Pi05, SmolVLA, Groot), validation uses **action prediction MSE** — the full denoising loop predicts actions, then compares against ground truth. This is deterministic and directly measures action quality, unlike the stochastic flow-matching training loss. For non-flow-matching models (ACT), the default training loss (L1) is already deterministic, so it is reused as-is.

## Type of Change

- [x] ✨ `feat` - New feature
- [x] 🧪 `test` - Tests

## Breaking Changes

None. All new parameters have backwards-compatible defaults (`val_split=0.1`). Existing training scripts and configs work without changes.

---

### Data Layer

- **`LeRobotDataModule`** — New `val_split: float` parameter. Reads total episode count from `meta/info.json` (with HF Hub download fallback), randomly splits episodes (seed=42) into train/val sets, and creates a separate validation dataset.
- **`DataModule` (base)** — New `val_eval_dataset` parameter and `val_dataloader()` that returns an eval-loss DataLoader (batch_size=1) when a validation dataset is provided.
- **`reformat_dataset_to_match_policy`** — Now iterates over both `train_dataset` and `val_eval_dataset` to set `delta_indices`.

### Policy Layer

- **`Model` (base)** — Two new methods:
  - `compute_loss(batch)` — abstract, computes training loss with gradients.
  - `compute_val_loss(batch)` — default falls back to `compute_loss` under `@torch.no_grad()`. Flow-matching models override this with action prediction MSE.
- **`Policy` (base)** — `validation_step` dispatches `Observation` batches to `compute_val_loss` and logs `val/loss`. `Gym` batches continue to use rollout evaluation.
- **Removed** dead `_eval_loss_step` indirection and unused `evaluation_step` from ACT policy.

### Model Implementations

| Model | `compute_loss` | `compute_val_loss` |
|-------|---------------|-------------------|
| **Pi05** | Delegates to `_flow_matching_loss` | Action prediction MSE (full denoising) |
| **SmolVLA** | Extracted from `forward()` inline code | Action prediction MSE (full denoising) |
| **Groot** | Delegates to `forward()`, reformats dict→tuple | Action prediction MSE (full denoising via `get_action`) |
| **Pi0** | Delegates to `_forward_loss` | Inherits default (falls back to `compute_loss`) |
| **ACT** | Extracted from `forward()` inline code | Inherits default (L1 loss is already deterministic) |

### Policy Overrides

- **Pi05, SmolVLA, ACT, Groot** policies override `compute_val_loss` to delegate to `model.compute_val_loss(preprocessed_batch)` — avoids the base class train-mode toggle which would activate dropout (ACT has `dropout=0.1`) or gradient checkpointing.

### Backend Integration

- **`TrainJobPayload`** — New `val_split` field (default 0.1, range [0, 1)).
- **`training_worker`** — Passes `val_split` to datamodule, monitors `val/loss` for checkpointing, enables `check_val_every_n_epoch=1`.

### Tests

- **`test_val_split.py`** (16 tests) — `_read_total_episodes` (local/HF fallback/missing), val split logic (episode partitioning, boundary cases, seed reproducibility, validation), `reformat_dataset_to_match_policy` with val dataset.
- **`test_eval_loss.py`** (5 tests) — Verifies `compute_val_loss` doesn't activate dropout, tests both override and default behavior paths.

## Examples

```python
from physicalai.data import LeRobotDataModule
from physicalai.train import Trainer
from lightning.pytorch.callbacks import ModelCheckpoint

model = Pi05(pretrained_name_or_path="lerobot/pi05_base")
datamodule = LeRobotDataModule(root="/data/my_dataset", val_split=0.1)

trainer = Trainer(
    callbacks=[ModelCheckpoint(monitor="val/loss", mode="min")],
    check_val_every_n_epoch=1,
)
trainer.fit(model, datamodule=datamodule)
```

Disable eval loss (val_split=0):

```python
datamodule = LeRobotDataModule(root="/data/my_dataset", val_split=0.0)
# No val/loss logged, no validation dataloader created
```


sovrasov:
Thanks Eugene! Could you provide an example of train / val loss curves to check how we can interpret these, like detect overfitting, mode improvement saturation, etc. ?

eugene123tw:
@sovrasov I ran a experiment on Arend Jan's collected episodes (pick-and-place) on pi05, and here's the result I have so far:

image

**Summary table from this run:**

| Epoch | train/loss (avg) | val/loss | |
|-------|-----------------|----------|---|
| 0 | 0.1554 | 0.1907 | |
| 1 | 0.1108 | 0.2017 | val jumps UP |
| 2 | 0.0957 | 0.1921 | |
| 3 | 0.0841 | 0.1860 | |
| 4 | 0.0798 | **0.1843** | **best val** |
| 5 | **0.0735** | 0.1864 | val rises again |

**why monitoring `val/loss`:**

1. **train/loss always decreases** — it drops monotonically from 0.155 → 0.074 across all epochs. If you checkpoint on min train/loss, you'd always pick the latest epoch (epoch 5/6), which is the most overfit.

2. **val/loss is non-monotonic** — it peaks at epoch 1 (0.202), recovers, hits its minimum at epoch 4 (0.184), then rises again at epoch 5 (0.186). This inflection point is where the model has the best generalization.

3. **The gap widens** — train/loss keeps dropping while val/loss stagnates or rises. This is the classic overfitting signal. By epoch 5, the train-val gap is ~0.113 (vs ~0.035 at epoch 0).

4. **Checkpointing on `min val/loss` saves epoch 4**, which generalizes better than epoch 5/6 despite having higher train loss.

library/src/physicalai/data/lerobot/datamodule.py
```diff
@@ -147,14 +204,54 @@ def __init__(  # noqa: PLR0913
...
 
+        # Split episodes into train / val based on val_split
+        train_episodes = episodes
+        val_episodes: list[int] | None = None
+        if val_split > 0 and repo_id is not None:
+            total_episodes = _read_total_episodes(repo_id, root)
+            all_episodes = episodes if episodes is not None else list(range(total_episodes))
+            n_val = max(1, int(len(all_episodes) * val_split))
+            # Randomly select val episodes with a fixed seed for reproducibility
+            rng = random.Random(42)  # noqa: S311
```
sovrasov: Aren't we usually relying on external seed?
Now, there's no way for the user to change the split for a particular dataset, it's deterministic. That could be surprising for the users who'd try to set a global `random.seed(seed)` and then see no changes
Since it's a torch data module, maybe we can rely on torch's rng here?

library/src/physicalai/data/datamodules.py
```diff
@@ -219,13 +231,24 @@ def train_dataloader(self) -> DataLoader[Any]:
...
-            DataLoader[Any]: Validation DataLoader with collate function for Gym environments,
-                           or empty DataLoader if no validation dataset is configured.
+            DataLoader[Any]: Validation DataLoader.
         """
+        # Eval-loss mode: return batched Observations
+        if self.val_eval_dataset is not None:
+            return DataLoader(
+                self.val_eval_dataset,
+                num_workers=self.num_workers,
+                batch_size=1,
```
sovrasov: Can batch size be > 1 here?

library/src/physicalai/policies/base/model.py
```diff
@@ -13,8 +17,47 @@ class Model(nn.Module, ABC):
...
+
+        Args:
+            batch: Preprocessed batch dict.
+
+        Returns:
+            Tuple of (loss tensor with grad, dict with at least a ``"loss"`` key).
+        """
+
+    @torch.no_grad()
+    def compute_val_loss(self, batch: dict[str, Any]) -> tuple[torch.Tensor, dict[str, float]]:
```
sovrasov, line 46: This method can be abstract as well, actually the only abstract method we need to enable val loss validation

library/src/physicalai/policies/base/policy.py
```diff
@@ -251,21 +251,51 @@ def evaluate_gym(self, batch: Gym, batch_idx: int, stage: str) -> dict[str, floa
...
+            batch_idx: Index of the batch.
 
         Returns:
-            Dictionary of metrics from the gym rollout evaluation
+            Gym mode: dictionary of rollout metrics.
+            Eval-loss mode: loss tensor.
         """
+        if isinstance(batch, Observation):
+            loss, loss_dict = self.compute_val_loss(batch)
+            self.log("val/loss", loss_dict["loss"], prog_bar=True, on_step=False, on_epoch=True, sync_dist=True)
```
sovrasov, line 271: Doesn't lightning already log validation results?

library/src/physicalai/data/lerobot/datamodule.py
```diff
@@ -147,14 +204,54 @@ def __init__(  # noqa: PLR0913
...
+            val_episodes = sorted(rng.sample(all_episodes, n_val))
+            train_episodes = sorted(ep for ep in all_episodes if ep not in set(val_episodes))
+            logger.info(
```
sovrasov: Will this debug info be printed by default?

library/src/physicalai/policies/base/policy.py
```diff
@@ -251,21 +251,51 @@ def evaluate_gym(self, batch: Gym, batch_idx: int, stage: str) -> dict[str, floa
...
+            return loss
         return self.evaluate_gym(batch, batch_idx, stage="val")
 
+    def compute_val_loss(self, batch: Observation) -> tuple[torch.Tensor, dict[str, float]]:
+        """Compute validation loss on a batch.
+
+        Override this in subclasses to define how validation loss is computed.
```
sovrasov: Given that hint, do we need any implementation here?
The one we have now looks a bit sketchy for a base class

library/src/physicalai/policies/groot/model.py
```diff
@@ -331,6 +331,53 @@ def forward(self, batch: Mapping[str, torch.Tensor]) -> dict[str, torch.Tensor]:
             backbone_outputs = self.backbone(groot_inputs)
             return self.action_head(backbone_outputs, groot_inputs)
 
+    def compute_loss(self, batch: Mapping[str, torch.Tensor]) -> tuple[torch.Tensor, dict[str, float]]:
```
sovrasov, line 353: Is that method used in forward, like it's suggested in the base model?

library/src/physicalai/train/trainer.py
```diff
@@ -160,6 +161,13 @@ def __init__(  # noqa: PLR0913
         if auto_scale_batch_size:
             callbacks.append(BatchSizeFinder(mode="power"))
 
+        # When using auto strategy with multiple GPUs, default to DDP with
```
sovrasov, line 164: Is that a leak from a different PR?

sovrasov:
> @sovrasov I ran a experiment on Arend Jan's collected episodes (pick-and-place) on pi05, and here's the result I have so far:
> 
> image
> **Summary table from this run:**
> 
> Epoch	train/loss (avg)	val/loss	
> 0	0.1554	0.1907	
> 1	0.1108	0.2017	val jumps UP
> 2	0.0957	0.1921	
> 3	0.0841	0.1860	
> 4	0.0798	**0.1843**	**best val**
> 5	**0.0735**	0.1864	val rises again
> **why monitoring `val/loss`:**
> 
> 1. **train/loss always decreases** — it drops monotonically from 0.155 → 0.074 across all epochs. If you checkpoint on min train/loss, you'd always pick the latest epoch (epoch 5/6), which is the most overfit.
> 2. **val/loss is non-monotonic** — it peaks at epoch 1 (0.202), recovers, hits its minimum at epoch 4 (0.184), then rises again at epoch 5 (0.186). This inflection point is where the model has the best generalization.
> 3. **The gap widens** — train/loss keeps dropping while val/loss stagnates or rises. This is the classic overfitting signal. By epoch 5, the train-val gap is ~0.113 (vs ~0.035 at epoch 0).
> 4. **Checkpointing on `min val/loss` saves epoch 4**, which generalizes better than epoch 5/6 despite having higher train loss.

Thanks, not it's more clear. In that experiment, it'd be interesting to see val loss dynamics on epochs 6, 7, 8, as val loss increase at the 5th epoch can be seen as a noise. Overall, that can be a good extra indicator when figuring out training strategy before doing the final shot on the whole dataset

library/src/physicalai/data/lerobot/datamodule.py
eugene123tw: Good point. I've updated the design:

- `val_split_seed` now defaults to `None`, which uses the global `random` module. This means `seed_everything(seed)` will naturally control the split — no surprise hidden seed.
- If the user explicitly passes `val_split_seed=42`, it creates an isolated `random.Random(42)` independent of the global seed — useful if they want a fixed split regardless of training seed.

I chose to keep `random` rather than switching to `torch.Generator` because `seed_everything()` already seeds `random`, `numpy`, and `torch` together, so using the global `random.sample()` respects the user's seed choice without extra coupling. The split is just picking episode indices (a pure Python operation) — no tensor ops involved, so there's no benefit to pulling in torch for this.

library/src/physicalai/data/lerobot/datamodule.py
eugene123tw: logger.info() doesn't show because LeRobotDataModule.__init__() runs before Trainer.fit(), and lightning only configures root logging to INFO level once fit() starts. 

Changed this to logger.warning()

library/src/physicalai/data/datamodules.py
eugene123tw: We could expose `val_batch_size`, but the gym-based validation path requires batch_size=1 (it yields full Gym environment objects, not tensor samples). Exposing a configurable batch size only on the eval-loss path could be confusing since both share `val_dataloader()`.

For now I'm keeping it at 1 for consistency with gym evaluation. 

Happy to make it configurable if we find eval-loss validation is too slow at batch_size=1.

library/src/physicalai/policies/base/model.py
eugene123tw: I'd keep it non-abstract. The default fallback to compute_loss is the right call for most models. Making it abstract would force every subclass to implement it even when the training loss is a fine validation metric. Models that need a different metric (e.g., action MSE after full denoising) can still override it.

library/src/physicalai/policies/base/policy.py
eugene123tw: I don't think so, lightning orchestrates when validation runs, but doesn't automatically log any metrics. We need to explicitly call self.log() inside validation_step() to record val/loss to wandb/tensorboard and the progress bar. The return value of validation_step() alone doesn't get logged anywhere.

library/src/physicalai/policies/groot/model.py
eugene123tw: Good point, no, GrootModel.forward() doesn't call compute_loss(). I added some changes to groot model and policy.

library/src/physicalai/train/trainer.py
eugene123tw: I did this on purpose. This is to fix multi-device training for Pi05. As it's a small change, I think it can be added. Let me know if you think we should make it into a new PR

eugene123tw (commented):
@sovrasov Thanks Vlad, could you take a look when you have time?

library/src/physicalai/policies/base/policy.py
eugene123tw: Fair point, the default toggles training mode mid-validation (might affects dropout, batchnorm, etc.). All our concrete policies (Groot, Pi05) already override this anyway. I'll raise NotImplementedError instead so new policies are forced to define their val loss explicitly.

sovrasov (approved):

PR #461, opened 2026-04-10
refactor: make job a descriminator type

MarkRedeman:
This PR makes preparations for us to enable both dataset and model import jobs.
There are two things this PR does:
1. Expose a new `GET /api/jobs/{job_id}` endpoint that we can use to poll the status of one specific job
2. Change the `Job` model to be a based on a discriminator type, similarly to how the `Camera` model behaves

This will then allow us to have a `DatasetImportJob`, `ModelImportJob` etc.

## Type of Change


RHeckerIntel (approved):
Nice this was on the todo.

PR #462, opened 2026-04-10
feat: SnapFlow self-distillation for SmolVLA and Pi0.5

Daankrol:
## Summary

Implements [SnapFlow](https://arxiv.org/abs/2604.05656) (One-Step Action Generation for Flow-Matching VLAs via Progressive Self-Distillation) for both SmolVLA and Pi0.5 policies. SnapFlow is an additive training mode that teaches the flow-matching denoiser to produce actions in a single step instead of 10, giving significant inference speedup with no architecture changes beyond a zero-initialized target-time MLP.

## Changes

- **SmolVLA** (`model.py`, `config.py`, `policy.py`): target-time MLP, `_predict_velocity` helper, `_forward_fm` extraction, SnapFlow routing in `forward()`, 1-step inference in `sample_actions()`
- **Pi0.5** (`model.py`, `config.py`, `policy.py`): same pattern with bfloat16 local copy safety in `_predict_velocity` and gradient checkpointing compatibility
- **Configs** (`smolvla.yaml`, `pi05.yaml`): commented-out SnapFlow sections for easy enablement
- **Benchmark** (`tests/snapflow_benchmark.py`): training + inference validation for both FM and SnapFlow modes

### New config fields (both policies)

| Field | Default | Description |
|---|---|---|
| `snapflow_enabled` | `False` | Master toggle — off by default, existing behavior unchanged |
| `snapflow_alpha` | `0.5` | Per-sample probability of FM vs consistency loss |
| `snapflow_lambda` | `1.0` | Consistency loss scaling factor |
| `snapflow_num_inference_steps` | `1` | Euler steps at inference (1 = one-shot) |

## Design decisions

- **Zero-init target-time MLP** is the only new parameter → identical behavior when `snapflow_enabled=False`
- **Per-sample Bernoulli mixing**: each batch sample randomly gets FM or consistency loss (matches paper)
- **Prefix caching**: VLM prefix computed once and reused for all 3 expert forward passes in the consistency path
- Target-time embedding added to `time_emb` **before** it becomes the expert conditioning signal

### Inference latency (the main win)

| Config | FM 10-step | SnapFlow 1-step | Speedup |
|---|---|---|---|
| 8 layers, bs=4 | 142 ms | 35 ms | **4.1x** |
| 16 layers, bs=4 | 233 ms | 57 ms | **4.1x** |
| 16 layers, bs=16 | 333 ms | 100 ms | **3.3x** |

Speedup is 3.3–4.1x on A100. Below the theoretical 10x because the VLM prefix forward pass (shared via KV-cache) is a fixed cost that doesn't scale with denoising steps. Larger models (Pi0.5 at 3B+) where the action expert is proportionally heavier will see closer to 10x.

### Training overhead

| Config | FM | SnapFlow | Ratio |
|---|---|---|---|
| 8 layers, bs=4 | 101 ms/step | 198 ms/step | 1.96x |
| 16 layers, bs=4 | 166 ms/step | 336 ms/step | 2.02x |
| 16 layers, bs=16 | 223 ms/step | 362 ms/step | 1.62x |

~1.6–2.0x slower per step (2 no-grad teacher passes + 1 grad student pass for consistency). Ratio improves at larger batch sizes.

### Memory

Virtually identical at small batches (~3.5 GB). At bs=16, SnapFlow uses ~1 GB more (4.6 vs 3.6 GB) due to teacher trajectory tensors.

## Backward compatibility

- All 116 existing unit tests pass (31 SmolVLA + 85 Pi0.5)
- `snapflow_enabled=False` by default — zero behavioral change unless opted in

## References

- Paper: [SnapFlow: One-Step Action Generation for Flow-Matching VLAs via Progressive Self-Distillation](https://arxiv.org/abs/2604.05656) (April 2026)


Daankrol:
Previously both gradient-tracked forward passes (v_fm and v_pred) were computed for the entire batch, roughly doubling activation memory. 
Now samples are split by fm_mask before forward: FM samples only run v_fm, consistency samples only run v_pred + no-grad teacher steps.

This is the same approach as in the paper. The benchmark results posted earlier are now outdated as we use almost the same memory as baseline.

Daankrol:
Local benchmarks on a PTL machine. I go from ~300ms to ~175ms inference latency. More benchmarks on physical robots are needed to confirm model accuracy.

library/src/physicalai/policies/pi05/model.py
```diff
@@ -936,6 +972,80 @@ def time_mlp_func(time_emb: Tensor) -> Tensor:
 
         return embs, pad_masks, att_masks, adarms_cond
 
+    def _predict_velocity(
```
eugene123tw, line 929: Could we share _predict_velorcity between SmolVLA and Pi05?

library/src/physicalai/policies/smolvla/policy.py
```diff
@@ -546,7 +573,7 @@ def compute_val_loss(self, batch: Observation) -> tuple[torch.Tensor, dict[str,
         processed_batch = self._preprocessor(batch.to_dict())
         return self.model.compute_val_loss(processed_batch)
 
-    def configure_optimizers(self) -> dict[str, Any]:
+    def configure_optimizers(self) -> Any:  # noqa: ANN401
```
sovrasov: What is the issue with the warning here? Previously, linter didn't complain, and the method didn't change

library/src/physicalai/policies/smolvla/policy.py
```diff
@@ -580,6 +607,23 @@ def configure_optimizers(self) -> dict[str, Any]:
             },
         }
 
+    def to_onnx(self, *args: Any, **kwargs: Any) -> None:  # noqa: ANN401
```
sovrasov: I believe this is an artifact, as ONNX export for SmolVLA is not even enabled

library/src/physicalai/policies/pi05/model.py
Daankrol: Although they structurally look the same, the have quite some differences that do not make sharing easy and worthwhile.

eugene123tw (approved):

eugene123tw (approved):

PR #465, opened 2026-04-12
refactor(robot): integrate so101 from framework

maxxgx:
# Pull Request

## Description

Changes:
- Use `SO101` implementation from `physicalai` framework via an adapter
- Add torque control during calibration (disable torque)

Note:
- The SO101 adapter converts from radians to degrees. Ideally, we'd want to use radians directly (TODO)
- Once we have trossen implementation in the framework, we need to make the adapter generic so that it supports all robots, without writing custom adapters for each.

## Type of Change

- [ ] ✨ `feat` - New feature
- [ ] 🐞 `fix` - Bug fix
- [ ] 🧪 `test` - Tests
- [ ] 🔧 `chore` - Maintenance

## Breaking Changes

---


application/backend/src/robots/so101/adapter.py
MarkRedeman, line 1: Could we keep this file in the `so101` folder and keep the name as `so101.py` until we do the full migration (including Trossen)? I think that would keep this a bit more organized and make it easier for others to see how to integrate new robots in the existing interface.

application/backend/src/robots/robot_adapter.py
```diff
@@ -0,0 +1,273 @@
...
+    def is_connected(self) -> bool:
+        return self._robot.is_connected()
+
+    # Normalization formulas match lerobot motors_bus.py _normalize/_unnormalize:
+    #   RANGE_M100_100 (body): norm = (((ticks - min) / (max - min)) * 200) - 100
+    #   RANGE_0_100 (gripper): norm = ((ticks - min) / (max - min)) * 100
+    #   If drive_mode == 1, flip sign (M100) or flip to 100-norm (0_100).
+
+    def _radians_to_ticks(self, radians: np.ndarray) -> np.ndarray:
+        cal = self._robot._calibration
```
MarkRedeman: Can we not rely on `self.calibration` here? That would allow to skip the `if cal is None` check since this adapter's interface guarantees the calibration's existence.

application/backend/src/robots/so101/adapter.py
maxxgx: Sure, renamed to `so101/adapter.py`

application/backend/src/robots/robot_adapter.py
maxxgx: I skipped the conversion to ticks and instead do radians -> normalized directly

application/backend/src/robots/so101/adapter.py
```diff
@@ -0,0 +1,231 @@
...
+                bias = -bias
+            elif cal_val.drive_mode == 1 and is_gripper:
+                scale = -scale
+                bias = 100.0 - bias
+
+            self._joint_params[name] = {
+                "scale": scale,
+                "bias": bias,
+                "is_gripper": is_gripper,
+            }
```
RHeckerIntel, line 95: This is rather specific code. I figure this was dealt with in lerobot using the Motor class etc?

RHeckerIntel (changes requested):
Getting errors when I start recording. I'll investigate. 
`Sync read failed with comm result -6`

thanks for the added tests for the existing code.

maxxgx:
> Getting errors when I start recording. I'll investigate. `Sync read failed with comm result -6`
> 
> thanks for the added tests for the existing code.

From `scservo_sdk`:
```python
# Communication Result
COMM_SUCCESS = 0  # tx or rx packet communication success
COMM_PORT_BUSY = -1  # Port is busy (in use)
COMM_TX_FAIL = -2  # Failed transmit instruction packet
COMM_RX_FAIL = -3  # Failed get status packet
COMM_TX_ERROR = -4  # Incorrect instruction packet
COMM_RX_WAITING = -5  # Now recieving status packet
COMM_RX_TIMEOUT = -6  # There is no status packet
COMM_RX_CORRUPT = -7  # Incorrect status packet
COMM_NOT_AVAILABLE = -9  #
```

So it seems it can't read the servo and times out. Does this happen only during recording, or are you also unable to connect to the arm on the robot page?

Thanks for investigating.

RHeckerIntel (approved):
LGTM. The timeout issue was caused by useMutation in the frontend on startEpisode. This causes the frontend to ask for a thumbnail (not sure why, but sure) and that causes the backend to hang.

We should probably move the robot control worker to a process. The downside is that we cannot use the stream reuse feature of framesource.

PR #466, opened 2026-04-13
feat: add training device information to train model dialog

MarkRedeman:
This PR adds a new `GET /api/system/devices/training` endpoint (taken anomalib studio's implementation as a reference). The training devices information is then used to determine if the user has enough VRAM to train amodel. If they don't then we show a warning message.
One open for this implementation is that for iGPU the memory that is reported to be used is the same as the system memory, but it is not realistic that the iGPU has access to all of this.

## Type of Change

- [x] ✨ `feat` - New feature

## Screenshots

| **Enough (V)RAM** | **Not enough VRAM** |
|------------|-----------|
| 192 168 2 101_3000_projects(Full HD)           | image          |

### DIFF
--- a/application/backend/src/api/dependencies.py
+++ b/application/backend/src/api/dependencies.py
@@ -22,6 +22,7 @@ from services.event_processor import EventProcessor
 from services.job_service import JobService
 from services.log_service import LogService
 from services.robot_calibration_service import RobotCalibrationService
+from services.system_service import SystemService
 from settings import get_settings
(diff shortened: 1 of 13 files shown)


RHeckerIntel (dismissed):
From a code POV this looks fine and pretty helpful. I cannot speak too much for the numbers and validity, but I believe this is mirrored from our other applications right?

RHeckerIntel (approved):

PR #468, opened 2026-04-13
chore(ci): add required-check job

AlexanderBarabanov:
# Pull Request

This PR adds required-check jobs to extend branch protection rules with required checks.

## Type of Change

- [x] 🔧 `chore` - Maintenance

### DIFF
--- a/.github/workflows/security-scan.yml
+++ b/.github/workflows/security-scan.yml
@@ -98,7 +98,7 @@ jobs:
           scan_type: "fs"
           scan-scope: all
           severity: "LOW"
-          scanners: "vuln,secret,config"
+          scanners: "secret,config"
           format: "sarif"
           timeout: "15m"
           ignore_unfixed: "false"
--- a/.github/workflows/ui.yml
+++ b/.github/workflows/ui.yml
@@ -8,9 +8,39 @@ on:
   pull_request:
   workflow_dispatch: # run on request (no need for PR)
 
+concurrency:
+  group: ui-${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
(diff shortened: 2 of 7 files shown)


.github/workflows/security-scan.yml
```diff
@@ -98,7 +98,7 @@ jobs:
           scan_type: "fs"
           scan-scope: all
           severity: "LOW"
-          scanners: "vuln,secret,config"
```
AlexanderBarabanov, line 101: To reduce duplication as the same information is already available in Dependabot dashboard.

.github/workflows/ui.yml
```diff
@@ -8,9 +8,39 @@ on:
...
+          persist-credentials: false
+
+      - name: Check paths
+        id: check_paths
+        uses: open-edge-platform/geti-ci/actions/check-paths@3c8d37fc4b106224f9e29d72db4a025b62e07e37 #FIXME
+        with:
+          files_yaml: |
+            test:
+              - application/ui/**
+              - application/backend/**
```
MarkRedeman, line 37: Potentially the UI could also run when library changes are made: if our API expose requests are response objects that directly use a library object, then the OpenAPI spec has a direct dependency on the library and thus we should rerun UI checks when this gets updated.

For now though let's keep it as is, since realistically it's unlikely that this catches regressions based on updates of the library. If we do notice this though then we can consider adding the library here?

MarkRedeman (dismissed):
Thanks Alexander! 

I added a minor comment for the UI checks. Not something that we need to address now but figured it would be good to be aware of.

ivanzati (approved):
Thank you, Alexander!

PR #469, opened 2026-04-14
refactor: Start Model worker processes on init

RHeckerIntel:
ModelWorker, a process, should be spawned at startup. 
The model worker registry follows the robot and camera worker registry pattern in that it keeps track of the available processes and allows for reserving a model worker.

Currently allowing only one model at a time since keeping track of vram and stability is pretty hard.

### DIFF
--- a/application/backend/src/workers/model_worker_registry.py
+++ b/application/backend/src/workers/model_worker_registry.py
@@ -59,9 +59,7 @@ class ModelWorkerRegistry:
         """
         async with self._lock:
             if not self._idle:
-                raise ValueError(
-                    f"No idle model workers available (all {self._max_workers} are busy)"
-                )
+                raise ValueError(f"No idle model workers available (all {self._max_workers} are busy)")
             worker_id = next(iter(self._idle))
             self._idle.discard(worker_id)
             self._busy.add(worker_id)
--- a/application/backend/src/workers/model_worker_registry.py
+++ b/application/backend/src/workers/model_worker_registry.py
@@ -1,0 +1,139 @@
+"""Registry for managing pre-spawned model worker processes."""
+
+import asyncio
+import time
+from multiprocessing.synchronize import Event as EventClass
+from types import TracebackType
+from typing import Self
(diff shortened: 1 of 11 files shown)


application/backend/src/workers/model_worker_registry.py
```diff
@@ -0,0 +1,137 @@
...
+
+    async def acquire(self, model: Model, backend: str) -> tuple[UUID, ModelWorker]:
+        """
+        Assign an idle worker to load the given model.
+
+        Returns (worker_id, worker). Raises ValueError if no idle workers.
+        """
+        async with self._lock:
+            if not self._idle:
+                raise ValueError(f"No idle model workers available (all {self._max_workers} are busy)")
```
MarkRedeman:
```suggestion
                raise ValueError(f"No idle model workers available (all {self._max_workers} workers are busy)")
```

application/backend/src/workers/model_worker_registry.py
```diff
@@ -0,0 +1,137 @@
...
+
+    async def release(self, worker_id: UUID) -> None:
+        """
+        Unload the model from a worker and return it to the idle pool.
+        """
+        async with self._lock:
+            if worker_id not in self._busy:
+                return
+            worker = self._workers.get(worker_id)
+
```
MarkRedeman, line 80: Should all of this below not be done within the lock? (since you keep a reference to `worker` and are modifying its state)

MarkRedeman (dismissed):

application/backend/src/workers/model_worker_registry.py
RHeckerIntel: Perhaps, but wouldnt you then not be able to release two models at the same time?

MarkRedeman: Hmm yes, I guess it depends on how you look at it: I thought the lock was representing the specific worker, but in hindsight it just represents the queue which makes more sense.
Let's leave it as is.

MarkRedeman (approved):

PR #470, opened 2026-04-14
refactor: make query invalidation opt in instead of opt out

MarkRedeman:
This is a follow up of #467 as it updates our query client settings so that query invalidation becomes opt in.
This follows what we also are using now on the other Geti apps.

The first commit of this PR contains the necessary configuration.
The second commit contains changes so that we invalidate specific queries (instead of all) when making mutations.

## Type of Change

### DIFF
--- a/application/ui/src/features/robots/robot-control-provider.tsx
+++ b/application/ui/src/features/robots/robot-control-provider.tsx
@@ -181,6 +181,7 @@ export const RobotControlProvider = (props: useRobotControlProps) => {
             ),
     });
 
+    // TODO: verify invalidation needs for WebSocket-only mutations
     const startEpisode = useMutation({
         meta: { skipInvalidation: true },
         mutationFn: async (task: string) => {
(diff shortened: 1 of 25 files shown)


application/ui/src/features/robots/robot-control-provider.tsx
```diff
@@ -181,6 +181,7 @@ export const RobotControlProvider = (props: useRobotControlProps) => {
             ),
     });
 
+    // TODO: verify invalidation needs for WebSocket-only mutations
```
RHeckerIntel: No invalidation needed during robot control worker. However, if the user has made changes to a dataset then we should probably find some mechanism to invalidate the dataset episodes query once the dataset changes have been persisted. 

Since we use a mutation this means that *currently* if the user goes back to the dataset screen, even on invalidate, it will still show the existing dataset episodes. A bit of a race condition there.

RHeckerIntel (dismissed):

application/ui/src/features/robots/robot-control-provider.tsx
MarkRedeman: Ah thanks for the clarification! I indeed wanted to check if we could use query invalidation here or if we'd need to commit the backend mutation first.
Ideally it would be nice if we could immediately invalidate and get the new episode, but I guess that requires spending some more time on the recording logic.

RHeckerIntel (approved):

PR #471, opened 2026-04-14
feat: add LeRobot-style image augmentations for all policies

Daankrol:
## Summary

- Implement `RandomSubsetApply` and `SharpnessJitter` transforms adapted from LeRobot (Apache-2.0) to enable stochastic image augmentation during training
- Each forward pass randomly selects 3 of 6 transforms (brightness, contrast, saturation, hue, sharpness, affine) via weighted multinomial sampling without replacement
- Apply uniform augmentation config to all model configs (ACT, ACT XPU, SmolVLA, Pi0, Pi0.5, GR00T), matching LeRobot's universal dataset-level augmentation approach

## Design Decisions

- **Pure YAML config approach** (`class_path`/`init_args`) — no new Python dataclass config needed, uses existing LightningCLI instantiation
- **Each color property is a separate ColorJitter instance** so `RandomSubsetApply` can independently select each one — matches LeRobot's design for stochastic subset selection
- **Default values match LeRobot exactly**: brightness (0.8-1.2), contrast (0.8-1.2), saturation (0.5-1.5), hue (-0.05 to 0.05), sharpness (0.5-1.5), affine (5° rotation, 5% translate)


library/src/physicalai/transforms/image_transforms.py
```diff
@@ -0,0 +1,170 @@
...
+from typing import Any
+
+import torch
+from torchvision.transforms.v2 import Transform
+from torchvision.transforms.v2 import functional as F  # noqa: N812
+
+_EXPECTED_SHARPNESS_LEN = 2
+
+
+class RandomSubsetApply(Transform):
```
samet-akcay: Is `RandomSubsetApply` also used by lerobot as the transform name? 
Would it be an idea to have a similar naming to torchvision itself, which already has `RandomChoice` [transform](https://docs.pytorch.org/vision/main/generated/torchvision.transforms.RandomChoice.html)
Maybe something like `RandomChoices` or `MultiRandomChoice` or something better?

library/src/physicalai/transforms/image_transforms.py
```diff
@@ -0,0 +1,170 @@
...
+            inputs = outputs if needs_unpacking else (outputs,)
+
+        return outputs  # type: ignore[possibly-undefined]
+
+    def extra_repr(self) -> str:
+        """Return a string representation of the transform configuration."""
+        return f"transforms={self.transforms}, p={self.p}, n_subset={self.n_subset}, random_order={self.random_order}"
+
+
+class SharpnessJitter(Transform):
```
samet-akcay: Does `SharpnessJitter` imply randomness?

`RandomAdjustSharpness` immediately tell user that it is random. Would be great if this transform could also imply that ?

library/src/physicalai/transforms/image_transforms.py
```diff
@@ -0,0 +1,170 @@
...
+    def make_params(self, flat_inputs: list[Any]) -> dict[str, Any]:  # noqa: ARG002
+        """Sample a random sharpness factor from the configured range.
+
+        Returns:
+            Dictionary containing the sampled ``sharpness_factor``.
+        """
+        sharpness_factor = torch.empty(1).uniform_(self.sharpness[0], self.sharpness[1]).item()
+        return {"sharpness_factor": sharpness_factor}
+
+    def transform(self, inpt: Any, params: dict[str, Any]) -> Any:  # noqa: ANN401
```
sovrasov, line 165: Would it be an idea to add at least smoke tests that call the transforms and check if the output image still has the same shape?

library/configs/physicalai/act.yaml
```diff
@@ -16,17 +16,44 @@ data:
     val_gym:
       class_path: physicalai.gyms.pusht.PushTGym
     image_transforms:
-      class_path: torchvision.transforms.v2.Compose
+      class_path: physicalai.transforms.RandomSubsetApply
```
sovrasov: Image augmentations in lerobot are turned off by default (you could double-check and correct me if I'm wrong). Are there any experiments that suggest we should do the opposite? Maybe create a bunch of standalone configs with augms enabled?

library/src/physicalai/transforms/image_transforms.py
Daankrol: Sure! Renamed it.

library/src/physicalai/transforms/image_transforms.py
Daankrol: Okay, renamed it

library/src/physicalai/transforms/image_transforms.py
Daankrol: done.

library/configs/physicalai/act.yaml
Daankrol: I double checked and it seems that lerobot indeed does not enable transforms by default. For environments like pusht or libero I don't expect a big difference. However for real life environments I can see why these augmentations might help. E.g. training a model during the day and testing at night. 

I currently can't benchmark this.

sovrasov (approved):

library/src/physicalai/transforms/image_transforms.py
```diff
@@ -0,0 +1,171 @@
+# Copyright (C) 2025 Intel Corporation
```
samet-akcay:
```suggestion
# Copyright (C) 2026 Intel Corporation
```

library/src/physicalai/transforms/image_transforms.py
```diff
@@ -0,0 +1,171 @@
...
+from typing import Any
+
+import torch
+from torchvision.transforms.v2 import Transform
+from torchvision.transforms.v2 import functional as F  # noqa: N812
+
+_EXPECTED_SHARPNESS_LEN = 2
+
+
+class RandomChoiceApply(Transform):
```
samet-akcay: On a second thought, I would prefer a drop in replacement as the name would be the same as torchvision and easier to remember?
```suggestion
class RandomChoice(Transform):
```

library/tests/unit/transforms/test_image_transforms.py
```diff
@@ -0,0 +1,109 @@
+# Copyright (C) 2025 Intel Corporation
```
samet-akcay:
```suggestion
# Copyright (C) 2026 Intel Corporation
```

library/configs/physicalai/act.yaml
samet-akcay, line 1: are these transforms intentionally commented out?

AlbertvanHouten (approved):

PR #474, opened 2026-04-14
fix: support non-flat images inputs in np inference preprocessors

sovrasov:
# Pull Request

- a call like `inference_model.select_action(obs.to_dict(flat=False))` could destroy OV inference for pi05 and SmolVLA
- pi05 preprocessor cleans up leftovers when collating several images

## Type of Change

- [x] 🐞 `fix` - Bug fix


library/src/physicalai/inference/preprocessors/pi05.py
```diff
@@ -114,13 +114,24 @@ def _preprocess_images(
...
+            )
+
         images: list[np.ndarray] = []
         masks: list[np.ndarray] = []
 
         max_image_dim = 5
-        for key in img_keys:
-            img: np.ndarray = inputs[key]
+        for image in input_images:
+            img = image
```
eugene123tw, line 134: why not just `for img in input_images`?

eugene123tw (commented):
Thanks Vlad, I have one minor comment

library/src/physicalai/inference/preprocessors/pi05.py
sovrasov: Linter doesn't like modifying the loop variable

eugene123tw (approved):

PR #476, opened 2026-04-15
refactor(robot): implement trossen widowxai in framework

maxxgx:
# Pull Request

## Description

- Move trossen widowxai implementation from backend to `physicalai`
- Merged follower and leader arm implementation into a single class
- Added `widowxai/adapter.py` to the backend, similar to the `SO101` adapter

Next step is to refactor `widowxai/adapter.py` and `so101/adapter.py` into a single generic one that will support all future robots coming to `physicalai`.

## Type of Change

- [ ] ✨ `feat` - New feature
- [ ] 🐞 `fix` - Bug fix
- [ ] 🧪 `test` - Tests
- [ ] 🔧 `chore` - Maintenance

## Breaking Changes

---


application/backend/src/robots/widowxai/adapter.py
```diff
@@ -0,0 +1,190 @@
...
+interface with joint/unit conversion and async-safe hardware access.
+"""
+
+import asyncio
+from typing import Literal
+
+import numpy as np
+from loguru import logger
+from physicalai.robot.trossen import WidowXAI
+from physicalai.robot.trossen.widowxai import WidowXAIObservation
```
samet-akcay: nitpicky, can these two imports be a one liner with relative imports?

physicalai/src/physicalai/robot/trossen/widowxai.py
```diff
@@ -0,0 +1,242 @@
+# Copyright (C) 2026 Intel Corporation
+# SPDX-License-Identifier: Apache-2.0
+# pyright: reportMissingImports=false, reportMissingModuleSource=false
```
samet-akcay: do we check pyright in the ci or tests?

physicalai/src/physicalai/robot/trossen/widowxai.py
```diff
@@ -0,0 +1,242 @@
...
+
+import numpy as np
+import trossen_arm
+from loguru import logger
+
+from physicalai.robot import Robot
+from physicalai.robot.trossen.constants import HOME_POSITION, VALID_ROLES, WIDOWXAI_JOINT_ORDER
+
+if TYPE_CHECKING:
+    from trossen_arm import TrossenArmDriver  # type: ignore[import-not-found]
```
samet-akcay: would an explicit import check avoid this `import-not-found` error ?

physicalai/src/physicalai/robot/__init__.py
```diff
@@ -15,9 +16,8 @@
 from physicalai.robot.interface import Robot
 from physicalai.robot.verify import verify_robot
 
-__all__ = [  # noqa: F822, RUF022
+__all__ = [
     "Robot",
-    "SO101",
```
samet-akcay, line 20: why is this import removed?

physicalai/tests/unit/robot/test_widowxai.py
```diff
@@ -0,0 +1,288 @@
...
+
+    robot = WidowXAI(ip="192.168.1.1", role=role)  # type: ignore[arg-type]
+    robot.connect()
+    return robot
+
+
+# ---------------------------------------------------------------------------
+# Tests: connect lifecycle
+# ---------------------------------------------------------------------------
+
```
samet-akcay, line 82: nitpicky -- do we need separators? You already nicely separate the tests with classes?

samet-akcay (dismissed):
thanks. reviewed it from coding perspective. looks nice, just added some super minor comments.

Once @ArendJanKramer validates and approves, we could merge it

physicalai/src/physicalai/robot/__init__.py
maxxgx: `__getattr__` should make it discoverable already

physicalai/tests/unit/robot/test_widowxai.py
maxxgx: Not really, but I do think it looks better 😄

physicalai/src/physicalai/robot/trossen/widowxai.py
maxxgx: I think I was getting this error because trossen_arm is an optional dependency and it wasn't installed. Pyrefly does not seem to complain anymore. Not sure it CI will fail though

physicalai/src/physicalai/robot/trossen/widowxai.py
maxxgx: Removed

application/backend/src/robots/widowxai/adapter.py
maxxgx: Done

MarkRedeman:
The last commit works for me

MarkRedeman (dismissed):
Let's get this merged. There's some improvements we will need to make to make the dataset recording in studio more stable:
1. Based on Arend Jan's feedback I increased the `goal_time` to be 3x of the default that we use. This should only be needed when recording with low Hz (like our 30Hz default)
2. Sometimes when moving the leader arm quickly we get an error that the joint velocity is too high. AJ shared this diff https://github.com/open-edge-platform/physical-ai-studio/compare/main...arendjan/trossen-limits which I tried applying to this code as well but I rarely still got these errors
3. On the Studio side we no longer apply the forces (`set_external_efforts`), I'll make a follow up PR for that

MarkRedeman (approved):

PR #477, opened 2026-04-15
feat: add compile_model and precision training options with fallback

Daankrol:
## Summary

- Add `compile_model` and `precision` training options across the full stack (backend API, training worker, all policies, UI)
- Native `torch.compile` support for all five policies (ACT, Pi0, Pi05, SmolVLA, Groot) following Pi05's pattern
- ~~Compile fallback mechanism: if compilation fails during training, automatically retry without compile and notify the user via UI toast~~
- Fix `torch.compile` incompatibility with gradient checkpointing on Pi05's vision tower

### Library — Policy compile support
- **ACT** (`config.py`, `model.py`, `policy.py`) — Added `compile_model` option; compiles `forward`
- **Pi0** (`config.py`, `model.py`, `policy.py`) — Added `compile_model` option; compiles `forward` + `sample_actions`
- **SmolVLA** (`config.py`, `model.py`, `policy.py`) — Added `compile_model` option; compiles `forward` + `predict_action_chunk`
- **Groot** (`config.py`, `model.py`, `policy.py`) — Added `compile_model` option; compiles `forward` + `get_action`
- **Pi05** (`model.py`, `config.py`, `policy.py`) — Fix eager attention on vision tower when gradient checkpointing is enabled; default `gradient_checkpointing=True`

### Library — Trainer
- **`trainer.py`** — Add `allow_tf32` and `cudnn_benchmark` flags for Ampere+ GPU speedup

### Backend
- **`schemas/job.py`** — Added `TrainingPrecision` enum, `precision` and `compile_model` fields to `TrainJobPayload`
- **`workers/training_worker.py`** — Pass `precision` to Trainer (`"default"` → `None`), `compile_model` to `setup_policy`; compile fallback with JOB_UPDATE notification on failure
- **`models/utils.py`** — `setup_policy` passes `compile_model` to all policy constructors

### UI
- **`train-model-dialog.tsx`** — Added Precision picker and Compile model checkbox in Advanced settings with info tooltips
- **`index.tsx`** — Toast notification (`notify('info', ...)`) when JOB_UPDATE arrives with a message during running status
- **`openapi-spec.d.ts`** — Updated `TrainJobPayload` type

### Tests
~~**`tests/workers/test_training_worker.py`** — 6 tests covering compile fallback (success, failure, double failure, no-compile passthrough) and precision mapping~~


Daankrol:
# Benchmark Results

## ACT (51.6M params)

| Configuration | Steps | Mean | Median | Warmup | Total | Peak Mem |
| --- | ---: | ---: | ---: | ---: | ---: | ---: |
| `precision=None, compile=False` | 1900 | 292.0 ms | 291.1 ms | 32.9 s | 920.1 s | 17069 MB |
| `precision=32, compile=False` | 1900 | 293.0 ms | 291.8 ms | 29.4 s | 937.2 s | 17065 MB (1.00x) |
| `precision=32, compile=True` | 1900 | 126.3 ms | 124.4 ms | 292.5 s | 1021.8 s | 15713 MB (2.31x) |
| `precision=bf16-mixed, compile=False` | 1900 | 247.0 ms | 245.7 ms | 26.2 s | 793.2 s | 15725 MB (1.18x) |
| `precision=bf16-mixed, compile=True` | 1900 | 107.9 ms | 104.6 ms | 300.6 s | 809.6 s | 11325 MB (2.71x) |

Warmup (`torch.compile`): fp32 = 2925 ms/step, bf16 = 3006 ms/step

## SmolVLA (450M params, 99.9M trainable)

| Configuration | Steps | Mean | Median | Warmup | Total | Peak Mem |
| --- | ---: | ---: | ---: | ---: | ---: | ---: |
| `precision=None, compile=False` | 1900 | 572.0 ms | 571.8 ms | 58.2 s | 1409.9 s | 8608 MB |
| `precision=32, compile=False` | 1900 | 573.7 ms | 573.5 ms | 57.9 s | 1422.4 s | 8608 MB (1.00x) |
| `precision=32, compile=True` | 1900 | 446.7 ms | 446.6 ms | 197.4 s | 1314.6 s | 7418 MB (1.28x) |
| `precision=bf16-mixed, compile=False` | 1900 | 424.3 ms | 423.0 ms | 42.5 s | 1127.5 s | 8059 MB (1.35x) |
| `precision=bf16-mixed, compile=True` | 1900 | 269.1 ms | 269.3 ms | 182.0 s | 974.5 s | 6699 MB (2.13x) |

Warmup (`torch.compile`): fp32 = 1974 ms/step, bf16 = 1820 ms/step

## Pi0 (3.5B params, 578M trainable)

| Configuration | Steps | Mean | Median | Warmup | Total | Peak Mem |
| --- | ---: | ---: | ---: | ---: | ---: | ---: |
| `precision=None, compile=False` | 1900 | 722.4 ms | 722.7 ms | 74.1 s | 1717.6 s | 11721 MB |
| `precision=32, compile=False` | 1900 | 721.6 ms | 721.8 ms | 73.6 s | 1716.9 s | 11718 MB (1.00x) |
| `precision=32, compile=True` | 1900 | 104.9 ms | 104.5 ms | 178.7 s | 731.8 s | 10146 MB (6.89x) |
| `precision=bf16-mixed, compile=False` | 1900 | 742.8 ms | 743.0 ms | 75.9 s | 1758.8 s | 11710 MB (0.97x) |
| `precision=bf16-mixed, compile=True` | 1900 | 102.6 ms | 102.3 ms | 171.5 s | 692.8 s | 10138 MB (7.04x) |

Warmup (`torch.compile`): fp32 = 1787 ms/step, bf16 = 1715 ms/step

## Pi05

| Configuration | Steps | Mean | Median | Warmup | Total | Peak Mem |
| --- | ---: | ---: | ---: | ---: | ---: | ---: |
| `precision=None, compile=False` | 180 | 2992.8 ms | 2992.0 ms | 65.4 s | 800.3 s | 19171 MB |
| `precision=32, compile=False` | 180 | 2996.6 ms | 2995.6 ms | 61.5 s | 797.2 s | 19173 MB (1.00x) |
| `precision=32, compile=True` | 180 | 1477.3 ms | 1477.2 ms | 201.0 s | 699.4 s | 18151 MB (2.03x) |
| `precision=bf16-mixed, compile=False` | 180 | 2963.5 ms | 2940.0 ms | 60.4 s | 814.1 s | 19188 MB (1.01x) |
| `precision=bf16-mixed, compile=True` | 180 | 1413.7 ms | 1414.2 ms | 107.0 s | 587.1 s | 18199 MB (2.12x) |

## GROOT

| Configuration | Steps | Mean | Median | Warmup | Total | Peak Mem |
| --- | ---: | --- | --- | --- | --- | --- |
| `precision=None, compile=False` | FAILED | CUDA OOM (needed 5.04 GiB, only 4.30 GiB free) |  |  |  |  |
| `precision=32, compile=False` | FAILED | CUDA OOM (needed 5.04 GiB, only 4.25 GiB free) |  |  |  |  |
| `precision=32, compile=True` | 1900 | 1106.2 ms | 1104.2 ms | 313.1 s | 3146.4 s | 31325 MB |
| `precision=bf16-mixed, compile=False` | 1900 | 1281.7 ms | 1279.4 ms | 130.8 s | 3241.2 s | 33563 MB |
| `precision=bf16-mixed, compile=True` | 1900 | 1118.1 ms | 1115.7 ms | 249.7 s | 3106.9 s | 31188 MB |

Warmup (`torch.compile`): fp32 = 3131 ms/step, bf16 = 2497 ms/step

## Cross-Policy Summary

| Policy | Best Config | Mean ms/step | Peak Mem | Speedup vs Baseline |
| --- | --- | ---: | ---: | ---: |
| ACT | `bf16-mixed, compile=True` | 107.9 ms | 11325 MB | 2.71x |
| SmolVLA | `bf16-mixed, compile=True` | 269.1 ms | 6699 MB | 2.13x |
| Pi0 | `bf16-mixed, compile=True` | 102.6 ms | 10138 MB | 7.04x |
| Pi05 | `bf16-mixed, compile=True` | 1413.7 ms | 18199 MB | 2.12x |
| GROOT | `fp32, compile=True` | 1106.2 ms | 31325 MB | 1.00x |

application/ui/src/routes/models/train-model-dialog.tsx
```diff
@@ -332,65 +81,156 @@ export const TrainModelDialog = ({ baseModel, close, defaultMaxSteps = 10000 }:
...
+                                    >
+                                        Torch compile
+                                    </Checkbox>
+                                    <ContextualHelp variant='info'>
+                                        <Heading>Torch compile</Heading>
+                                        <Content>
+                                            <Text>
+                                                Compiles the model using torch.compile for faster training.
+                                                First epoch will be slower due to compilation, but subsequent
+                                                epochs will be significantly faster.
```
MarkRedeman: Perhaps we should also warn here that this may require more VRAM? I remember that Vlad was warning about this?

application/ui/src/routes/models/train-model-dialog.tsx
```diff
@@ -332,65 +81,156 @@ export const TrainModelDialog = ({ baseModel, close, defaultMaxSteps = 10000 }:
...
+                                                    usage with minimal impact on model quality. Full precision
+                                                    (32-bit) is slower but may be needed for debugging.
+                                                </Text>
+                                            </Content>
+                                        </ContextualHelp>
+                                    }
+                                >
+                                    <Item key='bf16-mixed'>Mixed precision (bf16)</Item>
+                                    <Item key='16-mixed'>Mixed precision (fp16)</Item>
+                                    <Item key='32'>Full precision (32-bit)</Item>
```
MarkRedeman: Do all the models support these?
I'm wondering if perhaps some of these changes could be made dynamic, e.g. depending on the model that was selected?

Daankrol:
@eugene123tw posting PI05 memory benchmarks here for reference:

Hardware: NVIDIA A100-PCIE-40GB  
Workload: `batch_size=2`, `bf16-mixed`, `5 steps`, `pick_and_place` dataset

## train_expert_only=True (693M trainable params)

| Config | Peak Memory | Status | Avg Step Time |
| --- | ---: | --- | ---: |
| `no_compile + no_grad_ckpt` | 16.46 GB | OK | 0.282 s |
| `compile + no_grad_ckpt` | 15.08 GB | OK | 0.195 s |
| `no_compile + grad_ckpt` | 12.87 GB | OK | 0.408 s |
| `compile + grad_ckpt` | 12.86 GB | OK | 0.142 s |

## train_expert_only=False (4.1B trainable params)

| Config | Peak Memory | Status | Avg Step Time |
| --- | ---: | --- | ---: |
| `no_compile + no_grad_ckpt` | 37.49 GB | OOM at step 2 | - |
| `compile + no_grad_ckpt` | 36.84 GB | OOM at step 1 | - |
| `no_compile + grad_ckpt` | 36.78 GB | OOM | - |
| `compile + grad_ckpt` | 36.98 GB | Completed 5 steps | 0.45 s (90 s compile warmup) |

application/backend/src/models/utils.py
```diff
@@ -35,15 +35,15 @@ def load_inference_model(model: Model, backend: str) -> InferenceModel:
     return InferenceModel(export_dir=export_dir, policy_name=model.policy, backend=backend, device=inference_device)
 
 
-def setup_policy(model: Model) -> Policy:
+def setup_policy(model: Model, *, compile_model: bool = False) -> Policy:
```
MarkRedeman, line 45: Do we also need to apply these in `load_policy` (defined above)?
That function gets used when retraining a previous model.

application/backend/src/workers/training_worker.py
```diff
@@ -158,11 +161,46 @@ async def _train_model(
...
             dispatcher.start()
-            trainer.fit(model=policy, datamodule=l_dm)
+
+            try:
+                trainer.fit(model=policy, datamodule=l_dm)
+            except Exception as fit_exc:
+                if not payload.compile_model:
+                    raise
+
+                logger.warning(f"Training with compile failed: {fit_exc} — retrying without compilation")
```
MarkRedeman: Perhaps change this to `logger.exception(`, if I understand correctly this will then make loguru also print out the full exception, which can be useful for debugging.

application/backend/src/workers/training_worker.py
```diff
@@ -158,11 +161,46 @@ async def _train_model(
...
+                compile_fallback_msg = "Model compilation failed, falling back to non-compiled training."
+                job = await JobService.update_job_status(
+                    job_id=job.id, status=JobStatus.RUNNING, message=compile_fallback_msg
+                )
+                self.queue.put((EventType.JOB_UPDATE, job))
+
+                policy = setup_policy(model, compile_model=False)
+                trainer = Trainer(
```
MarkRedeman: This part feels a bit  icky, i.e. it's the same as the previous step but only with `compile_model=False`, maybe we can extract the construction of `Trainer` into a private method and call this with the new policy setup?
That way if we change the config of the `Trainer` there will still be only 1 place to change it.

application/ui/src/routes/models/train-model-dialog.tsx
```diff
@@ -210,79 +214,127 @@ const TrainingParameters = ({
...
+                width='100%'
+                label='Precision'
+                selectedKey={precision}
+                onSelectionChange={onPrecisionChange}
+                contextualHelp={
+                    <ContextualHelp variant='info'>
+                        <Heading>Training precision</Heading>
+                        <Content>
+                            <Text>
+                                Controls numerical precision during training. BF16 Mixed uses half-precision where safe
```
MarkRedeman, line 333: I like this description, but do wonder now: what does `default` do? Would `32-bit` not be the default?

application/ui/src/routes/models/train-model-dialog.tsx
```diff
@@ -210,79 +214,127 @@ const TrainingParameters = ({
...
+            <Flex direction='column' gap='size-150' width='100%' justifyContent='end'>
+                <Flex direction='row' gap='size-100' alignItems='center'>
+                    <Checkbox isSelected={compileModel} onChange={onCompileModelChange}>
+                        Compile model
+                    </Checkbox>
+                    <ContextualHelp variant='info'>
+                        <Heading>Compile model</Heading>
+                        <Content>
+                            <Text>
+                                Enables torch.compile for all policies. Can significantly speed up training after an
```
MarkRedeman, line 354: Do we want to mention here that this may require more vram?

MarkRedeman (dismissed):

application/ui/src/routes/models/train-model-dialog.tsx
Daankrol: it can increase vram a bit at start but generally lowers vram usage. I don't think we should mention.

application/ui/src/routes/models/train-model-dialog.tsx
Daankrol: default is device specific. On CUDA we use bf16-mixed as this is safe to do so for all models. For XPU we use 32 as training in bf16 is not yet stable for all models. So default is just the automatic safe choice.

application/backend/src/models/utils.py
Daankrol: yeah, good find!

application/ui/src/routes/models/train-model-dialog.tsx
Daankrol: changed to "auto" with clearer description.

library/scripts/benchmark_training_optimizations.py
samet-akcay, line 1: are these benchmarking capabilities not possible with the existing `Benchmark` object?

library/src/physicalai/train/trainer.py
```diff
@@ -68,6 +66,8 @@ def __init__(  # noqa: PLR0913
         # physicalai-specific parameters
         experiment_name: str | None = None,
         auto_scale_batch_size: bool = False,
+        allow_tf32: bool = True,
+        cudnn_benchmark: bool = True,
```
samet-akcay: Does lightning have built-in support for this? If yes, I'd prefer using their native approach before adding our custom args?

library/src/physicalai/policies/act/model.py
```diff
@@ -186,6 +188,10 @@ def __init__(  # noqa: PLR0913
         )
         self._model = _ACT(self._config)
 
+        if compile_model:
+            torch.set_float32_matmul_precision("high")
+            self.forward = torch.compile(self.forward, mode="default")  # type: ignore[method-assign]
+
```
samet-akcay, line 187: I'm checking this link on pytorch lightning where they recommend we should be using `configure_model` hook for this. 
https://lightning.ai/docs/pytorch/stable/advanced/compile.html#apply-torch-compile-in-configure-model

Would it be possible if this is handled by the base `Policy` somehow, or does it require policy specific handle?

library/src/physicalai/train/trainer.py
```diff
@@ -161,6 +165,13 @@ def __init__(  # noqa: PLR0913
         if auto_scale_batch_size:
             callbacks.append(BatchSizeFinder(mode="power"))
 
+        if allow_tf32:
+            torch.backends.cuda.matmul.allow_tf32 = True
```
sovrasov, line 170: This looks a bit cuda-specific. Would that fail on xpu, or would it be silently ignored?

Daankrol: It is cuda specific and these will be silently ingored when XPU is used. This is why I don't think we need a device check here.

library/scripts/benchmark_training_optimizations.py
Daankrol: The existing `Benchmark` class is for inference evaluation; trained policy in a gym env where performance metrics are measured like success rate.

The benchmark script here is for training performance profiling. It measures speed up in wall time, time per step, peak GPU mem etc. 

I don't think they share any overlap. I think that adding this functionality into the Benchmark class will bloat that class. 

Since this script is a one-off dev script, we could decide to remove it but I think it might be handy in the future.

library/src/physicalai/train/trainer.py
Daankrol: Nope, these are PyTorch flags and not handled by Lightning.

library/src/physicalai/policies/act/model.py
Daankrol: I agree that it would be nice to move all compiling logic to the `configure_model` hook. I think that we could even move it to the Policy base class somehow. That will be a substantial change, so I would suggest to do that in another PR.

samet-akcay: Would it be possible to create an issue for this so we don't forget

RHeckerIntel (dismissed):

application/backend/src/models/utils.py
```diff
@@ -11,18 +12,24 @@
...
+        policy = Pi05.load_from_checkpoint(model_path)
+    elif model.policy == "smolvla":
+        policy = SmolVLA.load_from_checkpoint(model_path)
+    else:
+        raise ValueError(f"Policy {model.policy} not implemented.")
+
+    if compile_model:
+        compile_mode = "default"
+        policy.forward = torch.compile(policy.forward, mode=compile_mode)  # type: ignore[method-assign]
+    return policy
```
eugene123tw, line 32: In `load_policy`, `compile_mode` is hardcoded to "default", but pi05 lightning checkpoints already save compile_mode in `hyper_parameters.config`. For example, Pi05 checkpoints store "max-autotune" — hardcoding "default" silently overrides that.

Could we not do this?

`compile_mode = getattr(policy.config, "compile_mode", "default")`

This way each policy can use its saved `compile_mode`, while policies without a compile_mode config fall back to "default".

application/backend/src/schemas/job.py
```diff
@@ -8,6 +9,14 @@
 from schemas.hardware import DeviceType
 
 
+class TrainingPrecision(StrEnum):
+    """Supported training precision modes."""
+
+    DEFAULT = "default"
+    BF16_MIXED = "bf16-mixed"
+    FULL_32 = "32"
```
eugene123tw: `TrainingPrecision` enum values don't align with lightning's precision strings. "default" is ambiguous, "32" is deprecated lightning precision value (it expects "32-true" now), and the naming conventions are inconsistent.

Consider aligning with Lightning's precision naming:

```python
class TrainingPrecision(StrEnum):
    FP32 = "32-true"
    BF16_MIXED = "bf16-mixed"
    BF16_TRUE = "bf16-true"
```

This removes the translation layer, values can be passed directly to Trainer(precision=...) and the member names are self-documenting.

application/backend/src/utils/device.py
```diff
@@ -33,18 +33,23 @@ def get_lightning_strategy(device: str | None = None) -> str:
...
-        return "xpu_single"
 
-    return "auto"
+def get_default_precision(device: str) -> str | None:
+    """Return the default training precision for a given device type.
+
+    Returns a Lightning-compatible precision string, or None to let
+    Lightning choose (which resolves to ``"32-true"``).
+    """
+    return _DEFAULT_PRECISION_BY_DEVICE.get(device)
```
eugene123tw: Per-device precision defaults create hidden behavior as the same training config silently produces different precision on different hardware. A user moving from CUDA to XPU wouldn't realize their precision changed from bf16-mixed to 32 unless they read this utility code.

Since `TrainJobPayload` already has an explicit precision field, consider making that the single source of truth rather than overriding it per-device here. If the user doesn't specify, "32-true" (Lightning's own default) is a safe universal default. Users who want `bf16-mixed` on CUDA can opt in explicitly.

Also, the XPU default uses "32" (legacy format) while the docstring references "32-true" — worth aligning with the non-deprecated format either way.

application/backend/src/workers/training_worker.py
```diff
@@ -142,27 +150,47 @@ async def _train_model(
...
+                    precision=precision,
+                    check_val_every_n_epoch=1,
+                )
+
+            trainer = _create_trainer()
 
             dispatcher.start()
-            trainer.fit(model=policy, datamodule=l_dm)
+
+            try:
```
eugene123tw: Problems with this approach:

* When `base_model` is not None, the first call uses `load_policy(base_model)` (loads checkpoint weights), but the fallback uses setup_policy(model) (creates from scratch). The retry silently discards the pretrained weights and trains from a fresh model.

* The try-catch is too broad. It catches any Exception, not just compilation errors. An OOM, data loading bug, or NCCL error during training would also trigger the fallback, masking the real issue.

* It rebuilds the entire policy and trainer. For Pi05 that means re-downloading/loading ~3B parameters just to retry.

* The job status says "running" with a message about compilation failure, but the model is now training without compile and potentially without pretrained weights. The user has no way to know the output model quality will be different.

Could we not just let it fail? If `torch.compile` doesn't work on the user's hardware, they should know and disable it explicitly rather than getting a silently degraded training run. If a fallback is really wanted, it should at minimum preserve `base_model` path:

`policy = load_policy(base_model, compile_model=False) if base_model else setup_policy(model, compile_model=False)`

library/src/physicalai/train/trainer.py
```diff
@@ -68,6 +66,8 @@ def __init__(  # noqa: PLR0913
         # physicalai-specific parameters
         experiment_name: str | None = None,
         auto_scale_batch_size: bool = False,
+        allow_tf32: bool = True,
```
eugene123tw: `allow_tf32` defaults to True, which silently changes PyTorch's default numerical behavior (PyTorch defaults to `False`). A few concerns:

* CUDA already defaults to bf16-mixed via `get_default_precision`. With bf16-mixed, most matmuls are already in bf16 so TF32 barely helps. Could you share the numbers how much would TF32 help in terms of speed?

* If a user tries to reproduce paper results where the original code didn't set allow_tf32=True, they'll get slightly different numbers with no obvious reason why. This matters for a training studio where users may benchmark against published baselines.

* *This sets torch.backends.cuda.matmul.allow_tf32 process-wide as a side effect of Trainer construction. It affects all subsequent float32 ops in the process, not just training.

Suggestion: Default to False (match PyTorch's default) and let users opt in explicitly. Alternatively, tie it to the precision setting so only enable TF32 automatically when precision is 32-true.

eugene123tw (changes requested):
Thanks Daan! I have a few comments:


1. `load_policy` hardcodes `compile_mode="default"` (utils.py)

Lightning checkpoints already save `compile_mode` in `hyper_parameters.config` (e.g. Pi05 stores `"max-autotune"`). Hardcoding `"default"` silently overrides it. Consider:

```python
compile_mode = getattr(policy.config, "compile_mode", "default")
```

2. `TrainingPrecision` enum naming (job.py)

Values don't align with Lightning's precision strings. `"default"` is ambiguous, `"32"` is the deprecated format. Consider aligning with Lightning's naming:

```python
class TrainingPrecision(StrEnum):
    FP32 = "32-true"
    BF16_MIXED = "bf16-mixed"
    BF16_TRUE = "bf16-true"
```

Values can then be passed directly to `Trainer(precision=...)` with no translation.

3. Per-device precision defaults (device.py)

Silently produces different precision on different hardware — CUDA gets `bf16-mixed`, XPU gets `32`. Since `TrainJobPayload` already has an explicit `precision` field, consider making that the single source of truth. Users who want `bf16-mixed` should opt in explicitly. 

4. Compile fallback catches all exceptions (training_worker.py)

The fallback catches bare `Exception`, so an OOM, data loading bug, or NCCL error also triggers the "compilation failed" retry path. Additionally, the fallback uses `setup_policy(model)` instead of `load_policy(base_model)`, silently discarding pretrained weights when fine-tuning. Consider either catching only compile-specific exceptions, or at minimum preserving the base model path in the retry.

5. `allow_tf32` defaults to `True` (trainer.py)

This overrides PyTorch's default (`False`) and mutates global state (`torch.backends.cuda.matmul.allow_tf32`) as a side effect of Trainer construction. With the current CUDA default of `bf16-mixed`, TF32 may provides minimal speedup since most matmuls are already in bf16. It mainly benefits `32-true` (~2-3x speedup). Defaulting to `False` matches PyTorch, avoids reproducibility surprises, and lets users opt in when they need speed. Alternatively, only enable TF32 automatically when precision is `32-true`.

application/backend/src/models/utils.py
Daankrol: agreed, much better approach. Updated.

application/backend/src/schemas/job.py
Daankrol: indeed cleaner approach. "default" was only explained in the UI. Updated to use lightning defaults.

application/backend/src/workers/training_worker.py
Daankrol: I see your issues with this approach. While the retry approach can be nice from a UX standpoint it also introduces some issues. We can also simplify by just notifying the user that compile fails and recommending to retry without compiling. I will remove the fallback behavior.

application/backend/src/utils/device.py
Daankrol: I think that per device precision is nice to have since XPU does yet have full stable support for bf16. The issue is indeed that we do this silently. I have removed this logic and replaced it with something simpler. In the UI, in the training popup, we recommend a precision based on the device (bf16 mixed for Cuda, 32 else). There is no default mode anymore and the user has to select a value. I have also updated the precision values to lightning defaults.

Daankrol: image

eugene123tw (approved):

Issue #482, opened 2026-04-15
[Bug]: SmolVLA expert layer condition inverted


alfieroddan:
### Description

https://github.com/open-edge-platform/physical-ai-studio/blob/475a2af96e5559376647fede22bc050a65c7ece0/library/src/physicalai/policies/smolvla/model.py#L1353-L1359

The condition is inverted. It raises when `len(layers)` is divisible by `num_expert_layers` which is the valid case. When the config is invalid (not divisible), the condition is false, the assignment is skipped, num_hidden_layers is set incorrectly, and no error is raised.

Should be:

```python
if num_expert_layers > 0:
    if len(self.get_vlm_model().text_model.layers) % num_expert_layers != 0:
        raise RuntimeError(
            f"VLM has {len(self.get_vlm_model().text_model.layers)} layers, "
            f"not divisible by num_expert_layers={num_expert_layers}"
        )
    lm_expert_config.num_hidden_layers = num_expert_layers
```

### Steps to Reproduce

-

### Expected Behavior

-

### Actual Behavior

-

### Component

Library (VLA framework)

### Environment

```markdown
-
```

### Logs / Error Output

```shell

```

PR #487, opened 2026-04-15
fix: SmolVLA condition inverted for expert VLM layers

alfieroddan:
# Pull Request

## Description

The condition is inverted. It raises when len(layers) is divisible by num_expert_layers which is the valid case. When the config is invalid (not divisible), the condition is false, the assignment is skipped, num_hidden_layers is set incorrectly, and no error is raised.

## Type of Change

- [ ] ✨ `feat` - New feature
- [x] 🐞 `fix` - Bug fix
- [ ] 🧪 `test` - Tests
- [ ] 🔧 `chore` - Maintenance

## Related Issues

#482 

## Breaking Changes

---


Daankrol (approved):
Good find!

sovrasov (approved):

Issue #483, opened 2026-04-15
[Bug]: SmolVLA pad language differs between API and config


alfieroddan:
### Description

https://github.com/open-edge-platform/physical-ai-studio/blob/475a2af96e5559376647fede22bc050a65c7ece0/library/src/physicalai/policies/smolvla/config.py#L115 
defaults to "max_length". 

https://github.com/open-edge-platform/physical-ai-studio/blob/475a2af96e5559376647fede22bc050a65c7ece0/library/src/physicalai/policies/smolvla/policy.py#L109
defaults to "longest". YAML constructs SmolVLAConfig directly uses "max_length". API training and all inference constructs via SmolVLA.__init__ uses "longest". 

### Steps to Reproduce

-

### Expected Behavior

-

### Actual Behavior

-

### Component

Library (VLA framework)

### Environment

```markdown
-
```

### Logs / Error Output

```shell

```

PR #488, opened 2026-04-15
fix: SmolVLA pad language differs between API and config

alfieroddan:
# Pull Request

## Description

Unify default arg of `pad_language_length` between config and api. Config defaults to "longest". API training and all inference constructs via SmolVLA.init uses "longest".

## Type of Change

- [ ] ✨ `feat` - New feature
- [x] 🐞 `fix` - Bug fix
- [ ] 🧪 `test` - Tests
- [ ] 🔧 `chore` - Maintenance

## Related Issues

#483 

## Breaking Changes

---

### DIFF
--- a/library/src/physicalai/policies/smolvla/config.py
+++ b/library/src/physicalai/policies/smolvla/config.py
@@ -112,6 +112,8 @@ class SmolVLAConfig(Config):
 
     prefix_length: int = -1
 
+    # NOTE max_length is fixed for export compatibility (avoids dynamic input shapes).
+    # It should not impact performance since masking ignores unused tokens.
     pad_language_to: str = "max_length"  # "longest"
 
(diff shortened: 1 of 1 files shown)


samet-akcay (dismissed):

sovrasov (changes requested):
max_length is set for compatibility with export to avoid dynamic shapes in inputs. That should not affect model performance ideally, as there's tokens mask that drops unused indices anyway

alfieroddan:
> max_length is set for compatibility with export to avoid dynamic shapes in inputs. That should not affect model performance ideally, as there's tokens mask that drops unused indices anyway

@sovrasov Makes sense! I've added a note. Probably never affected performance anyway was just combing through.

sovrasov (approved):

PR #490, opened 2026-04-16
fix: restore delta_timestamps callback flow for lerobot 0.5.1

samet-akcay:
## Summary

- Fix `_LeRobotDatasetAdapter.delta_indices` getter/setter to target `DatasetReader` instead of the (now non-existent) dataset-level attribute in lerobot >=0.5.1

## Problem

In lerobot 0.5.1, `delta_indices` moved from `LeRobotDataset` to `DatasetReader` as part of the dataset reader/writer split. The adapter's getter/setter was writing to a dangling attribute on the dataset object, so the `PolicyDatasetInteraction` callback (`on_fit_start`) had no effect. 

## Solution

The adapter layer absorbs the lerobot API change — the getter/setter now targets `reader.delta_indices` directly. This restores the original flow:

1. User creates `LeRobotDataModule(repo_id=...)` — **no delta_timestamps needed**
2. User creates policy
3. `trainer.fit()` triggers `PolicyDatasetInteraction` callback
4. Callback reads delta indices from policy and sets them on the dataset via the adapter

### DIFF
--- a/library/src/physicalai/data/lerobot/dataset.py
+++ b/library/src/physicalai/data/lerobot/dataset.py
@@ -227,17 +227,19 @@ class _LeRobotDatasetAdapter(Dataset):
     @property
     def delta_indices(self) -> dict[str, list[int]]:
         """Expose delta_indices from the DatasetReader (lerobot >=0.5.1)."""
-        reader = self._lerobot_dataset.reader
-        if reader is None:
-            return {}
-        return reader.delta_indices or {}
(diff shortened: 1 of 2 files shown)


sovrasov (approved):
Thanks for a quick fix. I need to double-check if our formula for delta timestamps is still aligned with lerobots new `resolve_delta_timestamps()`, but it's outside of the scope of this PR

PR #493, opened 2026-04-16
fix: restore PI0 and GROOT compatibility with transformers 5.3

samet-akcay:
## Summary
- restore PI0 compatibility with transformers 5.3 by updating PaliGemma component access to the nested `model` layout and fixing the suffix attention mask used by the action expert
- restore GROOT compatibility with transformers 5.3 by routing `return_tensors` through `text_kwargs` and `images_kwargs` so the vendored Eagle processor returns tensors without runtime monkey-patching
- validate the hotfix with targeted unit and first-party integration coverage for both PI0 and GROOT


Daankrol (approved):
Good fix! Tested locally to confirm.

PR #494, opened 2026-04-16
refactor: set model name based on policy

MarkRedeman:
This PR removes the name field from the train model dialog in favor of always using the selected policy's name.
Later we can consider simplifying the backend and removing the name from our database.

## Type of Change

image


Daankrol (approved):
LGTM!

PR #499, opened 2026-04-16
fix: show error message on identify robot error

### LINKED ISSUE #496: Identify robot does not work if user has no permission to access it
When the user is not in the `dialout` group or has not set the correct permissions on the robot's usb location then the identify button won't work.
While we can't easily fix this we should at the very least give them a clear error message why it does not work. Similarly to the message added in #453

MarkRedeman:
This PR catches the permission error that the backend raises when it's unable to write to (and therefore connect0 a SO101 robot.
Based on this error we then show a inline alert to the user informing them how to possibly fix this.

## Type of Change

- [x] 🐞 `fix` - Bug fix
## Related Issues

Fixes #496

## Screenshots

image


RHeckerIntel (approved):
Nice

PR #500, opened 2026-04-17
fix: fix unable to load episode thumbnail after lerobot update

MarkRedeman:
After updating to lerboto v0.5.1 the studio's thumbnails broke due to the endpoint raising,
```
RuntimeError: Cannot read from a dataset that is being recorded. Call finalize() first, then access items.
````

This is because Lerobot's dataset refactor blocks `LeRobotDataset.__getitem__` while a writer is active, which made thumbnail generation fail on dataset frame access.

Read thumbnail frames through the dataset reader directly so thumbnail extraction still works without requiring `finalize()`.

## Type of Change

- [x] 🐞 `fix` - Bug fix

### DIFF
--- a/application/backend/src/internal_datasets/lerobot/lerobot_dataset.py
+++ b/application/backend/src/internal_datasets/lerobot/lerobot_dataset.py
@@ -366,7 +366,10 @@ class InternalLeRobotDataset(DatasetClient):
 
         from_idx = int(episode["dataset_from_index"])
         try:
-            image = self._dataset[from_idx][image_key].permute(1, 2, 0).detach().numpy()
+            reader = self._dataset._ensure_reader()
+            if reader.hf_dataset is None:
+                reader.load_and_activate()
+            image = reader.get_item(from_idx)[image_key].permute(1, 2, 0).detach().numpy()
         except Exception:
             return None
 

### CODE REVIEW

File: application/backend/src/internal_datasets/lerobot/lerobot_dataset.py [hunk 1]
```diff
@@ -366,7 +366,10 @@ def _build_thumbnail_png_bytes(self, episode: dict, image_key: str, width: int,
 
         from_idx = int(episode["dataset_from_index"])
         try:
-            image = self._dataset[from_idx][image_key].permute(1, 2, 0).detach().numpy()
```
maxxgx, line 369: `_build_thumbnail() @ L389` also does the same read as here, shouldn't that be updated as well?

Maybe a little helper function could help:
```python
def _read_frame(self, idx: int) -> dict:
    reader = self._dataset._ensure_reader()
    if reader.hf_dataset is None:
        reader.load_and_activate()
    return reader.get_item(idx)
```

MarkRedeman: Yes we should also update that one, however when I checked it seemed like that method isn't being called anywhere. So probably we can just remove it?

After this PR I plan to do a follow up PR with some design improvements based on the changes from lerobot 0.5.1 (making reading and writing to/from datasets more explicit). 
I expect this should make the app a bit more robust, but need to spend a bit more time on testing this before making a PR.

maxxgx (approved):

ArendJanKramer (approved):

PR #502, opened 2026-04-17
fix: invalidate episodes query after closing recording

MarkRedeman:
This makes sure that we refetch the dataset episodes after finishing a recording. If we don't do this and navigate back to the previous page then the user would not be able to see their newly recorded episodes.

## Type of Change

- [x] 🐞 `fix` - Bug fix

### DIFF
--- a/application/ui/src/features/robots/robot-control-provider.tsx
+++ b/application/ui/src/features/robots/robot-control-provider.tsx
@@ -1,10 +1,11 @@
 import { createContext, ReactNode, RefObject, useContext, useRef, useState } from 'react';
 
-import { useMutation, UseMutationResult } from '@tanstack/react-query';
+import { useMutation, UseMutationResult, useQueryClient } from '@tanstack/react-query';
 
 import { fetchClient } from '../../api/client';
 import { SchemaDatasetOutput, SchemaEnvironmentWithRelations, SchemaModel } from '../../api/openapi-spec';
 import useWebSocketWithResponse from '../../components/websockets/use-websocket-with-response';
+import { useDatasetId } from '../datasets/use-dataset';
 
 type FollowerSource = 'teleoperation' | 'model' | null;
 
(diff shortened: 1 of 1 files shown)


application/ui/src/features/robots/robot-control-provider.tsx
```diff
@@ -101,13 +119,18 @@ export const RobotControlProvider = (props: useRobotControlProps) => {
...
         {
             shouldReconnect: () => true,
             onMessage: (event: WebSocketEventMap['message']) => onMessage(event),
             onError: console.error,
-            onClose: () => setState(createRobotControlState()),
+            onClose: () => {
+                invalidateEpisodesQuery();
+
+                setState(createRobotControlState());
+            },
```
RHeckerIntel, line 133: Does this get triggered correctly and is it losing the race condition with the mutation? Say we press the back button (in the header, not the browser button). The user goes back, the socket gets disconnected and the onClose gets triggered. On the backend the api/record/robot_control websocket gets the boot and tells process to disconnect. It joins with 10 second timeout.

If I read this correctly we're flipping a coin on the order of dispatching of process "thread" and the new episodes request.


This also lays bare another issue. We should probably handle the saving and persisting more gracefully. Give the user a choice to persist them and then let the user wait a bit until they are and then go back and show them?

RHeckerIntel (approved):
Looks like this might work most of the times in our current situation. But check the long comment on future work.

application/ui/src/features/robots/robot-control-provider.tsx
MarkRedeman: Yep that was also my concern. I tried this a few times while recording some fake episodes but never saw that race condition take in effect.

> This also lays bare another issue. We should probably handle the saving and persisting more gracefully. Give the user a choice to persist them and then let the user wait a bit until they are and then go back and show them?

Yes I think that would allow us to handle this on a less-bug-prone way. Let's discuss this later?

Issue #504, opened 2026-04-17
Unable to delete dataset if it was used to train a model


MarkRedeman:
If a user used a dataset to train a model and then wants to delete the original dataset then the "Delete dataset" menu option always fails, without showing any information on why it did not work.

There are a few steps to fixing this bug:
1. Make sure to delete snapshots when deleting a model as otherwise there will be a "stale foreign key"
2. Show a proper error / warning message to the user when deleting a dataset that has associated model(s)
3. Allow to remove datasets without deleting its associated models

I think most users would expect that they can safely remove their dataset without requiring to remove any of their models.
Another easy fix could therefore be to remove the foreign key constraint and/or make the `dataset_id` and `snapshot_id` on a model be nullable (and then update any associated model/snapshot so that we make the dataset id null before deleting the actual dataset). 
This will also be required for importing models so that's probably an ok decision.

Long term we will need to come up with a plan on how important it is for us to guarantee data consistency, i.e. if the user has trained a model, do we then want to guarantee that they can see the data that was used to train it.
Snapshots are probably a good way to solve this as well but atm they still directly rely on the original dataset.

PR #505, opened 2026-04-17
fix: delete model snapshot during model removal

MarkRedeman:
This PR fixes an issue where deleting a model wasn't deleting its associated dataset snapshot. Now this isn't a huge deal since it's only a single row in the database, but it prevents the user from deleting their datasets if they had any trained model on it (due to foreign key constraints).

This is an initial fix for #504, as without it users will never be able to delete datasets if they used it to train a model.

## Type of Change

- [x] 🐞 `fix` - Bug fix

### DIFF
--- a/application/backend/src/services/model_service.py
+++ b/application/backend/src/services/model_service.py
@@ -4,7 +4,7 @@ from uuid import UUID
 
 from db import get_async_db_session_ctx
 from exceptions import ResourceNotFoundError, ResourceType
-from repositories import ModelRepository
+from repositories import ModelRepository, SnapshotRepository
 from schemas import Model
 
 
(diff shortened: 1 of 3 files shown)


application/backend/src/services/model_service.py
```diff
@@ -40,10 +40,17 @@ async def update_model(model: Model, update: dict) -> Model:
...
+            snapshot_repo = SnapshotRepository(session)
+
+            await model_repo.delete_by_id(model.id)
+
+            # Remove the associated snapshot row to avoid stale FK references
+            if model.snapshot_id is not None:
+                await snapshot_repo.delete_by_id(model.snapshot_id)
+
+        model_path = Path(model.path).expanduser()
+        shutil.rmtree(model_path)
```
RHeckerIntel, line 53: Since we store snapshot in model path it should be fine for now. But in future we might want to split this up.

RHeckerIntel (approved):

application/backend/src/services/model_service.py
MarkRedeman: Yeah I guess perhaps in the future we could also opt for sharing snapshots (at least the data?) across multiple models? That would complicate stuff but make data storage less heavy.

PR #511, opened 2026-04-17
fix: ModelWorker teardown stalls shutdown.

RHeckerIntel:
ModelRegistry was being requested to shutdown without giving the mp_stop_event. This meant that the shutdown hit the timeout limit.

The scheduler already has stop functionality and shuts down gracefully.
A cleanup of the registry might be nice to simplify most of it.
Unsure if this is the right call. The ModelWorkerRegistry might be a
wrong pattern and a simpler solution in the scheduler might be warranted.


application/backend/src/core/lifecycle.py
```diff
@@ -61,9 +61,6 @@ async def lifespan(app: FastAPI) -> AsyncGenerator[None]:
     robot_registry: RobotWorkerRegistry = app.state.robot_registry
     await robot_registry.shutdown_all()
 
-    model_registry: ModelWorkerRegistry = app.state.model_registry
```
MarkRedeman, line 64: What about moving the `app_scheduler.shutdown()` that is called below above the registry shut downs?

I guess though the main issue is that we are shutting things down in two places?
Do we need the model worker to be aware of the scheduler's shut down events?

RHeckerIntel: > I guess though the main issue is that we are shutting things down in two places?
Indeed exactly.

> Do we need the model worker to be aware of the scheduler's shut down events?
The mp_stop_event is the global close event for all processes. Currently I'm not sure how we should approach this. I think we need to streamline this to one shutdown location. As for how, not sure yet.

MarkRedeman (approved):

PR #515, opened 2026-04-20
fix: Policy selection triggeres form submit

RHeckerIntel:
Card has button that submits the form. Easiest fix is to remove form tag.


MarkRedeman (approved):

MarkRedeman:
I will check if we can change the card to not be a submit button (but rather a selection box), not high prio though since this is already fixing it.

PR #517, opened 2026-04-20
fix: enable sycl cache

### LINKED ISSUE #516: [Bug]: Extreme slow loading of SmolVLA on model inference
### Description

On inference of SmolVLA, it would take up to 6 minutes, especially on first inference.

### Steps to Reproduce

1. Run first inference of SmolVLA

### Expected Behavior

1. SmolVLA initial inference should be fast if cached.

### Actual Behavior

On inference of SmolVLA, it would take up to 6 minutes, especially on first inference.

### Component

Library (VLA framework)

### Environment

```markdown

```

### Logs / Error Output

```shell

```

### Additional Context

_No response_

desmondsow:
Enable sycl cache for model loading, resolve SmolVLA and others too loading slow on subsequent restarts

Fix #516

### DIFF
--- a/application/docker/.env.example
+++ b/application/docker/.env.example
@@ -17,6 +17,12 @@ PORT=7860
 # profiled service to start (see docker-compose.yaml).
 COMPOSE_PROFILES=cpu
 
+# Persist the SYCL kernel cache across container restarts so the first
+# model inference (which compiles hundreds of SYCL kernels) only pays
+# the full ~minutes-long compile cost once.
+SYCL_CACHE_PERSISTENT=1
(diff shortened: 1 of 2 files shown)


MarkRedeman (approved):
Thanks! I didn't see such slow loading times myself yet (tested mainly on Lunar Lake 258), but will try to reproduce your findings.

I'm checking https://intel.github.io/llvm/EnvironmentVariables.html now and see that there is also a `SYCL_CACHE_MAX_SIZE` which is set to 8Gb by default.
Do you know if this is enough?

Issue #508, opened 2026-04-17
Implement loss graphs in model page


RHeckerIntel:
There should be a visual representation of the loss graphs in the model page.

Specifically both for the currently training models and the models already trained.

PR #518, opened 2026-04-21
feat: Model Metrics

RHeckerIntel:
GH-508

Adds showing loss graph in the model screen for both train jobs and models.

**A storage system change was needed for this.** 
Models in training are now stored in `storage_dir / cache`. Upon completion they are moved to the models dir. This is required because train job is unaware of the model id since it's not persisted yet. Theoretically we could also assign a model id, but separating completed models with models in training seemed neater to me.

**Snapshots**
Aside from that there is also a `storage_dir / snapshots` for storing the snapshots. Previously they were stored with the model. But since they are needed during training and the snapshots are created directly before training - moving them and then updating the database path seemed bad. This also laid bare that a lot of snapshots are simply duplicates. We should perhaps check that in some way and just reuse the existing snapshot to save disk space.

image


application/backend/src/api/job.py
```diff
@@ -59,10 +67,26 @@ async def interrupt_job(
...
+    if job is None:
+        raise ResourceNotFoundError(ResourceType.JOB, job_id)
+
+    if job.status == JobStatus.RUNNING:
+        scheduler.training_interrupt_event.set()
+    await job_service.update_job_status(job_id, status=JobStatus.CANCELED)
+
+
+@router.get("/{job_id}/model_metrics")
+async def get_model_job_metrics(
```
MarkRedeman, line 79: I feel like having both a `stream_metrics` from the models api and a `get_model_job_metrics` is a bit overkill, partially as well because not all jobs would be able to support this (in #520 ) we are adding a new dataset import job)

I guess the main reason for adding this is that we want to stream the metrics before actually storing the model to the database? (e.g. while it's training)

At the moment I think I'd prefer to immediately store the model in the database but have it get a `training_status: "training" | "done" | "error"` that we can then use to determine how to visualize the model. 
That way we might also be able to open up more advanced features like showing different stats for models that are still training vs models that finished their training? Or even allow to export a model that is still training?

application/backend/src/services/job_service.py
```diff
@@ -100,6 +100,12 @@ async def update_job_status(
                 updates["end_time"] = datetime.datetime.now(tz=datetime.UTC)
             return await repo.update(job, updates)
 
+    @staticmethod
+    async def update_job(job: Job, update: dict) -> Job:
+        async with get_async_db_session_ctx() as session:
+            repo = JobRepository(session)
```
MarkRedeman, line 136: Perhaps add a check here that `progress`, `end_time` and `status` should be changed via `update_job_status`, so that we can guarantee that the timings etc are still ok?

application/ui/src/routes/models/index.tsx
MarkRedeman, line 1: Quick general note: ideally the `routes` folder should only contain files that are actually route objects and not necessarily ui components (like job row, model row etc) those should be put into a corresponding `feature` folder so that the routes folder stays somewhat clean.

The main benefit of that architecture is that `src/routes` could be used as an "entrypoint" of the application: if you want to know how to change a component in the UI you'd first go to the route that shows the component and then dive in from there.

MarkRedeman (commented):
Some quick initial comments, didn't finish review yet

MarkRedeman (approved):

ArendJanKramer (approved):

PR #521, opened 2026-04-21
feat(capture): add basler camera support

maxxgx:
# Pull Request

## Description

Add Basler GigE/USB camera support to `physicalai.capture` via pypylon, tested on [a2A1920-51gcBAS](https://www.baslerweb.com/en/shop/a2a1920-51gcbas/) camera.

- Captures at full sensor resolution and resizes in software (color GigE models don't support binning). 
- Thread-safety contract documented on the `Camera` base class. Basler and RealSense camera implementation are not thread safe. 
- Includes example notebook, discovery + basic usage.

Note: multi consumer access (multi-thread and IPC) is WIP and coming soon for all camera types

## Type of Change

- [x] ✨ `feat` - New feature
- [ ] 🐞 `fix` - Bug fix
- [ ] 🧪 `test` - Tests
- [ ] 🔧 `chore` - Maintenance

## Breaking Changes

---


physicalai/src/physicalai/capture/cameras/basler/_camera.py
```diff
@@ -0,0 +1,257 @@
...
+        height: int | None = None,
+        color_mode: ColorMode = ColorMode.RGB,
+    ) -> None:
+        super().__init__(color_mode=color_mode)
+        self._serial_number = serial_number
+        self._fps = fps
+        self._requested_width = width
+        self._requested_height = height
+        self._width: int = 0
+        self._height: int = 0
```
MarkRedeman, line 43: What's the difference between `_requested_height` and `height`? Perhaps we can add a clarifying comment here?

physicalai/src/physicalai/capture/cameras/basler/_camera.py
```diff
@@ -0,0 +1,257 @@
...
+        self._connected = False
+
+    def _resize(self, data: np.ndarray) -> np.ndarray:
+        """Downsample full-sensor frame to the requested output dimensions.
+
+        Returns:
+            Resized frame array, or the original if no resize is needed.
+        """
+        if self._needs_resize and self._row_indices is not None and self._col_indices is not None:
+            return data[np.ix_(self._row_indices, self._col_indices)]
```
MarkRedeman: Would this not "truncate" the camera view rather than resizing (with interpolation?)?

physicalai/src/physicalai/capture/cameras/basler/_camera.py
```diff
@@ -0,0 +1,257 @@
...
+            self._converter.OutputPixelFormat = pylon.PixelType_RGB8packed
+        elif self._color_mode == ColorMode.BGR:
+            self._converter.OutputPixelFormat = pylon.PixelType_BGR8packed
+        else:
+            self._converter.OutputPixelFormat = pylon.PixelType_Mono8
+
+        self._camera.StartGrabbing(pylon.GrabStrategy_LatestImageOnly)
+        return self._camera, self._converter
+
+    def _wait_first_frame(
```
MarkRedeman, line 118: This implementation looks almost the same as `read`, would using read here make things easier?

physicalai/src/physicalai/capture/cameras/basler/_camera.py
```diff
@@ -0,0 +1,257 @@
...
+                # GetArray() returns a view into the PylonImage's C++ buffer; copy
+                # so the numpy array survives after converted is garbage-collected.
+                data = self._resize(converted.GetArray().copy())
+                grab_result.Release()
+                self._last_frame_data = data
+                self._sequence += 1
+                self._last_timestamp = time.monotonic()
+                return Frame(data=data, timestamp=self._last_timestamp, sequence=self._sequence)
+            grab_result.Release()
+
```
MarkRedeman, line 252: Could we not skip this and immediately check `self._last_frame_data` instead? (assuming `read` would have been used before?

physicalai/src/physicalai/capture/cameras/basler/_camera.py
maxxgx: It is resizing via subsampling, basically taking every nth pixel and discarding the rest.

It produces worse quality than interpolation (e.g. aliasing), but it's faster and does not require opencv dependecy. 

I'm actually not sure if quality matters, if so we should do interpolation and accept opencv as dependecy.

MarkRedeman: Ok makes sense. Perhaps instead of adding a direct dependency on opencv, which is kinda big we could add a `resize_method` argument to the camera, which can default to this subsampling method while allowing users to replace it with other methods that use interpolation.

physicalai/src/physicalai/capture/cameras/basler/_camera.py
maxxgx: The logic here is to grab the latest frame the C++ `pylon` thread has in its buffer. If no frames have been captured since the last read and the buffer is empty, we return the cached frame. The difference from `read()` is that `read()` waits for a new frame.

physicalai/src/physicalai/capture/cameras/basler/_camera.py
maxxgx: I've added `_convert_result` to simplify the conversion step.

I considered a shared `_grab_one()` helper for all three read methods, but `read_latest()` is non-blocking and uses a cached fallback, so its grab semantics are fundamentally different. For the two blocking callers, `read()` and `_wait_first_frame()`, extracting the loop would save ~ 10 lines, but each caller would still need its own exception handling and post-grab logic, so the net gain is minimal. I think the current structure, with `_convert_result` as the shared helper, is good enough.

~PS: still need to push update with `_convert_result`~

physicalai/src/physicalai/capture/cameras/basler/_camera.py
maxxgx: Added comment

physicalai/src/physicalai/capture/cameras/basler/_camera.py
maxxgx: I've switched to interpolation via `cv2.resize`, I think keeping image quality is preferable when in doubt. Subsampling with numpy is probably premature optimization at this point

MarkRedeman (approved):

PR #523, opened 2026-04-22
refactor: move extra export args from internal model to policy

sovrasov:
# Pull Request

## Description

- `extra_export_args` property was moved from torch model to lightning policy. The reason is that the property is more related to the whole policy now, as it contains full pre/post processing parameters, which the model is not aware of.
- forward empty cameras to pi05 numpy preprocessing

## Type of Change


eugene123tw (approved):
Thanks Vlad!

PR #525, opened 2026-04-23
fix: disable training expert only mode in Pi05 configuration

eugene123tw:
# Pull Request

Disable training expert only mode in Pi05 configuration

## Description

Disable expert-only training mode for pi05. It limits model capability and prevents convergence. 

Caveat: this change increase memory usage, however, it is the only way right now to get a workable model.

## Type of Change

- [ ] ✨ `feat` - New feature
- [x] 🐞 `fix` - Bug fix
- [ ] 🧪 `test` - Tests
- [ ] 🔧 `chore` - Maintenance

## Breaking Changes

---


RHeckerIntel (dismissed):

MarkRedeman (dismissed):
Is this something that we could provide as a "Advanced parameter" when training in studio?

RHeckerIntel:
The PR is missing the why. 

Pi0.5 doesn't produce good values with "train_export_only=True". Sadly, this does increase the training memory foodprint enormously. But having it work is more important than having it the memory issue.

eugene123tw:
> Is this something that we could provide as a "Advanced parameter" when training in studio?

yes, but not only this. There are a few parameters we should expose as "Advanced parameters". I believe this should be also the case for other models where we could configure advanced parameters from the UI.

eugene123tw:
> The PR is missing the why.
> 
> Pi0.5 doesn't produce good values with "train_export_only=True". Sadly, this does increase the training memory foodprint enormously. But having it work is more important than having it the memory issue.

Thanks added.

MarkRedeman (approved):

library/src/physicalai/policies/pi05/policy.py
```diff
@@ -306,15 +306,15 @@ def _from_hf(  # noqa: PLR6301, PLR0913, PLR0915
         self,
         pretrained_name_or_path: str | Path,
         *,
-        dtype: Literal["bfloat16", "float32"] = "float32",
-        n_action_steps: int | None = 10,
+        dtype: Literal["bfloat16", "float32"] = "bfloat16",
+        n_action_steps: int | None = 50,
```
sovrasov, line 310: Is this related to chunk size? Does it affect LIBERO score?

eugene123tw: chunk_size is how many action steps the model predicts in one forward pass (horizon).
n_action_steps is how many of those predicted steps are actually executed before querying the model again.

So n_action_steps <= chunk_size. This is to predict a full chunk but only execute a prefix of it, then re-plan.

PR #529, opened 2026-04-24
feat(capture): SharedCamera via iceoryx2 transport

maxxgx:
# Pull Request

## Description

`SharedCamera` provides zero-copy access to a physical camera through shared memory. It looks like a normal camera API: users call `connect()`, `read()`, or `read_latest()`, while the transport layer handles the rest.

A publisher process owns the device behind the scenes, but that is an internal detail. API users do not start it, manage it, or coordinate access to it. They simply use `SharedCamera`, and multiple consumers can read from the same camera without hitting exclusive-access issues.

### Why use `SharedCamera`?

- **Simple API.** Use it like a normal camera. The publisher is hidden.
- **No device ownership conflicts.** One process owns the camera, many consumers can subscribe.
- **Zero-copy.** Frames stay in shared memory while the subscriber holds the frame, so reads avoid serialization and extra copies.
- **Cross-process and cross-language.** Any runtime with iceoryx2 bindings can consume the same stream.

Other changes:
- Replaced with omni_camera with `pynokhwa` ([published on pypi](https://pypi.org/project/pynokhwa/)) dependency.

## Type of Change

- [x] ✨ `feat` - New feature
- [ ] 🐞 `fix` - Bug fix
- [ ] 🧪 `test` - Tests
- [ ] 🔧 `chore` - Maintenance

## Breaking Changes

---

## Examples

Two independent processes can read from the same camera stream concurrently:

```python
# process A
from physicalai.capture import CameraType, SharedCamera

cam = SharedCamera(CameraType.UVC, device_id=0, zero_copy=True)
cam.connect()
frame = cam.read_latest()

# process B
from physicalai.capture import CameraType, SharedCamera

cam = SharedCamera(CameraType.UVC, device_id=0, zero_copy=True)
cam.connect()
frame = cam.read_latest()
```

### Current architecture
Untitled-2026-04-22-1536(2)

### With SharedCamera
Untitled-2026-04-22-1536(3)


physicalai/docs/design/components/camera-interface.md
```diff
@@ -979,21 +994,24 @@ def discover_all() -> dict[str, list[DeviceInfo]]:
...
+camera.connect()
 
 # Pass to multiple consumers explicitly
-display_thread = Thread(target=display_loop, args=(cam,))
-record_thread = Thread(target=record_loop, args=(cam,))
+display_thread = Thread(target=display_loop, args=(camera,))
+record_thread = Thread(target=record_loop, args=(camera,))
 ```
 
+**Publisher Lifecycle Note**: In auto-spawn mode, `SharedCamera.connect()` will start the background publisher if needed. However, `SharedCamera.disconnect()` intentionally does NOT stop the publisher. The publisher is designed to self-terminate after 5 seconds of having zero active subscribers. This ensures that rapid disconnect/reconnect cycles do not cause unnecessary hardware flapping.
```
MarkRedeman, line 1013: Is the 5 second configurable, or could we have some hard disconnect option?
I guess that might help with the "bug" you mentioned where the UI couldn't easily refresh the camera resolution?

physicalai/docs/design/components/camera-interface.md
```diff
@@ -1022,6 +1040,70 @@ in `connect()`, raising `DeviceInUseError` on conflicts. Each subclass would ide
...
+  - **Subscribe-only mode**: `SharedCamera.from_publisher(service_name="...")` — connects to an existing publisher
+    without attempting to spawn one.
+- **CameraPublisher** (internal): Runs in a background process. Opens the real camera
+  via `CameraSpec.build()`, reads frames in a loop, encodes each frame with a packed
+  binary header (`FrameHeader`), and publishes to an iceoryx2 service. It monitors
+  subscriber counts and exits if idle for more than 5 seconds.
+
+**Binary Protocol** (`_header.py`):
+
+Each SHM payload is structured as `[FrameHeader (40 bytes)][RGB data][optional depth data]`.
```
MarkRedeman, line 1077: Out of curiosity is the 40 bytes of protocol information configurable? (would we ever need to have more or less information?)

physicalai/docs/design/components/camera-interface.md
```diff
@@ -1022,6 +1040,70 @@ in `connect()`, raising `DeviceInUseError` on conflicts. Each subclass would ide
...
+from physicalai.capture.transport import SharedCamera
+
+# Auto-spawn mode: Spawns publisher for "realsense" if needed
+camera = SharedCamera("realsense", serial_number="12345678")
+camera.connect(timeout=5.0)
+frame = camera.read_latest()
+camera.disconnect()
+
+# Subscribe-only mode: Connects to an existing publisher
+camera = SharedCamera.from_publisher(service_name="cam/wrist")
```
MarkRedeman, line 1094: What is the `service_name` that's used here?

physicalai/docs/design/components/camera-interface.md
```diff
@@ -1022,6 +1040,70 @@ in `connect()`, raising `DeviceInUseError` on conflicts. Each subclass would ide
...
+number, and depth offset. Helper functions `encode_frame()`, `decode_rgb()`,
+`decode_rgb_view()`, and `decode_depth()` handle serialisation.
+
+**Usage:**
+
+```python
+from physicalai.capture.transport import SharedCamera
+
+# Auto-spawn mode: Spawns publisher for "realsense" if needed
+camera = SharedCamera("realsense", serial_number="12345678")
```
MarkRedeman, line 1088: Do we also allow something like `camera = SharedCamera(reaslense_camera)`, where `realsense_camera` is an instance of a normal realsense camera?

physicalai/docs/design/components/camera-interface.md
```diff
@@ -1022,6 +1040,70 @@ in `connect()`, raising `DeviceInUseError` on conflicts. Each subclass would ide
...
+camera.disconnect()
+
+# Subscribe-only mode: Connects to an existing publisher
+camera = SharedCamera.from_publisher(service_name="cam/wrist")
+camera.connect(timeout=5.0)
+frame = camera.read_latest()
+camera.disconnect()
+
+# Zero-copy mode: frame.data is a read-only view into SHM
+camera = SharedCamera("uvc", device=0, zero_copy=True)
```
MarkRedeman, line 1100: With this example it might be confusing which arguments are related to the shared camera configuraiton and which to the uvc camera. Hence my other question on using `camera = SharedCamera(reaslense_camera)`.

Perhaps we could have,
`camera = SharedCamera(camera=reaslense_camera, zero_copy=True)` instead to make it more explicit?

physicalai/docs/design/components/camera-interface.md
maxxgx: It's configurable in the new PR. I solve the bug by adding an option to reconfigure the publisher's camera settings instead of turning on/off

physicalai/docs/design/components/camera-interface.md
maxxgx: Yes we add more metadata. Now it's 44 bytes :)

physicalai/docs/design/components/camera-interface.md
maxxgx: This is an option to pub-sub aware users. The service name is constructed by the sharedcamera, so you can do the following if you want:
```
cam1 = SharedCamera(...)
cam2 = SharedCamera.from_publisher(service_name=cam1.service_name)
```

physicalai/docs/design/components/camera-interface.md
maxxgx: We are not using camera class instances because of multiprocessing. When spawining the pub subprocess, the args needs to be pickled

physicalai/docs/design/components/camera-interface.md
maxxgx: Yeah it's not perfect, users need to know which parameters are required. It's a current limitation

maxxgx:
Closing this in favor of https://github.com/open-edge-platform/physical-ai-studio/pull/553

PR #530, opened 2026-04-24
feat: bimanual widow x ai arms

MarkRedeman:
This PR allows users to create new bimanual Widowx AI robot arms that can be used for recording datasets, training models and inference.
As it is the PR contains some changes that are not directly related to adding a new robot arm - I plan to extract these into a separate PR so that this PR can end up being used as an example of how to add new robots.
Currently this PR contains
1. Increase goal time used by WidowX AI arms to prevent oscillating behaviour (will be extracted to a separate PR)
2. Apply force feedback in the Studio's teleoperation worker (will be extracted to a separate PR)
3. Add bimanual WidowX AI to backend
4. Make sure WidowX AI arms don't move too fast (will be extracted to a separate PR)
5. Change the database type of `robot_type` to be a string. While enums are nice, it will make it harder for us to add new robots, as it would require us to add a new migration everytime we add a new robot. (can be extracted in a separate PR)
6. Integrate bimanual WidowX AI arm into the library, and use it from the application (can be extracted in a separate PR)
7. Integrate robot setup form for bimanual WidowX AI arm
8. Synchronize robot joint state with URDF model (will partially be extracted to a separate PR)

## Type of Change

- [x] ✨ `feat` - New feature


physicalai/src/physicalai/robot/trossen/bimanual_widowxai.py
maxxgx, line 1: I think it would be cool to implement the bimanual mode as a mixin class. That way you can essentially get bimanual mode "for-free" for any type of single arm robot. Then we'd get SO101 bimanual as well.

In this case, with a BimanualMixin, you'd only need to add a `set_external_efforts` method since the all other functionalities are covered by the interface.

I think it would very nice to add it now, but I'll leave the decision to you.

physicalai/src/physicalai/robot/trossen/bimanual_widowxai.py
```diff
@@ -0,0 +1,184 @@
...
+        Raises:
+            RuntimeError: If called on a leader robot.
+            ValueError: If action shape is not ``(14,)``.
+        """
+        if self.role == "leader":
+            msg = "Cannot send actions to a leader robot."
+            raise RuntimeError(msg)
+
+        n = self._left.NUM_JOINTS
+        expected_shape = (2 * n,)
```
maxxgx: A slightly more generic approach would be `n = self._left.NUM_JOINTS + self._right.NUM_JOINTS`. 

In case we want to enable asymmetrical bimanual robots for fun in the future.

application/backend/src/alembic/versions/b9f3e2a1c7d8_add_bimanual_widowxai_robot_types.py
```diff
@@ -0,0 +1,66 @@
...
+    with op.batch_alter_table("project_robots", schema=None) as batch_op:
+        # Re-declare the column with the extended set of allowed values.
+        # batch_alter_table recreates the table, so the new enum values are
+        # picked up without any explicit ALTER TYPE.
+        batch_op.alter_column(
+            "type",
+            existing_type=sa.String(255),
+            type_=sa.String(255),
+            nullable=False,
+        )
```
maxxgx, line 37: Is type an enum column? I think this needs to be changed to a generic str type. Especially, if we plant to keep adding new robots to the application and have a plugin system. I don't think it's sustainable to write a DB migration for each new robot

application/backend/src/robots/discovery/ip.py
```diff
@@ -21,6 +22,19 @@ async def ping(ip: str, ping_timeout: float = 1.0) -> bool:
...
 
     async def is_reachable(self, robot: Robot) -> bool:
+        if not isinstance(robot, SO101Robot | TrossenSingleArmRobot):
+            return False
         if not robot.payload.connection_string:
             return False
         return await self.ping(robot.payload.connection_string)
+
+    async def is_reachable_bimanual(self, robot: Robot) -> bool:
+        """Ping both arms of a bimanual robot; returns True only if both are reachable."""
```
maxxgx, line 32: The current ping method does not see to check actual robot connectivity. It only creates an event (where is it consumed?) without going through IP/serial connections

application/backend/src/robots/widowxai/bimanual_adapter.py
maxxgx, line 1: Just an observation: looks very similar to single arm WidowXAI adapter. That's good, it means that a generalized adapter can  work.

maxxgx (commented):
Looking good, thanks!

physicalai/src/physicalai/robot/trossen/bimanual_widowxai.py
MarkRedeman: Yeah I think that would be nice. I'd prefer to first keep it as is though and iterate over this a bit more.
On the library side I don't think it will be difficult, but on the application side we need to think about how to properly visualize this and what kind of restrictions we would want to add.
For instance would we allow adding a WidowX arm as left and SO101 as right? I don't see why not, but would be good to first think of some consequences.

We could also make it even more general and add a `CompositeRobot` that can work with any N robots? Should be relatively easy to add into the library, but harder to figure out for the application.

application/backend/src/alembic/versions/b9f3e2a1c7d8_add_bimanual_widowxai_robot_types.py
MarkRedeman: Merged via #536, with this change we use a string for the robot types.

physicalai/src/physicalai/robot/trossen/bimanual_widowxai.py
```diff
@@ -0,0 +1,184 @@
...
+    def __init__(self, left: WidowXAI, right: WidowXAI) -> None:
+        """Initialize BimanualWidowXAI with two single-arm drivers.
+
+        Raises:
+            ValueError: If left and right arms have different roles.
+        """
+        if left.role != right.role:
+            msg = f"Both arms must have the same role; got left={left.role!r}, right={right.role!r}."
+            raise ValueError(msg)
+        if type(left) is not type(right):
```
MarkRedeman: @maxxgx I added this constraint in the constructor to enfoce that `left` and `right` are the same type.
The main motivation for that was that we then need to do less defensive programming in other parts of this class. 
For instance when getting observation and zipping the observation and joint names together. Or when sending actions.

ArendJanKramer (dismissed):

ArendJanKramer (approved):

PR #537, opened 2026-04-28
fix: misc trossen widowx improvements

MarkRedeman:
This PR introduces some fixes based on the recent (bimanual) widowx integrations:
1. Set a higher `goal_time` to reduce oscillations during teleopration. Alternatively we could (should?) increase the FPS at which we read and set joint states
2. Apply force feedback during teleoperation
3. Prevent high joint velocity when teleoperating

## Type of Change

- [x] 🐞 `fix` - Bug fix

### DIFF
--- a/application/backend/src/robots/widowxai/adapter.py
+++ b/application/backend/src/robots/widowxai/adapter.py
@@ -115,7 +115,9 @@ class WidowXAIAdapter(RobotClient):
             await asyncio.to_thread(
                 self._robot.send_action,
                 positions,
-                goal_time=goal_time,
+                # Increase default goal_time to reduce oscillations due to the application
+                # performing teleoperation at 30Hz
+                goal_time=3 * goal_time,
             )
 
         return self._create_event("joints_state_was_set", joints=joints)
--- a/physicalai/src/physicalai/robot/trossen/widowxai.py
+++ b/physicalai/src/physicalai/robot/trossen/widowxai.py
@@ -211,9 +211,9 @@ class WidowXAI(Robot):
 
         if self._last_positions is not None:
             velocities = ((action - self._last_positions) / goal_time).tolist()
-            driver.set_all_positions(action.tolist(), goal_time, False, velocities)  # noqa: FBT003
+            driver.set_all_positions(action.tolist(), 3 * goal_time, False, velocities)  # noqa: FBT003
         else:
-            driver.set_all_positions(action.tolist(), goal_time, False)  # noqa: FBT003
+            driver.set_all_positions(action.tolist(), 3 * goal_time, False)  # noqa: FBT003
 
         self._last_positions = action.copy()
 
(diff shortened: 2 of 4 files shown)


physicalai/src/physicalai/robot/trossen/widowxai.py
```diff
@@ -209,13 +221,10 @@ def send_action(self, action: np.ndarray, *, goal_time: float = 0.1) -> None:
...
-        if self._last_positions is not None:
-            velocities = ((action - self._last_positions) / goal_time).tolist()
-            driver.set_all_positions(action.tolist(), goal_time, False, velocities)  # noqa: FBT003
-        else:
-            driver.set_all_positions(action.tolist(), goal_time, False)  # noqa: FBT003
+        present_positions = np.asarray(driver.get_all_positions(), dtype=np.float32)
+        safe_action = self._ensure_safe_goal_position(action, present_positions, self.MAX_RELATIVE_TARGET)
 
-        self._last_positions = action.copy()
+        driver.set_all_positions(safe_action.tolist(), 3 * goal_time, False)  # noqa: FBT003
```
maxxgx: Why not make the default `goal_time` longer or have the caller set a higher value? I think adding an arbitrary constant multiplier can be confusing for the library users

MarkRedeman: Agree! I've moved it into the application side.

application/backend/src/robots/widowxai/adapter.py
```diff
@@ -115,7 +115,9 @@ async def set_joints_state(self, joints: dict, goal_time: float) -> dict:
             await asyncio.to_thread(
                 self._robot.send_action,
                 positions,
-                goal_time=goal_time,
+                # Increase default goal_time to reduce oscillations due to the application
+                # performing teleoperation at 30Hz
+                goal_time=3 * goal_time,
```
RHeckerIntel, line 120: This is a pretty big change. Won't the follower lag too much behind?

physicalai/src/physicalai/robot/trossen/widowxai.py
```diff
@@ -209,13 +221,10 @@ def send_action(self, action: np.ndarray, *, goal_time: float = 0.1) -> None:
 
         driver = self._require_driver()
 
-        if self._last_positions is not None:
-            velocities = ((action - self._last_positions) / goal_time).tolist()
-            driver.set_all_positions(action.tolist(), goal_time, False, velocities)  # noqa: FBT003
-        else:
-            driver.set_all_positions(action.tolist(), goal_time, False)  # noqa: FBT003
```
RHeckerIntel, line 216: We drop the velocities? Is this tested in any way?

application/backend/src/robots/widowxai/adapter.py
MarkRedeman: I didn't notice any, but didn't do thorough testing on it. This setting and the velocity check was suggested by @ArendJanKramer

physicalai/src/physicalai/robot/trossen/widowxai.py
MarkRedeman: Yes but I won't mind either if we spend some more time on this.
Without this check the thing I kept seeing is that the teleoperator worker would crash due to an exception that was raised by the trossen driver, due to the joints moving too quickly.

I didn't have enough time to properly check this though so would be good to have another pair of eyes on it.

RHeckerIntel (approved):

PR #538, opened 2026-04-28
refactor: move inference

sovrasov:
# Pull Request

## Description

- Move inference module to `physicalai`
- Torch and executorch adapters stayed in `physicalai-train`
- A registry for inference adapters registers collects difference adapters from train and inference packages
- Split and move related unit tests

## Type of Change

## Breaking Changes

Some inference-related imports may break, like explicit import of torch adapter


application/backend/pyproject.toml
```diff
@@ -10,7 +10,7 @@ dependencies = [
   'harvesters; sys_platform != "darwin"',
   "fastapi[standard]",
   "physicalai-train[pi0,smolvla]",
-  "physicalai[so101, trossen]",
+  "physicalai[inference,so101,trossen]",
```
samet-akcay: is `physicalai[inference]` temporary? We thought `physicalai` would be inference by default ?

library/src/physicalai/inference/adapters/_register.py
```diff
@@ -0,0 +1,36 @@
...
+
+from __future__ import annotations
+
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from physicalai.inference.adapters.registry import InferenceBackendRegistry
+
+
+def register(registry: InferenceBackendRegistry) -> None:
```
samet-akcay: maybe `AdapterRegistry` for consistency? or `BackendRegistry`?

application/backend/pyproject.toml
sovrasov: No, I was thinking that all submodules are equal after we dropped `inference` from the core package name:
`physicalai-inference` -> `physicalai`. Installing inference deps by default restrict users who might use the repo for camera or robot support only. I'll make it default then

library/src/physicalai/inference/adapters/_register.py
sovrasov: I'll rename the registry for now

samet-akcay (dismissed):

library/src/physicalai/inference/adapters/_register.py
maxxgx, line 1: What's the reason to keep `library/src/physicalai/inference/adapters/_register.py` here?

Do we still need this folder in library?

sovrasov: `_register.py` is some sort of replacement of init.py, which we can not use because of shared namespace.
This folder extends the frameworks adapters folder with the adapters we can't move. Overall, now inference subfolder here is for initial models enablement only. Like we add a new pytorch policy -> prototype export / inference in the library and while still polishing it, can utilize it from the same namespace `physicalai.inference.*`

maxxgx (dismissed):

physicalai/src/physicalai/inference/adapters/__init__.py
```diff
@@ -0,0 +1,114 @@
...
+# dependencies (onnxruntime, openvino) are part of the `inference` extra.
+from physicalai.inference.adapters.openvino import OpenVINOAdapter
+from physicalai.inference.adapters.onnx import ONNXAdapter
+
+logger = logging.getLogger(__name__)
+
+#: Entry-point group used by other distributions to contribute adapters.
+#:
+#: Each entry point must point at a callable with the signature
+#: ``register(registry: InferenceBackendRegistry) -> None``.  The callable
```
samet-akcay: left-over?

physicalai/src/physicalai/inference/adapters/__init__.py
```diff
@@ -0,0 +1,114 @@
...
+from pkgutil import extend_path
+from typing import Any
+
+# Allow `physicalai.inference.adapters` to be split across multiple
+# distributions sharing the `physicalai` namespace.
+__path__ = extend_path(__path__, __name__)
+
+from physicalai.inference.adapters.base import RuntimeAdapter
+from physicalai.inference.adapters.registry import (
+    InferenceAdapterRegistry,
```
samet-akcay: nitpicky, but still think this might look a bit inconsistent as `InferenceAdapter` is not mentioned anywhere else? Do you think if it is the right terminology?

samet-akcay (dismissed):

physicalai/src/physicalai/inference/adapters/__init__.py
sovrasov: My bad, I'm too biased from model_api times

physicalai/src/physicalai/inference/adapters/__init__.py
sovrasov: Right, updated

samet-akcay (approved):

PR #539, opened 2026-04-28
feat: add compute_val_loss method for validation loss calculation in ACT model

eugene123tw:
# Pull Request

## Description

During validation, PyTorch Lightning sets the model to eval mode (`self.training = False`). The ACT model's `_ACT.forward()` gates the VAE encoder behind `self.training`, so during validation loss computation `mu` and `log_sigma_x2` are `None`. This causes a `TypeError` in the KL divergence calculation: `unsupported operand type(s) for +: 'int' and 'NoneType'`.

This PR adds a dedicated `compute_val_loss` method to the `ACT` model that temporarily sets the inner `_ACT` module to training mode so the VAE encoder runs and produces valid latent distribution parameters. The model is restored to eval mode via `try/finally`, and `@torch.no_grad()` ensures no gradients are computed.

## Type of Change

- [x] 🐞 `fix` - Bug fix

## Breaking Changes

None. The base class `Model.compute_val_loss` already delegates to `compute_loss` by default. This override only affects the ACT policy's validation path.

---

## Examples

**Before:** Training with `use_vae=True` crashes at the first validation epoch:
```
File "library/src/physicalai/policies/act/model.py", line 283, in compute_loss
    mean_kld = (-0.5 * (1 + log_sigma_x2_hat - mu_hat.pow(2) - ...
TypeError: unsupported operand type(s) for +: 'int' and 'NoneType'
```

**After:** Validation loss is computed correctly, reporting both `l1_loss` and `kld_loss` in the loss dict.

### DIFF
--- a/library/src/physicalai/policies/act/model.py
+++ b/library/src/physicalai/policies/act/model.py
@@ -290,6 +290,45 @@ class ACT(ExportableModelMixin, Model):
         return loss, loss_dict
 
     @torch.no_grad()
+    def compute_val_loss(self, batch: dict[str, torch.Tensor]) -> tuple[torch.Tensor, dict[str, float]]:
+        """Compute validation loss (L1 + optional KL divergence).
+
+        Temporarily sets the inner model to training mode so the VAE encoder
+        runs and produces latent distribution parameters (mu, log_sigma_x2).
+        The model is restored to eval mode afterwards.  Dropout and batch-norm
+        behaviour during this call will follow training-mode semantics, but
+        gradients are still disabled by the ``@torch.no_grad()`` decorator.
+
+        Args:
+            batch: Preprocessed batch dict.
+
+        Returns:
+            Tuple of (loss tensor, loss dict with ``"l1_loss"`` and optionally ``"kld_loss"``).
+        """
+        batch = self._input_normalizer(batch)
+
+        self._model.train()
+        try:
+            actions_hat, (mu_hat, log_sigma_x2_hat) = self._model(batch)
+        finally:
+            self._model.eval()
+
+        l1_loss = (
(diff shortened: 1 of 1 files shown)


MarkRedeman (approved):
Can confirm that this fixes training ACT for me.

> The ACT model's _ACT.forward() gates the VAE encoder behind self.training, so during validation loss computation mu and log_sigma_x2 are None.

Would it not be possible to remove that gating behaviour? The current solution feels more like we are circumventing the actual buggy behaviour?

eugene123tw:
> Can confirm that this fixes training ACT for me.
> 
> > The ACT model's _ACT.forward() gates the VAE encoder behind self.training, so during validation loss computation mu and log_sigma_x2 are None.
> 
> Would it not be possible to remove that gating behaviour? The current solution feels more like we are circumventing the actual buggy behaviour?

Removing the `self.training` gate from `_ACT.forward()` would cause the VAE encoder to run whenever `ACTION` is in the batch, potentially slowing down inference and test-time prediction. To avoid that, this fix is scoped to a dedicated `compute_val_loss` override that temporarily enables training mode only for the forward pass, leaving the original `forward()` logic untouched.

sovrasov:
> Can confirm that this fixes training ACT for me.
> 
> > The ACT model's _ACT.forward() gates the VAE encoder behind self.training, so during validation loss computation mu and log_sigma_x2 are None.
> 
> Would it not be possible to remove that gating behaviour? The current solution feels more like we are circumventing the actual buggy behaviour?

It looks that way because the val loss technique is a bit non standard and now implemented outside of pytorch lightning logic. Forward implementation we have matches standard torch flow, and the buggy behavior rather comes from design issues on our side

sovrasov (approved):

PR #541, opened 2026-04-28
docs: modular packages in `physicalai` repo

samet-akcay:
# Pull Request

## Type of Change

## Description

This PR discusses how far to modularize the runtime as `physicalai` evolves into its own repo.

Some parts of the runtime, especially inference and capture, are generic enough that they could be published independently. At the same time, splitting them into separate repositories too early would add coordination, release, and maintenance overhead.

This proposal is intended to make that tradeoff explicit and gather feedback before we commit to a packaging direction.

### DIFF
--- a/physicalai/docs/design/packaging/modular-packages-in-one-repo.md
+++ b/physicalai/docs/design/packaging/modular-packages-in-one-repo.md
@@ -1,0 +1,286 @@
+# Modular Packages in One Repo
+
+Design note for shipping reusable libraries such as `inferencekit` and `capturekit`
+from the `physical-ai` repo without forcing an immediate repo split.
+
+---
+
+## Problem
+
+Some parts of the `physicalai` runtime are more general than physical AI itself.
+For example,
+
+- `physicalai.inference` is a generic inference engine with backends, manifests, and runners
+- `physicalai.capture` is a generic camera interface with hardware backends
+
+These modules may eventually be useful to other domains and other repositories. At the
+same time, creating separate repositories too early adds release and maintenance overhead.
+
+We want a structure that lets us:
+
+- keep development in one repository for now
+- publish reusable pieces as standalone packages
+- avoid duplicate implementations
+- make future extraction cheap if a separate repo becomes justified
+
+---
+
+## Approach
+
+This document proposes to use a **multi-package monorepo** approach.
+
+We could keep `inferencekit`, `capturekit`, and `physicalai` in the same repository, but publish
+them as separate Python distributions.
+
+```text
+physical-ai/
+├── packages/
+│   ├── inference/
+│   │   ├── pyproject.toml
+│   │   └── src/inferencekit/
+│   ├── capture/
+│   │   ├── pyproject.toml
+│   │   └── src/capturekit/
+│   └── physicalai/
+│       ├── pyproject.toml
+│       └── src/physicalai/
+└── pyproject.toml
+```
+
+This gives us one repo, one CI surface, and one development workflow, while still allowing:
+
+- `pip install inferencekit`
+- `pip install capturekit`
+- `pip install physicalai`
+
+The important point is that each distribution owns its own import path.
+
+---
+
+## Motivation
+
+This could be a better fit than either of the two common alternatives.
+
+### 1. Better than premature repo splits
+
+Creating a new repository for every reusable module sounds clean, but it creates real cost:
+
+- separate CI and release pipelines
+- version coordination across repos
+- more difficult local development
+- more coordination overhead for small API changes
+
+If there is not yet a concrete external consumer, that overhead is usually wasted.
+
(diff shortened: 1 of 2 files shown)


physicalai/docs/design/packaging/modular-packages-in-one-repo.md
```diff
@@ -0,0 +1,286 @@
...
+- no cross-imports from generic packages into `physicalai`
+- package-local tests
+- package-local `pyproject.toml`
+- package-local README and API ownership
+
+### Phase 2
+
+Publish `inferencekit` and/or `capturekit` from the same repo.
+
+`physicalai` switches from local module ownership to dependency + re-export.
```
maxxgx: What are the conditions for publishing these as packages?

I think we should validate them with an actual consumer first. Other Geti projects are a good candidate. Publishing would add overhead around releases and versioning, and it'd also means that we have to think about breaking changes carefully if other projects depend on them. IMO, we should publish when we are confident that the API is stable.

Also, would `inferencekit`, `capturekit`, and `physicalai` share one version number or version independently?

physicalai/docs/design/packaging/modular-packages-in-one-repo.md
```diff
@@ -0,0 +1,286 @@
...
+For example, we have seen that we could easily inject physical-ai related context
+into this generic `inferencekit` interfaces.
+
+Separate distributions create a harder boundary:
+
+- `inferencekit` cannot import `physicalai`
+- `capturekit` cannot import `physicalai`
+- `physicalai` depends on them, not the reverse
+
+That dependency direction is the main design goal.
```
maxxgx, line 93: How should we enforce this boundary? Perhaps, we could use something like `import-linter`

physicalai/docs/design/packaging/modular-packages-in-one-repo.md
```diff
@@ -0,0 +1,286 @@
...
+packages/physicalai/src/physicalai/
+```
+
+Then `physicalai` can provide thin re-exports:
+
+```python
+# physicalai/inference/__init__.py
+from inferencekit import InferenceModel
+
+__all__ = ["InferenceModel"]
```
maxxgx: Users would be able to import from two places, `inferencekit` and `physicalai.inference`. For example, this could be an issue when pickling two objects: one from `inferencekit` and the same object re-exported from `physicalai.inference`, since pickles save the class path. The former approach, requires `inferencekit`, while the latter, `physicalai.inference`

maxxgx (commented):
A few comments, but I like the overall direction

physicalai/docs/design/packaging/modular-packages-in-one-repo.md
samet-akcay: Agreed. If we go in this direction, the boundary could be enforced in CI rather than relying on discipline alone. `import-linter` sounds like a good fit, or something equivalent if we already have another preferred mechanism.

I've updated the document.

physicalai/docs/design/packaging/modular-packages-in-one-repo.md
samet-akcay: Previous example might not be a good one, so I have updated it. 

I think we may still want two layers, but not as a pure re-export. Maybe something like this : 

- `inferencekit.InferenceModel` as the generic base API
- `physicalai.InferenceModel` or `InferencePolicy` as a domain-specific subclass.

In our case, `physicalai` needs user-facing methods such as `select_action` and `predict_action_chunk` that do not belong in the generic base class anyway. 

I have updated the document and example with more realistic use-case.

physicalai/docs/design/packaging/modular-packages-in-one-repo.md
samet-akcay: good point, I agree that we should not treat them publishing as automatic.

my intent here is to describe a package shape that keeps the boundaries clean if we decide to publish later, not to say that inferncekit or capture should be released immediately. 

I have updated the proposal accordingly. 

regarding versioning, I left it intentionally open for now. Shared versioning is simpler if they always move together; independent versioning only makes sense once they become real separately consumed products

physicalai/docs/design/packaging/modular-packages-in-one-repo.md
```diff
@@ -0,0 +1,350 @@
...
+That dependency direction is the main design goal.
+
+If we adopt this direction, we should enforce it in CI rather than rely on convention alone.
+`import-linter` or an equivalent import-boundary check would be a reasonable way to do that.
+
+---
+
+## Package Roles
+
+### `inferencekit`
```
MarkRedeman, line 102: I don't quite get what the responsibility for this package (vs `physicalai`) will be. Who are its users? Is it replacing ModelAPI?

maxxgx (approved):

PR #547, opened 2026-04-30
refactor: move logs to footer to make them more visible

MarkRedeman:
This PR adds a footer to the project layout and uses it to show a button to open the application logs as well as show training status:

image

Later I'd also like to add a "jobs management" panel here where users can more easily find failed, running and pending jobs as well as their associated logs.

(in the screenshot I also removed the "Models" heading, but I reverted that as it made the empty models page look too empty, will extract that in a separate PR)

## Type of Change


RHeckerIntel (dismissed):
❤️

RHeckerIntel (dismissed):

RHeckerIntel (approved):

PR #549, opened 2026-04-30
chore: remove heading from models page

MarkRedeman:
Adding some minor visual tweaks before recording a video of the application.

- Small visual improvements that make the empty dataset and model page feel more consistent.
- The "Models" heading when the user does have models has been removed as it didn't add much.

## Type of Change

- [x] 🔧 `chore` - Maintenance

## Screenshots

| **With models** | **Empty models** | **Empty dataset** |
|------------|-----------|-----------|
| image | image | image |


RHeckerIntel (approved):
Also have this in the model metrics pr. We should probably update that one and merge it.

PR #551, opened 2026-04-30
feat: Use available backends based on dir

RHeckerIntel:
Adds available_backends for a model which checks the directory if the export is there.
Adds get_supported_backends api call for all available policies.
Use an *AND* of these two lists to show the list of backends in the inference modal.

# Background
After training we export the policy to the policy's available backends. We do this because loading the policy to then export will be very slow for the bigger models. 

This of course creates a caching problem.
If per chance the policy gets more or less supported backends we need to account for that.

# Future work
We can add a export on the fly option, or even a job to update the exports.


application/backend/src/api/policies.py
```diff
@@ -0,0 +1,22 @@
...
+}
+
+
+@router.get("/backends")
+def get_supported_backends_per_policy() -> dict[str, list[str]]:
+    """Return the supported export backends for each policy."""
+    return {
+        name: [str(b) for b in cls.get_supported_export_backends()]
+        if hasattr(cls, "get_supported_export_backends")
+        else []
```
MarkRedeman, line 20: Could be simplified to `getattr(cls, "get_supported_export_backends", [])`?

MarkRedeman (approved):

application/backend/src/api/policies.py
RHeckerIntel: Could, but its a function so either lambda or still an if else.

PR #553, opened 2026-05-01
feat(capture): Replace framesource with PAI capture SharedCamera

maxxgx:
# Pull Request

This PR replaces https://github.com/open-edge-platform/physical-ai-studio/pull/529

## Description

`SharedCamera` provides zero-copy access to a physical camera through shared memory. It looks like a normal camera API: users call `connect()`, `read()`, or `read_latest()`, while the transport layer handles the rest.

A publisher process owns the device behind the scenes, but that is an internal detail. API users do not start it, manage it, or coordinate access to it. They simply use `SharedCamera`, and multiple consumers can read from the same camera without hitting exclusive-access issues.

### Why use `SharedCamera`?

- **Simple API.** Use it like a normal camera. The publisher is hidden.
- **No device ownership conflicts.** One process owns the camera, many consumers can subscribe.
- **Zero-copy.** Frames stay in shared memory while the subscriber holds the frame, so reads avoid serialization and extra copies.
- **Cross-process and cross-language.** Any runtime with iceoryx2 bindings can consume the same stream.

Other changes:
- Replaced `omni_camera` with `pynokhwa` ([published on pypi](https://pypi.org/project/pynokhwa/)) dependency.
- Removed `AsyncCameraCapture`
- Integrated `SharedCamera` for both recording and preview

## Type of Change

- [x] ✨ `feat` - New feature
- [ ] 🐞 `fix` - Bug fix
- [ ] 🧪 `test` - Tests
- [ ] 🔧 `chore` - Maintenance

## Breaking Changes

---

## Examples

Two independent processes can read from the same camera stream concurrently:

```python
# process A
from physicalai.capture import CameraType, SharedCamera

cam = SharedCamera(CameraType.UVC, device=0, zero_copy=True)
cam.connect()
frame = cam.read_latest()

# process B
from physicalai.capture import CameraType, SharedCamera

cam = SharedCamera(CameraType.UVC, device=0, zero_copy=True)
cam.connect()
frame = cam.read_latest()
```

### Current architecture
Untitled-2026-04-22-1536(2)

### With SharedCamera
Untitled-2026-04-22-1536(3)


maxxgx:
CI failing due to missing `libjpeg-turbo`

MarkRedeman:
> CI failing due to missing `libjpeg-turbo`

It worked as soon as I installed `libturbojpeg` on my laptop, before that starting up the server would immediately crash with,
```
uv run --no-sync src/main.py
Traceback (most recent call last):
  File "/home/msredema/projects/geti/action/geti-action/application/backend/src/main.py", line 13, in <module>
    from api.camera import router as camera_router
  File "/home/msredema/projects/geti/action/geti-action/application/backend/src/api/camera.py", line 11, in <module>
    from api.dependencies import CameraRegistryDep, RecordingLockedCamerasDep
  File "/home/msredema/projects/geti/action/geti-action/application/backend/src/api/dependencies.py", line 9, in <module>
    from core.scheduler import Scheduler
  File "/home/msredema/projects/geti/action/geti-action/application/backend/src/core/__init__.py", line 1, in <module>
    from core.lifecycle import lifespan
  File "/home/msredema/projects/geti/action/geti-action/application/backend/src/core/lifecycle.py", line 11, in <module>
    from workers.camera_worker_registry import CameraWorkerRegistry
  File "/home/msredema/projects/geti/action/geti-action/application/backend/src/workers/camera_worker_registry.py", line 9, in <module>
    from .camera_worker import CameraWorker
  File "/home/msredema/projects/geti/action/geti-action/application/backend/src/workers/camera_worker.py", line 15, in <module>
    tj = TurboJPEG()
         ^^^^^^^^^^^
  File "/home/msredema/projects/geti/action/geti-action/application/backend/.venv/lib/python3.12/site-packages/turbojpeg.py", line 301, in __init__
    self.__find_turbojpeg() if lib_path is None else lib_path)
    ^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/msredema/projects/geti/action/geti-action/application/backend/.venv/lib/python3.12/site-packages/turbojpeg.py", line 996, in __find_turbojpeg
    raise RuntimeError(
RuntimeError: Unable to locate turbojpeg library automatically. You may specify the turbojpeg library path manually.
e.g. jpeg = TurboJPEG(lib_path)
```

We will need to update the docs with this info.

Two other quick observations (will try to do review and more extensive testing later):
- My laptop's main webcam was detected twice, i.e. the picker in the camera form showed duplicate cameras
- When playing around with having two tabs at the same time I saw: `RuntimeError: Could not get device property CameraFormat: Failed to Fufill` in the server's logs. I think this was due to trying to get the supported formats of one of the duplicated webcams.

maxxgx:
> > CI failing due to missing `libjpeg-turbo`
> 
> It worked as soon as I installed `libturbojpeg` on my laptop, before that starting up the server would immediately crash with,
> 
> ```
> uv run --no-sync src/main.py
> Traceback (most recent call last):
>   File "/home/msredema/projects/geti/action/geti-action/application/backend/src/main.py", line 13, in <module>
>     from api.camera import router as camera_router
>   File "/home/msredema/projects/geti/action/geti-action/application/backend/src/api/camera.py", line 11, in <module>
>     from api.dependencies import CameraRegistryDep, RecordingLockedCamerasDep
>   File "/home/msredema/projects/geti/action/geti-action/application/backend/src/api/dependencies.py", line 9, in <module>
>     from core.scheduler import Scheduler
>   File "/home/msredema/projects/geti/action/geti-action/application/backend/src/core/__init__.py", line 1, in <module>
>     from core.lifecycle import lifespan
>   File "/home/msredema/projects/geti/action/geti-action/application/backend/src/core/lifecycle.py", line 11, in <module>
>     from workers.camera_worker_registry import CameraWorkerRegistry
>   File "/home/msredema/projects/geti/action/geti-action/application/backend/src/workers/camera_worker_registry.py", line 9, in <module>
>     from .camera_worker import CameraWorker
>   File "/home/msredema/projects/geti/action/geti-action/application/backend/src/workers/camera_worker.py", line 15, in <module>
>     tj = TurboJPEG()
>          ^^^^^^^^^^^
>   File "/home/msredema/projects/geti/action/geti-action/application/backend/.venv/lib/python3.12/site-packages/turbojpeg.py", line 301, in __init__
>     self.__find_turbojpeg() if lib_path is None else lib_path)
>     ^^^^^^^^^^^^^^^^^^^^^^^
>   File "/home/msredema/projects/geti/action/geti-action/application/backend/.venv/lib/python3.12/site-packages/turbojpeg.py", line 996, in __find_turbojpeg
>     raise RuntimeError(
> RuntimeError: Unable to locate turbojpeg library automatically. You may specify the turbojpeg library path manually.
> e.g. jpeg = TurboJPEG(lib_path)
> ```
> 
> We will need to update the docs with this info.
> 
> Two other quick observations (will try to do review and more extensive testing later):
> 
> * My laptop's main webcam was detected twice, i.e. the picker in the camera form showed duplicate cameras
> * When playing around with having two tabs at the same time I saw: `RuntimeError: Could not get device property CameraFormat: Failed to Fufill` in the server's logs. I think this was due to trying to get the supported formats of one of the duplicated webcams.

I've switch the camera discovery to pynokhwa's backend which should not contain duplicates

maxxgx:
@MarkRedeman I've tested the following flow on linux:
- Add camera with preview
- Change camera settings (added guard to not allow it during dataset recording)
- Setup environment
- Dataset recording and playback

~One issue might be that previous camera IDs might not work, but simply updating the form with the correct camera solves this issue.~

MarkRedeman:
> One issue might be that previous camera IDs might not work, but simply updating the form with the correct camera solves this issue.

That's good to know. I guess since we haven't officially released it shouldn't be a big issue, but we should inform the other teams that have been testing out the studio.
If possible it would be nice if we can migrate the cameras to new ids but I guess that might be hard to detect?

maxxgx:
> > One issue might be that previous camera IDs might not work, but simply updating the form with the correct camera solves this issue.
> 
> That's good to know. I guess since we haven't officially released it shouldn't be a big issue, but we should inform the other teams that have been testing out the studio. If possible it would be nice if we can migrate the cameras to new ids but I guess that might be hard to detect?

Added backward compatibility for Linux, though the index number might still change after a reboot, which means we have to update it manually, just like before.

application/backend/src/api/camera.py
```diff
@@ -17,21 +18,46 @@
...
+
+        return _formats_to_response(UVCCamera.query_formats(fingerprint))
+
+    if driver == "realsense":
+        from physicalai.capture.cameras.realsense import RealSenseCamera
+
+        return _formats_to_response(RealSenseCamera.query_formats(fingerprint))
+
+    if driver == "basler":
+        # TODO: Replace with cached Basler hardware discovery once implementated in physicalai.capture
```
MarkRedeman, line 47: We currently hardcode this, is that something you'd want to handle in this PR?

application/backend/src/api/project_camera.py
```diff
@@ -47,8 +53,12 @@ async def update_project_camera(
     camera_id: Annotated[UUID, Depends(get_camera_id)],
     camera_service: Annotated[ProjectCameraService, Depends(get_camera_service)],
     camera: Camera,
+    locked_camera_fingerprints: RecordingLockedCamerasDep,
 ) -> Camera:
     """Set camera."""
+    existing = await camera_service.get_camera_by_id(project_id, camera_id)
+    _ensure_not_recording(existing.fingerprint, locked_camera_fingerprints)
+    _ensure_not_recording(camera.fingerprint, locked_camera_fingerprints)
```
MarkRedeman, line 61: What's the use case for ensuring this? Suppose the user edits their camera then we show a preview of that camera while the user selects the camera + resolution, so would this then not conflict with that?

application/backend/src/api/camera.py
```diff
@@ -79,12 +105,12 @@ async def camera_websocket(
...
     await websocket.accept()
 
+    is_locked = camera.fingerprint in locked_camera_fingerprints
     worker_id = uuid4()
     camera.id = worker_id
 
-    # Use WebSocket transport
     transport = WebSocketTransport(websocket)
-    worker = CameraWorker(camera, transport)
+    worker = CameraWorker(camera, transport, is_locked=is_locked)
```
MarkRedeman, line 113: Would we not also need to add a lock to `locked_camera_fingerprints` here, since this endpoint captures the camera stream?

application/backend/src/api/record.py
```diff
@@ -93,20 +107,23 @@ async def robot_control_websocket(
...
-        process.disconnect()
-        process.join(10)
+        if process is not None:
+            process.disconnect()
+            process.join(10)
 
-    queue.close()
+        queue.close()
+    finally:
+        locked_camera_fingerprints.clear()
```
MarkRedeman, line 130: Is this something that could be handled via a context manager?

application/backend/src/api/camera.py
maxxgx: I think we need to change the flow a bit for Basler. In theory, you can configure any resolution you want in their proprietary app, Pylon Viewer, and that setting will persist. That's why I thought we could open the camera with whatever resolution is already set. The problem is that the UI always expects a width and height, so the frames couldn't be displayed.

As a temporary fix, I think we should just tell the user to configure the camera in Pylon Viewer instead.

application/backend/src/api/project_camera.py
maxxgx: `_ensure_not_recording` just check whether a camera is being used for recording. If so, we don't allow changing settings. Otherwise, camera can be updated and preview will update as well

application/backend/src/api/camera.py
maxxgx: Not sure I follow, but `is_locked` is used to determine `overwrite_settings` flag for the SharedCamera. If a camera is locked (i.e. recording in progress), then a SharedCamera will NOT overwrite the settings.

application/backend/src/api/record.py
maxxgx: I don't think a context manager is worth it here, we are only clearing the camera set in this location. 

However, this locking mechanism currently would not work with multiple recordings since it would reset all camera locks. I added a comment, I think we only support one recording at a time so this should be fine for now, right?

application/backend/src/api/project_camera.py
MarkRedeman: Ah I think my initial confusion was that this would lock the camera so that no one else could use it (hence also my other comments).
So probably we can ignore this and the other comments below.

MarkRedeman (approved):

MarkRedeman:
Have tested a few times, one issue is that users do need to reconfigure their cameras if they previously used Studio, that's a bit annoying but it's better to do this now than later.

PR #554, opened 2026-05-01
chore(ci): add step-security/harden-runner and cooldown period

AlexanderBarabanov:
# Pull Request

## Description

This PR adds step-security/harden-runner action to audit outbound network connections in CI (latter it will be switched to blocking mode) and Renovate cooldown period to mitigate potential supply chain attacks.

## Type of Change

- [x] 🔧 `chore` - Maintenance


ivanzati (approved):
Thank you, Alexander!

PR #565, opened 2026-05-07
docs: add policy runtime design

samet-akcay:
## Summary

Comprehensive runtime system design documentation with explicit API contracts.


physicalai/docs/design/components/runtime-system/design_review_summary.md
```diff
@@ -0,0 +1,118 @@
...
+
+model = InferenceModel.load("./exports/act_policy")
+robot = SO101(port="/dev/ttyACM0")
+
+runtime = PolicyRuntime(
+    robot=robot,
+    model=model,
+    execution=SyncExecution(mode="chunk"),
+    fps=30,
+)
```
maxxgx, line 22: This example assumes the camera is part of the robot class, which is conceptually fine since it simplifies the example. However, a slightly more precise example would be:
```python
from physicalai.inference import InferenceModel
from physicalai.runtime import PolicyRuntime, SyncExecution
from physicalai.capture import UVCCamera
from physicalai.robot.so101 import SO101

model = InferenceModel.load("./exports/act_policy")
robot = SO101(port="/dev/ttyACM0")
cameras = {"top": UVCCamera(device=0), "gripper": UVCCamera(device=1)}

runtime = PolicyRuntime(
    robot=robot,
    cameras=cameras,
    model=model,
    execution=SyncExecution(mode="chunk"),
    fps=30,
)
```

physicalai/docs/design/components/runtime-system/design_review_summary.md
```diff
@@ -0,0 +1,118 @@
...
+from physicalai.robot.so101 import SO101
+
+model = InferenceModel.load("./exports/act_policy")
+robot = SO101(port="/dev/ttyACM0")
+
+runtime = PolicyRuntime(
+    robot=robot,
+    model=model,
+    execution=SyncExecution(mode="chunk"),
+    fps=30,
```
maxxgx, line 21: I wonder if `control_hz` might be clearer than `fps` here? FPS could be interpreted as applying only to camera reading, or at least that was my initial thought.

physicalai/docs/design/components/runtime-system/policy_runtime_design.md
```diff
@@ -0,0 +1,403 @@
+# Policy Runtime Design
+
+> **Open discussion (temporary, remove after team sync):**
```
maxxgx: I think we can remove this section

maxxgx (dismissed):
Nice design, thanks!

A couple of comments, but I think we can update the design as implementation progresses since it's unreasonable to get every detail right from the start

physicalai/docs/design/components/runtime-system/design_review_summary.md
samet-akcay: actually we have the full api later in the design document. But the example is missing cameras. maybe I should have placed it at the top 
https://github.com/samet-akcay/physical-ai-studio/blob/robot-inference/physicalai/docs/design/components/runtime-system/policy_runtime_design.md#5-core-runtime-api

physicalai/docs/design/components/runtime-system/policy_runtime_design.md
samet-akcay: yes, we aligned on this today will remove

physicalai/docs/design/components/runtime-system/design_review_summary.md
samet-akcay: I'm fine with that , we could rename it to `control_hz` or `control_frequency` or better.

maxxgx (approved):

PR #566, opened 2026-05-07
fix: don't use form around model picker

MarkRedeman:
Having a form makes it so that when the user picks a model this gets immediately submitted. This was fixed before but got reverted due to a merge conflict.

## Type of Change

- [x] 🐞 `fix` - Bug fix


MarkRedeman:
Note: I will check as well if we can fix the `Card` component so that it doesn't auto submit a form when being clicked on.

RHeckerIntel (approved):

PR #568, opened 2026-05-08
refactor: replace empty_cameras with explicit image_features in Pi05 pipeline

eugene123tw:
# Pull Request

## Description

Replace the `empty_cameras` integer counter with an explicit `image_features` list across the Pi05 pipeline. This aligns PAS image handling with lerobot's present/missing camera pattern: cameras not in the batch are filled with `-1` images and zero masks based on the known feature list, rather than a hardcoded count.

Also fixes several bugs in the HF pretrained loading path where image features with IDENTITY normalization were silently dropped from `_dataset_stats`, and where `sample_input` produced malformed keys like `"images.observation.images.image"`.

## Type of Change

- [ ] ✨ `feat` - New feature
- [x] 🐞 `fix` - Bug fix
- [x] 🧪 `test` - Tests
- [ ] 🔧 `chore` - Maintenance

## Breaking Changes

- `empty_cameras` parameter removed from `Pi05Config`, `Pi05` policy, `Pi05Preprocessor` (library), `make_pi05_preprocessors`, and inference `Pi05Preprocessor`.
- Inference `Pi05Preprocessor` now requires `image_features` as a keyword-only argument (no default).
- Export manifests generated before this change that reference `empty_cameras` will need re-export.

---

### Core: `empty_cameras` → `image_features`

**`library/src/physicalai/policies/pi05/config.py`**
- Removed `empty_cameras: int = 0` field.

**`library/src/physicalai/policies/pi05/policy.py`**
- Removed `empty_cameras` from `__init__`, `_initialize_model`, and `_update_preprocessor_stats`.
- Added inline logic in `extra_export_args` to derive `image_feature_keys` from `_dataset_stats` and pass them to the `ComponentSpec(type="pi05", image_features=...)`.

**`library/src/physicalai/policies/pi05/preprocessor.py`**
- Removed `empty_cameras` from `Pi05Preprocessor.__init__` and `make_pi05_preprocessors`.
- Added `self.image_features` list cached from the `features` dict (filtered by `FeatureType.VISUAL`), with `"images."` prefix normalization.
- `_preprocess_images` now uses `present_img_keys` / `missing_img_keys` derived from `self.image_features` instead of `Observation.get_flattened_keys`.
- Missing cameras filled with `-1` images and zero masks after stacking.
- Added `FeatureType.VISUAL → NormalizationType.IDENTITY` to `_norm_map_for_mode`.
- `make_pi05_preprocessors` now recognizes `VISUAL` features from `stat["type"]`.

**`physicalai/src/physicalai/inference/preprocessors/pi05.py`**
- Replaced `empty_cameras: int = 0` with `image_features: list[str]` (required, keyword-only).
- Removed all fallback image discovery logic (dict-nested, `startswith(IMAGES)` scan).
- `_preprocess_images` iterates `self._image_features` in order, simple flat-key lookup.
- Added flatten step at the top of `__call__` to handle nested dict inputs (`{"images": {"cam": arr}}` → `{"images.cam": arr}`).
- Removed `_get_present_image_keys` helper.

### Bug fixes

**`library/src/physicalai/policies/pi05/model.py`** — `sample_input` property
- Fixed `KeyError` crash when `_dataset_stats` entries lack a `"type"` key (HF pretrained path vs Lightning checkpoint).
- Added `_is_image_feature` helper that checks `IMAGES in key` or `.get("type")` safely.
- Fixed malformed keys: `"images.observation.images.image"` → `"images.image"` by stripping `"observation."` prefix from feature names.

**`library/src/physicalai/policies/pi05/pretrained_utils.py`**
- `parse_preprocessor_stats`: Extracts `"type"` from `step["config"]["features"]` and propagates it to stat entries. New `_collect_identity_features` adds image features with IDENTITY normalization that have no safetensor stats.
- `parse_config_features` (fallback): Now includes image features (not just state/action) with their type.

**`library/src/physicalai/data/lerobot/utils/quantile_stats.py`**
- `augment_dataset_quantile_stats` now adds full stats for keys computed by LeRobot but missing from `meta.stats` (e.g. image observations omitted by older datasets).

**`library/src/physicalai/data/lerobot/datamodule.py`**
- Stats propagation to train/val adapters now adds missing keys instead of silently skipping them.

### Tests

**`physicalai/tests/unit/inference/preprocessors/test_pi05.py`**
- All tests updated to pass `image_features` (now required).
- `test_empty_cameras_appended` → `test_missing_image_features_filled` (uses `image_features` list).
- Added `test_no_image_features_raises` (TypeError on missing kwarg).
- `test_dict_images_stacked` → `test_flat_key_images_stacked` (dict-nested no longer the primary API).
- Input helper `_make_inputs` uses flat keys (`"images.image"`) matching `image_features`.

**`library/tests/unit/policies/test_pi05.py`**
- Expanded test coverage for preprocessor, pretrained utils, and multi-camera scenarios.

**`library/tests/unit/data/test_quantile_stats.py`**
- Added `test_adds_stats_for_keys_missing_from_meta`.

## Examples

Before (hardcoded empty camera count):
```python
# Training config
policy = Pi05(empty_cameras=1, image_resolution=(224, 224))

# Inference preprocessor
prep = Pi05Preprocessor(image_resolution=(224, 224), empty_cameras=1)
```

After (explicit image features, missing cameras derived automatically):
```python
# Training — empty_cameras removed, image_features derived from dataset stats
policy = Pi05(image_resolution=(224, 224))

# Inference — image_features required, missing cameras filled automatically
prep = Pi05Preprocessor(
    image_resolution=(224, 224),
    image_features=["images.image", "images.image2", "images.empty_camera_0"],
)
```


library/src/physicalai/policies/pi05/model.py
```diff
@@ -676,20 +676,34 @@ def sample_input(self) -> dict[str, torch.Tensor | str]:
...
         sample_input = {}
 
-        num_image_features = sum(
-            1 for key in self._dataset_stats if str(FeatureType.VISUAL) in self._dataset_stats[key]["type"]
-        )
+        def _is_image_feature(key: str, stats: dict[str, Any]) -> bool:
+            if IMAGES in key:
+                return True
+            feat_type = stats.get("type")
+            return feat_type is not None and str(FeatureType.VISUAL) in str(feat_type)
```
sovrasov: Are cases where feature type is not provided possible?

library/src/physicalai/policies/pi05/policy.py
```diff
@@ -680,6 +675,16 @@ def extra_export_args(self) -> dict[str, ExportParameters]:
             )
             raise ValueError(msg)
 
+        # Map "observation.images.image" → "images.image" to match preprocessor batch keys
+        image_feature_keys = []
+        for key, stat in self._dataset_stats.items():
+            if stat.get("type") != FeatureType.VISUAL.value:
+                continue
+            name = str(stat.get("name", key)).rsplit("observation.", maxsplit=1)[-1]
```
sovrasov: "observation" is lerobot stuff should not leak to "name", we have to clear it once we acquire stats instead of cleaning every time we access stats

library/src/physicalai/policies/pi05/preprocessor.py
```diff
@@ -162,9 +161,19 @@ def __init__(
...
         self._tokenizer = None
-        self.empty_cameras = empty_cameras
         self.normalization_mode = normalization_mode
 
+        # Cache image feature keys from the features dict, ensuring "images." prefix
+        self.image_features = (
+            [
+                k if k.startswith(IMAGES) else f"{IMAGES}.{k}"
+                for k, f in features.items()
+                if f.ftype == FeatureType.VISUAL
```
sovrasov, line 171: Same here: if we cleanup the feature name and guarantee type is always specified on `stats` acquisition, we can avoid extra checks

sovrasov (commented):
Thanks Eugene, that way of handling empty cameras is definitely more clear

library/src/physicalai/policies/pi05/model.py
eugene123tw: @sovrasov  `"type"` is missing for dataset metadata entries like `episode_index`, `frame_index`, `timestamp`, `index`, and `task_index`.** These appear in `_dataset_stats` but aren't real features, just bookkeeping fields from the LeRobot dataset format and have no `"type"` key. The `.get("type")` guard prevents a `KeyError` when iterating over those entries.

Also, on the HF pretrained path (`_from_hf` → `parse_preprocessor_stats`), the `"type"` field is only present if the normalizer config included feature metadata. Older pretrained checkpoints or configs that predate the `"type"` field would also lack it.

The first check (`IMAGES in key`) is the primary path for image detection; the `feat_type` check is a fallback for edge cases where the key name doesn't contain `"images"` but the feature is still typed as `VISUAL`.

PR #569, opened 2026-05-08
feat(config): unify config instantiation for API and CLI parity

samet-akcay:
## Problem

We plan to move `physicalai/` into a dedicated inference/runtime repo. Once that happens, the runtime package should not depend on `torch` or Lightning, which means we will not be able to reuse the existing `LightningCLI`-based CLI there.

At the same time, we still want strong API/CLI parity:

- the same object should be constructible from Python, YAML, JSON, dataclasses, or Pydantic models
- the same config shape should work in both Python APIs and a future `jsonargparse`-based CLI
- config authors should be able to choose between typed config objects (`ModelConfig`, `PolicyConfig`) and dynamic `class_path` / `init_args` configs depending on the use case

The cleanest way to get there is to standardize instantiation now, so both training and future runtime CLIs can build on the same config model.

## Solution

This PR unifies config-driven object construction around a single generic backend:

```python
instantiate_obj(config, target_cls=...)
```

Behavior:

- if `config` contains `class_path`, import and instantiate that class
- otherwise, if `target_cls` is provided, call `target_cls(**config)`
- recursively instantiate nested `class_path` configs inside dicts, lists, and tuples

On top of that shared backend, this PR:

- adds `target_cls` support to `instantiate_obj(...)` and related helpers
- simplifies `FromConfig` into a thin wrapper over the generic instantiation utilities
- exports the new `@from_config` decorator
- extends `dataclass_to_dict(..., recursive=False)` so shallow dataclass conversion is supported
- refreshes config docs to match the actual behavior and remove stale claims

This is partly a refactor internally, but it is also user-facing work: it adds missing config/instantiation capability that we will need to support a future self-written, torch-free `jsonargparse` CLI while preserving a consistent Python API.

### Typed config objects

Dataclass configs remain a good authoring experience for explicit typed APIs:

```python
@dataclass
class ModelConfig(Config):
    hidden_size: int = 256
    num_layers: int = 3

model = Model.from_config(ModelConfig())
```

Likewise for Pydantic:

```python
class PolicyConfig(BaseModel):
    optimizer_lr: float = 1e-4
    n_action_steps: int = 50

policy = Policy.from_config(PolicyConfig())
```

This supports patterns like `ModelConfig` vs `Model` and `PolicyConfig` vs `Policy`.

### Dynamic CLI-friendly configs

For CLI and interchange use cases, the canonical dynamic shape is still:

```yaml
class_path: mypkg.policy.Policy
init_args:
  model:
    class_path: mypkg.model.Model
    init_args:
      hidden_size: 256
  optimizer_lr: 1e-4
```

That same config can be used from Python:

```python
policy = Policy.from_config("policy.yaml")
```

and is also the shape we want a future `jsonargparse`-based runtime CLI to consume:

```bash
physicalai run --config policy.yaml
```

### API / CLI parity

The same conceptual object can now be created through one API entry point from multiple config sources:

```python
model = Model.from_config({"hidden_size": 256, "num_layers": 3})
model = Model.from_config(ModelConfig(hidden_size=256))
model = Model.from_config(ModelConfigPydantic(hidden_size=256))
model = Model.from_config("model.yaml")
```

Nested construction also works the same way:

```yaml
class_path: mypkg.policy.Policy
init_args:
  model:
    class_path: mypkg.model.Model
    init_args:
      hidden_size: 256
  optimizer_lr: 1e-4
```

```python
policy = Policy.from_config("policy.yaml")
```

That parity is the main goal of this refactor: one config model that works cleanly for both Python APIs and future CLI entry points.


physicalai/pyproject.toml
```diff
@@ -22,6 +22,7 @@ dependencies = [
     "omni_camera @ git+https://github.com/ArendJanKramer/OmniCamera.git@master",
     "pydantic>=2.5",
     "pyyaml>=6.0",
+    "jsonargparse[signatures]>=4.27.0",
```
maxxgx: Shouldn't we update `library/project.toml` instead of `physicalai/project.toml`?

samet-akcay: hm, good catch, we already have it in library. maybe we could remove this for now, and add when we move it?

maxxgx: sounds good

maxxgx (approved):

PR #580, opened 2026-05-12
docs: Add prerequisite and updating sections to README

MarkRedeman:
This PR adds a new prerequisite and updating section to our `application/README.md`. I've also slightly tweaked the main `README.md` so that it's clearer that you need to run the backend and ui in a separate shell.

## Type of Change


Radwan-Ibrahim (approved):
LGTM

MarkRedeman:
Will probably make another PR based on #583 as the current setup might not allow you to compile models (for a training speedup) when training.
There might also be some other prerequisites for exporting to specific formats (OpenVINO, ONNX, Executorch) which I will try to investigate.

PR #583, opened 2026-05-13
fix: fix compile model when using docker

MarkRedeman:
Install build-essential in runtime images so torch.compile/Triton can
compile and link CUDA helper modules during training.

This fixes training failures in containers that reported 'Failed to find
C compiler' and avoids secondary DataLoader worker aborts.

Limit CUDA cleanup to /usr/local/cuda* static libs so system archives
like libc_nonshared.a remain available for torch.compile/Triton runtime
linking in CUDA containers.

## Type of Change

- [x] 🐞 `fix` - Bug fix

### DIFF
--- a/application/docker/Dockerfile
+++ b/application/docker/Dockerfile
@@ -169,7 +169,9 @@ RUN groupadd --gid "${APP_GID}" appuser \
     && useradd --uid "${APP_UID}" --gid "${APP_GID}" --create-home appuser
 
 # Common runtime system dependencies (OpenCV, USB access, video decoding)
+# Includes runtime toolchain/linker deps required by torch.compile (Inductor/Triton)
 RUN apt-get update && apt-get install -y --no-install-recommends \
+        build-essential \
         ffmpeg \
         libgl1 \
         libglib2.0-0 \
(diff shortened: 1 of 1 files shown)


application/docker/Dockerfile
```diff
@@ -310,7 +314,9 @@ RUN set -eux; \
     find /usr/local/cuda* /usr/lib/x86_64-linux-gnu/ \
         \( -path '*/include' -o -path '*/headers' \) -type d \
         -exec rm -rf {} + 2>/dev/null || true; \
-    find /usr/local/cuda* /usr/lib/x86_64-linux-gnu/ \
+    # Keep system static archives (e.g. libc_nonshared.a) required by
+    # torch.compile/Triton runtime linking; remove only CUDA static libs.
+    find /usr/local/cuda* \
```
MarkRedeman, line 319: @AlexanderBarabanov this is needed as otherwise compilation of models won't work in Docker as the code tries to use `libc_nonshared.a`.
I can try to harden this a bit more, but I'm not sure what our legal requirements are here.

AlexanderBarabanov: Thanks, we will review and address it (if any) in a separate PR.

AlexanderBarabanov (approved):

PR #584, opened 2026-05-13
fix: enforce spawn multiprocessing for CUDA/XPU-safe workers

MarkRedeman:
Add a shared `ensure_spawn_start_method()` helper and invoke it at startup before scheduler workers are created.

This prevents accelerator re-initialization errors in forked subprocesses by guaranteeing the 'spawn' start method for training and worker processes, including non-`__main__` startup paths.

## Type of Change

- [x] 🐞 `fix` - Bug fix


maxxgx (approved):
I don't think we need `spawn` when only CPU is available, but having always the same multiprocessing mode is simpler

PR #585, opened 2026-05-13
feat: lerobot torch inference parity

samet-akcay:
## Summary

- add torch-export and load support for LeRobot wrappers so exported checkpoints retain LeRobot config/state and can be reloaded through `InferenceModel(backend="torch")`
- extend LeRobot wrapper equivalence integration coverage to compare exported torch inference against both the wrapped LeRobot policy and a native LeRobot replay model on the same input
- calibrate deterministic inference checks for stochastic/VLA policies and verify GPU parity locally for `act`, `smolvla`, and `pi05`

## Testing

Integration tests are passing for `ACT`, `SmolVLA` and `Pi05`

- `uv run pytest 'tests/integration/test_lerobot_wrapper_equivalence.py::TestTorchInferenceEquivalence::test_torch_inference_matches_native_lerobot[act]' -q`
- `uv run pytest 'tests/integration/test_lerobot_wrapper_equivalence.py::TestTorchInferenceEquivalence::test_torch_inference_matches_native_lerobot[smolvla]' -q`
- `uv run pytest 'tests/integration/test_lerobot_wrapper_equivalence.py::TestTorchInferenceEquivalence::test_torch_inference_matches_native_lerobot[pi05]' -q`
- `uv run pytest tests/unit/inference/test_adapters.py::TestTorchAdapter tests/unit/policies/test_lerobot.py::TestLeRobotPolicyCheckpoint::test_load_from_checkpoint_universal_wrapper -q`


library/src/physicalai/inference/adapters/pytorch.py
```diff
@@ -81,10 +81,14 @@ def load(self, model_path: Path | str) -> None:
                 raise KeyError(msg)
 
         try:
-            _, class_name = policy_class_path.rsplit(".", 1)
-            policy_class = get_policy_class(class_name)
+            module_name, class_name = policy_class_path.rsplit(".", 1)
+            policy_class = getattr(importlib.import_module(module_name), class_name)
```
sovrasov: Is that import safe? Why don't use a lookup like `get_policy_class`?

samet-akcay: `get_policy_class` only looks up the first party policies via short strings like `act`. here we want to also support lerobot policies. so say metadata points to `physialai.policies.lerobot.aliases.ACT`. if we use get-policy-class, it would resolve it to first party act. 

so this new approach is intentional with the cost of broader dynamic loading. to keep it tighter i've refactored `import_class` and added a callable check

alfieroddan (commented):
@sovrasov seems as though Pi0(.5) are still blocked for me.

Can you test these work for you? Otherwise all done I think.

```bash
uv run pytest 'tests/integration/test_lerobot_wrapper_equivalence.py::TestTorchInferenceEquivalence::test_torch_inference_matches_native_lerobot[pi0]' -q

uv run pytest 'tests/integration/test_lerobot_wrapper_equivalence.py::TestTorchInferenceEquivalence::test_torch_inference_matches_native_lerobot[pi05]' -q
```

samet-akcay:
will work on a new pr after v0.2.0 release that follows https://github.com/open-edge-platform/physical-ai-studio/pull/1040

PR #591, opened 2026-05-14
chore: harden uv package resolution with exclude-newer

MarkRedeman:
Similar to #582 set `exclude-newer` for our python packages to minimize risk of supply chain attacks.

See the [uv docs](https://docs.astral.sh/uv/concepts/indexes/#configuring-exclude-newer-for-an-index) for more info.

## Type of Change

- [x] 🔧 `chore` - Maintenance


MarkRedeman:
For reference the "stabilize library uv lock resolution for xpu extras" commit, be55287230d4aac5f2983766b399f0af7a20a525, is based on the following findings from opencode:
```
# Fix `uv lock` in `library/` (`physicalai-train`)

## Goal

Make `uv lock` succeed reliably in `library/` while preserving the existing dependency policy.

## Plan

1. Align `library/pyproject.toml` index strategy with the known-working backend setup.
   - Add a named PyPI index (`https://pypi.org/simple`) before PyTorch indexes.
   - Change `pytorch-xpu` index to `explicit = false`.
   - Set `[tool.uv] index-strategy = "unsafe-best-match"`.

2. Make XPU transitive source routing explicit.
   - Add `triton-xpu = { index = "pytorch-xpu" }` under `[tool.uv.sources]`.
     (Using `extra = "xpu"` on this source is invalid unless `triton-xpu` is explicitly listed in `project.optional-dependencies.xpu`.)
   - Keep existing `pytorch-triton-xpu` mapping.

3. Validate resolution with a dry-run first.
   - Run `uv lock --dry-run --python 3.12 -v` in `library/`.

4. If dry-run still fails due cutoff/upload-time issues, apply targeted fallback.
   - Add `[tool.uv] exclude-newer-package` exemptions for XPU-only packages, starting with:
     - `triton-xpu = false`
     - `pytorch-triton-xpu = false`
   - Re-run dry-run.

5. If failure remains only on unsupported universal branches, constrain environments.
   - Add `[tool.uv] environments = ["sys_platform == 'linux' or sys_platform == 'darwin' or sys_platform == 'win32'"]`.
   - Re-run dry-run.

6. Finalize lockfile.
   - Run `uv lock` in `library/`.
   - Verify lock includes XPU packages from XPU index and mixed-source dependencies resolve correctly.

## Why this is still a workaround

The local config changes are safe, but they are not the cleanest long-term outcome:

- `index-strategy = "unsafe-best-match"` is needed because package names can appear on both PyPI and the PyTorch XPU index, and the default `first-index` behavior can stop too early on an incomplete source.
- `pytorch-xpu` had to be non-explicit so transitive XPU-only dependencies can still be discovered during resolution.
- We also needed explicit source routing for transitive packages (for example `triton-xpu`), which ideally should not be required by every downstream project.

## Preferred upstream fix (PyTorch/XPU side)

To remove this local "hack" in consumers, the PyTorch XPU packaging/index side should make resolution deterministic under default resolver settings (`first-index`, explicit secondary index):

1. **Use one stable dependency name for Triton XPU across releases**
   - Avoid switching between `pytorch-triton-xpu` and `triton-xpu` in `torch+XPU` dependency metadata.
   - Pick one canonical package name and keep it consistent for all supported `torch+XPU` versions.

2. **Make canonical XPU packages fully available on the primary index path**
   - Either publish canonical XPU transitive packages on PyPI at matching versions, or
   - ensure dependency names used by `torch+XPU` do not collide with partially populated PyPI packages.
   - This avoids "first index has package name but not compatible versions" dead-ends.

3. **Provide complete index metadata for time-based pinning**
   - Ensure all XPU wheel links expose reliable upload timestamps/metadata in the simple index.
   - This is required for deterministic behavior with policies like `exclude-newer`.

4. **Tighten platform markers in upstream wheel metadata**
   - Keep XPU-only transitive requirements scoped to actually supported platforms.
   - Avoid broad marker branches that force resolver exploration of unsupported targets.

## What we can revert once upstream is fixed

If the upstream fixes above land, this project should be able to simplify back to default behavior:

- revert to `index-strategy = "first-index"`
- set `pytorch-xpu` back to `explicit = true`
- remove the extra transitive source override for `triton-xpu`
- keep only direct package source mapping for `torch` / `torchvision` extras
```

This is the same issue we also saw in the `application/backend/pyproject.toml` setting.

AlexanderBarabanov (approved):

PR #594, opened 2026-05-14
refactor: move robot adapter logic into physical ai implementation

MarkRedeman:
# Unify Robot Adapter PR

## Summary

This PR unifies the backend robot adapter path for `physicalai` robots and moves unit conversion logic into the `physicalai` robot implementations so direct library users can use the same model I/O units as the application.

The end result is:

- one backend adapter implementation (`PhysicalAIRobotAdapter`) for both SO101 and WidowXAI,
- robot-specific unit semantics owned by the `physicalai` drivers,
- removed duplicated conversion logic from backend adapters,

## Why

The main goal is to make model outputs from the application usable directly with `physicalai` without users needing manual radian/degree/normalized mapping.

Previously, the backend adapters did much of this mapping (especially SO101 and WidowXAI), which meant direct `physicalai` usage did not match app-facing units.

Example: users could record datasets and train/export models in the application where non-gripper joint values were represented in degrees, while direct `physicalai` robot APIs expected radians.
After export, those models still expected degree-based inputs/outputs, so users could not use the library directly without adding their own conversion layer around inference.

## Old vs New behavior 

- **Old**
  - backend adapters owned most unit conversion,
  - direct `physicalai` users saw driver-native units (e.g., radians) and had to map manually.

- **New**
  - library drivers own conversion to app-facing units,
  - backend adapter is largely transport/event mapping,
  - direct `physicalai` usage is much closer to app model I/O conventions.

## Known caveats

I have not been able to test the WidowXAI integration. I expect most to work but we should do some actual checks by comparing the action states before/after this PR. 

## Type of Change

- [x] 🐞 `fix` - Bug fix

## Breaking Changes

The library now uses degrees as the primary unit output for robot actions.


ArendJanKramer:
if we merge from mark/robot-adapter-with-bimanual it's rebased, and tested

MarkRedeman:
> if we merge from mark/robot-adapter-with-bimanual it's rebased, and tested

Yes, I will rebase #530 once this is merged, and update it to also use the same adapter implementation. That will also make the implementation on that PR simpler.

maxxgx (approved):
Works like a charm. Thanks, Mark!

PR #595, opened 2026-05-15
feat: pi05 batch all camera images into single vision encoder call

eugene123tw:
## Description

- **Eval / tracing**: batches all camera images into one encoder call (`[N*B, C, H, W]`), avoiding redundant sequential vision encoder invocations.
- **Training**: iterates per-camera with gradient checkpointing support.

## Type of Change

- [x] ✨ `feat` - New feature
- [ ] 🐞 `fix` - Bug fix
- [ ] 🧪 `test` - Tests
- [ ] 🔧 `chore` - Maintenance

## Changes Made

| File | Change |
|------|--------|
| `library/src/physicalai/policies/pi05/model.py` | Add images batching; added `.bool()` cast in `_prepare_attention_masks_4d` for JIT tracing compatibility |
| `library/tests/unit/policies/test_pi05.py` | Added `TestEmbedPrefix` class with 8 unit tests covering shapes, encoder call count, mask behavior, and train/eval parity |

## Examples

```python
def embed_prefix(self, ...):
    use_batched = not self.training or torch.jit.is_tracing() or torch.onnx.is_in_onnx_export()
    if use_batched:
        # single batched encoder call
    for cam_idx in range(num_cameras):
        if use_batched:
            img_emb = all_img_embs[cam_idx]
        else:
            img_emb = self._apply_checkpoint(image_embed_func, images[cam_idx])
    ...
```

### DIFF
--- a/library/src/physicalai/policies/pi05/model.py
+++ b/library/src/physicalai/policies/pi05/model.py
@@ -836,7 +836,7 @@ class Pi05Model(ExportableModelMixin, Model):
         Returns:
             Tuple of (embeddings, padding masks, attention masks).
         """
-        use_batched = not self.training or torch.jit.is_tracing() or torch.onnx.is_in_onnx_export()
+        use_batched = not self.training
 
         num_cameras = images.shape[0]
         bsize = images.shape[1]
--- a/library/src/physicalai/policies/pi05/model.py
+++ b/library/src/physicalai/policies/pi05/model.py
@@ -774,7 +774,9 @@ class Pi05Model(ExportableModelMixin, Model):
         Returns:
             4D attention mask tensor.
         """
-        att_2d_masks_4d = att_2d_masks[:, None, :, :]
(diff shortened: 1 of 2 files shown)


library/src/physicalai/policies/pi05/model.py
```diff
@@ -821,45 +823,65 @@ def embed_prefix(
...
+            tokens: ``(batch, seq_len)`` tokenized prompt.
+            masks: ``(batch, seq_len)`` prompt attention mask.
+
         Returns:
             Tuple of (embeddings, padding masks, attention masks).
         """
-        embs = []
-        pad_masks = []
-        att_masks = []
+        use_batched = not self.training or torch.jit.is_tracing() or torch.onnx.is_in_onnx_export()
```
sovrasov: The latter checks might be a bit extra, because we don't export in training mode: that would result in exporting loss computation

sovrasov (approved):

PR #598, opened 2026-05-15
refactor: introduce Cursor for select_action() and cleanup Runner abstraction

sovrasov:
# Pull Request

## Description

- replace `ActionChunking` runner with Cursor for implementing `select_action` behavior
- `select_action` and `predict_action_chunk` are now always available
- Add chunks trimming postprocessor to algin with python policy usage
- Fix inference call in backend: instead of `SinglePass` runner + `select_action` the right way is `predict_action_chunk`

## Type of Change

## Breaking Changes

`InferenceModel.use_action_queue` is removed as well as `ActionChunking` runner. The latter was not used by default, so default exported models are not affected.

## Examples
```python
model = InferenceModel(...)
model.select_action(observation) # [bs,act_dim]
model.predict_action_chunk(observation) # [bs,chunk_size,act_dim]
```


physicalai/docs/design/components/inferencekit.md
```diff
@@ -845,30 +845,13 @@ from inferencekit import InferenceModel
...
-    Policies output action chunks (multiple future actions).
-    This runner queues them and dispenses one action per call.
-    """
-
-    def __init__(self, chunk_size: int = 16, n_action_steps: int = 1):
-        self.chunk_size = chunk_size
-        self.n_action_steps = n_action_steps
-        self._action_queue = []
+# Policy-specific runner example
+class PolicyRunner(InferenceRunner):
```
samet-akcay, line 849: Can you elaborate why `PolicyRunner` is neede on top of `InferenceRunner`?

sovrasov: `InferenceRunner` is an abstract class, which means we can not directly create it's instance. It's hard to say in advance how many sub-classes of `InferenceRunner` we might need, but for sure there's a space for customization, and the document reflects that without elaborating why do we need it

physicalai/src/physicalai/inference/utils/action_cursor.py
samet-akcay, line 1: the current name `Cursor` would work once a reader has read the docstring, but a cold reader won't immediately know what kind of cursor this is or what it operates? Especially with the following path: `physicalai/src/physicalai/inference/utils/cursor.py`?

The design doc refers to it `ActionChunkCursor`. It is descriptive, but a bit verbose. 

Others would perhaps be: 
- `ChunkCursor`
- `ActionCursor`
- or maybe something else? Thoughts?

samet-akcay (dismissed):
I've approved the pr, but added a minor comment about the naming of `Cursor` class?

physicalai/src/physicalai/inference/utils/action_cursor.py
sovrasov: Renamed to `ActionCursor`

samet-akcay (approved):
thanks for the effort!

PR #606, opened 2026-05-19
feat: import model directory from cli

MarkRedeman:
This PR allows users to import their models from a directory into Physcial AI Studio. The use case for this is that you might train a model on an external system and want to test it out on your local system.
I will be adding a follow up PR that also allows the user to upload a model via a zip using the UI, but this is likely to be a lot slower than the cli approach as you'd need to zip + upload the zip via the browser. 
Assuming that a user has ssh access to the server (or desktop) that runs that studio then the cli approach will be faster.

Example import command:
```bash
uv run src/cli.py models import-dir --source-dir=/data/storage/models/pi05-dice-cleanup/ \
    --project-id="d8b99bd2-5730-48f2-8873-a45960e8b5c0"\
    --dataset-id="a245ca5b-a410-49ea-8741-fd1ad3a4f575" \
    --model-name="Pi0.5 Cleanup Dice" \
    --copy

Importing model from folder: /data/storage/models/pi05-dice-cleanup
Mode: copy
Model imported successfully!
Model ID: b3b394f5-5101-4a54-bf5b-4faf6ef9eb57
Model path: /home/mark/.cache/physicalai/models/b3b394f5-5101-4a54-bf5b-4faf6ef9eb57
```

## Type of Change

- [x] ✨ `feat` - New feature


MarkRedeman:
At the moment this command requires the user to manually provide the project and dataset ids which is a bit annoying. You can copy them based on the url you see when using the UI to look at your dataset, but it's not an ideal UX.

I'm considering making this command more interactive where the CLI could get all available projects, let the user select one of them, then show all datasets and select it etc.
But this is likely to make the cli script more complicated. I'm not sure if that's worth it?

application/backend/src/services/model_import_service.py
```diff
@@ -0,0 +1,226 @@
...
+        if not file_path.is_file():
+            return None
+        try:
+            with file_path.open(encoding="utf-8") as fobj:
+                data = json.load(fobj)
+        except (OSError, ValueError):
+            return None
+        if isinstance(data, dict):
+            return data
+        return None
```
MarkRedeman, line 61: This bit of extra abstractions makes it so that later we can also import based on a `SafeZipArchive` which was introduced by the dataset import PR.

application/backend/src/services/model_import_service.py
```diff
@@ -0,0 +1,226 @@
...
+from services.dataset_service import DatasetService
+from services.job_service import JobService
+from services.model_service import ModelService
+from services.snapshot_service import SnapshotService
+from settings import get_settings
+
+# We assume the directory/zip is taken directly from Physical AI Studio, either
+# by exporting the model from the UI, or by taking it from our storage dir
+_REQUIRED_FILES = (
+    "version_0/hparams.yaml",
```
maxxgx, line 21: Is the prefix always `version_0`?

application/backend/src/services/model_import_service.py
```diff
@@ -0,0 +1,226 @@
...
+        policy: str,
+        base_model_id: UUID | None,
+        version: int,
+    ) -> Model:
+        """Create snapshot, job, and model record after files are in place."""
+        project_id = dataset.project_id
+        dataset_id = dataset.id
+
+        snapshot_dir = model_dir / SnapshotService.generate_snapshot_folder_name()
+        snapshot = await SnapshotService.create_snapshot_for_dataset(dataset, destination=snapshot_dir)
```
maxxgx: When we import a model, I think we don't have the dataset that was used to train it. So assigning a dataset and snapshot to it anyway seems a bit forced because of the backend constraints. 
Maybe we can assign an empty dataset/snapshot?

application/backend/src/services/model_import_service.py
```diff
@@ -0,0 +1,226 @@
...
+
+# We assume the directory/zip is taken directly from Physical AI Studio, either
+# by exporting the model from the UI, or by taking it from our storage dir
+_REQUIRED_FILES = (
+    "version_0/hparams.yaml",
+    "version_0/metrics.csv",
+    "exports/torch/manifest.json",
+)
+_TORCH_MANIFEST_PATH = "exports/torch/manifest.json"
+_SUPPORTED_POLICIES = frozenset({"act", "smolvla", "pi05"})
```
maxxgx, line 26: I wonder if we can fetch the supported policies from library of somewhere else instead of defining a constant here, as that could reduce the number of places to update in the future.

Also, shouldn't it also include `pi0`?

MarkRedeman: At the moment I was mainly thinking of restricting to these because these are expected to work. There's also  `groot` but like `pi0` I think that hasn't been tested.

application/backend/src/services/model_import_service.py
MarkRedeman: I think it's a good idea, but it might also complicate things: at the moment when you click "Run model" in the UI we fetch the environment that is associated to the recorded dataset and then use that to setup the inference screen.
If we were to add an empty dataset then we'd also need to add an empty environment and loading inference then won't work.

My assumption was that a user would be recording their dataset on their local device, train Pi0.5 on their server, export it and then import to their local device using the dataset that they recorded.

This is UX that we need to fix: me and Ronald had been talking about that we probably need some explicit mappings of "Environment <-> Dataset" (when recording a dataset), "Dataset <-> Model" (when training a model) and "Environment <-> Model" (when running inference).
This will be a bit more work though as atm this is tightly coupled.

application/backend/src/services/model_import_service.py
MarkRedeman: It is when the model was trained by Physical AI Studio. This won't be the case when your model is trained by lerobot or by the physicalai train library.
I would like us to also support importing models from other sources (especially importing lerobot models from huggingface would be great) but that requires some more effort.

application/backend/src/services/model_import_service.py
MarkRedeman: Actually for the snapshot: we do allow a `None` value, so I think the better solution here might be to set it to None as generating it from the provided dataset is likely to be incorrect (the dataset might have been updated).

@RHeckerIntel what do you think?

maxxgx (dismissed):

maxxgx (approved):

PR #613, opened 2026-05-19
feat: Improve teleoperation performance

RHeckerIntel:
# Rework robot control worker so that it better handles robot control.

## Motivation
All actions when controlling robots were happening in the same loop. This limits different frequencies for different processes. The camera can run at 30hz, the robot teleoperates at 100hz, and the recording might be at 50hz. 
This was not possible.

Secondly, by having it all in one loop any hiccup would tank performance.

## Core changes

Moved the cameras, teleoperation and recording to their own workers.
They share their data using shared memory.
One orchestrator  handles communication between them all using a data manifest.
This data manifest contains all the info needed to build observations from  shared memory.

## Secondary changes
 
- Removed a lot of layers, transporters etc.
- Made some functions in physical ai robot client adapter sync. We might make more sync.
- Moved torch imports inline so server starts quick.
- Viewing your environment now automatically tests teleoperation too.


MarkRedeman:
Few notes/findings from this branch:
- Teleoperation is a LOT smoother. Before I had jitters when using the SO101 robot, this is gone now.
- We should add a toggle or some control form that lets us enable/disable teleoperation in the environment page. This way we don't immediately trigger teleoperation when going to the environments page
- I had an issue just now where I was getting many `Device or resource busy (os error 16) mjpeg 640x480@30fps` errors which did not seem to stop when I closed the browser. I'm not sure yet how to reproduce or if this is specific to this branch, it seemed to happen when configuring a new environment with 2 cameras.

application/backend/src/control/environment_integration.py
```diff
@@ -1,194 +1,82 @@
...
+                self._workers.append(worker)
+                camera_entries.append(
+                    CameraManifestEntry(
+                        id=str(camera.id),
+                        name=camera.name,
+                        width=worker._width,
+                        height=worker._height,
+                        frame_data=worker._frame_data,
+                    )
+                )
```
MarkRedeman, line 65: Would it be possible to do this after the workers have started?

I believe that this is the reason that right now on startup you first see a black screen as the camera output before seeing the actual feed.

PR #614, opened 2026-05-19
fix: chunk truncation handling in torch

sovrasov:
# Pull Request

## Description

The PR fixes overall behavior of `select_action` in case if chunk truncation is specified in policy configs.
- base policy doesn't drop the beginning of the chunk when it doesn't fit into action queue
- first-party policies add chunk trimmer to the manifest to allow `InferenceModel` manage trimming on it's end, as it's done for other inference backends

## Type of Change

- [x] 🐞 `fix` - Bug fix

### DIFF
--- a/library/src/physicalai/export/mixin_policy.py
+++ b/library/src/physicalai/export/mixin_policy.py
@@ -31,6 +31,7 @@ from physicalai.export.backends import (
     ExportParameters,
     ONNXExportParameters,
     OpenVINOExportParameters,
+    TorchExportParameters,
 )
 from physicalai.train import __version__
 
@@ -198,6 +199,11 @@ class ExportablePolicyMixin:
         model_path = self._prepare_export_path(checkpoint_path, ".pt")
         export_dir = model_path.parent
 
+        extra_model_args: TorchExportParameters = cast(
+            "TorchExportParameters",
+            self._get_export_extra_args(ExportBackend.TORCH),
+        )
+
         checkpoint = {}
         checkpoint["state_dict"] = self.state_dict() if hasattr(self, "state_dict") else {}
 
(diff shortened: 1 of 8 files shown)


library/tests/unit/inference/test_adapters.py
```diff
@@ -42,10 +43,18 @@ class TestTorchAdapter:
     @staticmethod
     def _write_policy_metadata(tmp_path: Path) -> Path:
```
alfieroddan: `_write_policy_manifest` now?

alfieroddan (commented):
Looks good, thanks!

alfieroddan (dismissed):
Same comment as above otherwise gtg.

library/src/physicalai/export/mixin_policy.py
```diff
@@ -213,7 +219,12 @@ def to_torch(self, checkpoint_path: PathLike | str) -> None:
         torch.save(checkpoint, str(model_path))  # nosec B614
 
         # Create metadata files
-        self._create_metadata(export_dir, ExportBackend.TORCH)
+        self._create_metadata(
```
samet-akcay, line 222: are you removing metadata in this pr, or will it be a separate one ?

library/tests/unit/inference/test_adapters.py
sovrasov: Updated

library/src/physicalai/export/mixin_policy.py
sovrasov: Metadata is to be completely removed in https://github.com/open-edge-platform/physical-ai-studio/pull/617
This PR sometimes skips it's reding (in TorchAdapter)

samet-akcay: ah right, thanks for the clarification

samet-akcay (approved):

PR #625, opened 2026-05-25
chore: make openvino the default selected inference backend when available

MarkRedeman:
This PR contains a few minor refactors that I made while preparing a PR for displaying model exports in the models screen.
- UI: If available make OpenVINO the default selected inference runtime, otherwise select the first available backend
- UI: Extract `backendConfigs` into a separate file (to be reused by model exports table)
- UI: Add a "Coming soon" illustration instead of empty heading placeholder
- backend: don't compress archives when downloading datasets or models (results in faster downloads, since we deal with binary data we can expect that compression won't be very effective)
- backend: offload archiving to event loop. Before when downloading a model or dataset this would make the server unresponsive until archiving was completed, this should fix that

## Type of Change

- [x] 🔧 `chore` - Maintenance

## Screenshots

image

### DIFF
--- a/application/backend/src/services/model_download_service.py
+++ b/application/backend/src/services/model_download_service.py
@@ -30,7 +30,7 @@ class ModelDownloadService:
 
         temporary_archive_path = Path(tempfile.gettempdir()) / f"model-{uuid4()}.zip"
 
-        with zipfile.ZipFile(temporary_archive_path, mode="w", compression=zipfile.ZIP_DEFLATED) as archive:
+        with zipfile.ZipFile(temporary_archive_path, mode="w", compression=zipfile.ZIP_STORED) as archive:
             for path in model_path.rglob("*"):
                 if not path.is_file():
                     continue
--- a/application/backend/src/api/dataset.py
+++ b/application/backend/src/api/dataset.py
@@ -1,3 +1,4 @@
+import asyncio
 from pathlib import Path
 from typing import Annotated
 from uuid import UUID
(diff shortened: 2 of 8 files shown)


application/backend/src/services/model_download_service.py
```diff
@@ -30,7 +30,7 @@ def create_model_archive(self, model_path: Path, *, include_snapshot: bool = Fal
 
         temporary_archive_path = Path(tempfile.gettempdir()) / f"model-{uuid4()}.zip"
 
-        with zipfile.ZipFile(temporary_archive_path, mode="w", compression=zipfile.ZIP_DEFLATED) as archive:
+        with zipfile.ZipFile(temporary_archive_path, mode="w", compression=zipfile.ZIP_STORED) as archive:
```
maxxgx, line 33: I think compression does result in a slightly smaller download zip size. Maybe it could be an optional path?

I quickly checked the result of compressing `.bin` files:

Image

Roughly `6.25 -> 5.07` GB. But I think that, compression in this case, would take longer than downloading ~1 GB extra

maxxgx (approved):

application/backend/src/services/model_download_service.py
MarkRedeman: Oh that is better than expected.
Perhaps we can leave it as is (no compression) for now but add it back as an option later if make these model and dataset exports be multi-step jobs. Where the user would then be allowed to set their preferred compression mode when opening the export dialog.

PR #627, opened 2026-05-26
feat: add model exports tab

MarkRedeman:
This PR populates the "Export" tab (renamed to "Export model") to include all available backends.
If a model is expected to support a specific backend, but there is no associated `exports/{backend}` folder, then we show "unavailable". Later we could consider adding a button to re-export the model to this format.

## Type of Change

- [x] ✨ `feat` - New feature

## Screenshots

| **All available** | **PyTorch Unavailable** |
|------------|-----------|
| image           |  image          |


maxxgx (approved):
Nice!

PR #628, opened 2026-05-26
chore: remove redundant files from container images

AlexanderBarabanov:
## Description

Removes security-flagged and non-redistributable files from container images:
- `executorch` ships `pytest`, `pytest-metadata`, `pytest-rerunfailures` and related test packages as unconditional runtime dependencies. These are uninstalled in each device-specific builder stage after `uv sync`.
- `lerobot==0.5.1` pulls in `wandb<0.25.0`, which ships `wandb-core` - a Go binary with multiple CVEs. The binary is not required by Physical AI Studio and is deleted from the builder stage.

## Type of Change

- [x] 🔧 `chore` - Maintenance

### DIFF
--- a/.github/workflows/docker-publish.yml
+++ b/.github/workflows/docker-publish.yml
@@ -85,7 +85,7 @@ jobs:
         id: meta
         uses: docker/metadata-action@030e881283bb7a6894de51c315a6bfe6a94e05cf # v6.0.0
         with:
-          images: ghcr.io/open-edge-platform/physical-ai-studio-${{ matrix.device }}
+          images: ghcr.io/alexanderbarabanov/physical-ai-studio-${{ matrix.device }}
           tags: |
             type=raw,value=${{ steps.version.outputs.version }}-dev-${{ steps.version.outputs.short-sha }}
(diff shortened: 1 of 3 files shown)


MarkRedeman:
Thanks Alexander!
If having those Nvidia libraries in our images is a big blocker for us then we could also consider disabling the compile model feature when the user uses our docker images. The UX would be slightly worse but probably still acceptable.

AlexanderBarabanov:
Thanks Mark! I reverted the Triton GPU backend - the compile model feature should now work.
Let's discuss the next steps separately.

MarkRedeman:
I tested the current Dockerfile and am seeing an error when using the "Compile model" option:
```
Training failed: Cannot find a working triton installation. Either the package is not installed or it is too old. More information on installing Triton can be found at: https://github.com/triton-lang/triton

Set TORCHDYNAMO_VERBOSE=1 for the internal stack trace (please do this especially if you're reporting a bug to PyTorch). For even more developer context, set TORCH_LOGS="+dynamo"
```

MarkRedeman (approved):

PR #631, opened 2026-05-27
refactor: add custom DockView theme with Spectrum integration

MarkRedeman:
- Map DockView CSS variables to Spectrum design tokens
- Configure compact layout with rounded corners and smooth animations
- Add energy-blue accent for drag-and-drop indicators

## Type of Change

## Screenshots

| **Before** | **After** |
|------------|-----------|
| image           |  image         |
| image           | image          |
|  image          |   image        |

### DIFF
--- a/application/ui/package-lock.json
+++ b/application/ui/package-lock.json
@@ -18,7 +18,7 @@
                 "@tanstack/react-query": "^5.80.10",
                 "@types/three": "^0.180.0",
                 "clsx": "^2.1.1",
-                "dockview-react": "^4.13.1",
+                "dockview-react": "^6.3.0",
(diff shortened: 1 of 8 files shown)


MarkRedeman:
One other alternative design I was tinkering with is this floating pill like design:

image

In this case the tabs no longer take up the full horizontal space and instead float on top of the panel content. Needs a bit more tweaking though.

maxxgx (approved):
Floating design looks quite clean as well!

MarkRedeman:
> Floating design looks quite clean as well!

I quite like it as well. Only the drag and drop behaviour felt a bit odd so need to spend a bit more time on that.

PR #636, opened 2026-05-28
chore(deps): upgrade pytest

AlexanderBarabanov:
## Description

This PR upgrades `pytest` as `executorch` requires `pytest<9.0` (its own test-suite constraint). Safe to override because `executorch` production code does not import `pytest` at runtime.

## Type of Change

- [x] 🔧 `chore` - Maintenance


MarkRedeman (approved):
Thanks!

PR #639, opened 2026-05-28
fix: lower z-index of separator so it's not placed on top of dialogs

MarkRedeman:
Fixes the separator in dockview being placed on top of spectrum dialogs by lowering its z-index.

## Type of Change

- [x] 🐞 `fix` - Bug fix

## Screenshots

| **Before** | **After** |
|------------|-----------|
|   image         |   image        |

### DIFF
--- a/application/ui/src/features/dockview/theme.module.scss
+++ b/application/ui/src/features/dockview/theme.module.scss
@@ -62,3 +62,8 @@
     --dv-floating-box-shadow: 0 8px 16px rgba(0, 0, 0, 0.3), 0 4px 8px rgba(0, 0, 0, 0.2);
     --dv-overlay-z-index: 999;
 }
+
+// Overwrite z-index of separator so that it doesn't get placed on top of dialogs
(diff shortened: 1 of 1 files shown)


application/ui/src/features/dockview/theme.module.scss
```diff
@@ -62,3 +62,8 @@
     --dv-floating-box-shadow: 0 8px 16px rgba(0, 0, 0, 0.3), 0 4px 8px rgba(0, 0, 0, 0.2);
     --dv-overlay-z-index: 999;
 }
+
+// Overwrite z-index of separator so that it doesn't get placed on top of dialogs
+:global(.dv-separator-border) :global(.dv-view:not(:first-child):before) {
```
MarkRedeman, line 67: This `:global` is needed because these css rules are inside of a css module. If we don't add it then this class would get a unique class name instead of targeting the actual separator from dockview.

maxxgx (approved):

PR #650, opened 2026-06-01
fix: ACT temporally disable preprocessing in torch

sovrasov:
# Pull Request

## Description

- skip preprocessing on training, because it's missing on inference. That'd allow to avoid some failures on inference
- add a placeholder for expected preprocessor

## Type of Change

- [x] 🐞 `fix` - Bug fix

## Breaking Changes

Old ACT checkpoints are not compatible: need to merge before release


alfieroddan (approved):
Trains fine for me.

PR #654, opened 2026-06-02
docs: add application docs

MarkRedeman:
This PR adds new application documentation under `./application/docs` including a short "Getting started" document and extra guides that go into a bit more depth:

| Guide                        | Description                                                                       |
|------------------------------|-----------------------------------------------------------------------------------|
| Installation                 | Install with Docker or run backend and UI natively.                               |
| Update Existing Installation | Refresh Docker images, dependencies, and services after pulling changes.          |
| Getting Started              | Create a project, set up hardware, record data, train a model, and run inference. |
| Environment Setup            | Configure robots, cameras, and environments.                                      |
| Recording Datasets           | Record, review, import, and export demonstration datasets.                        |
| Training Policies            | Train model policies from recorded datasets.                                      |
| Deploying Model Policies     | Run trained policies in Studio or deploy them externally.                         |

## Type of Change


application/docs/01-installation.md
```diff
@@ -0,0 +1,105 @@
+# Installation
+
+This guide helps you install Physical AI Studio and confirm that the UI is ready.
+
+## Choose how to run
+
+Most users should start with Docker:
```
sovrasov: Sounds a bit strong, maybe rephrase towards something like Docker is the easiest distribution to start with ~ highlight advantages of it for most of potential users?

maxxgx (dismissed):
Really nice docs! It will be really helpful for both users and contributors

application/docs/01-installation.md
MarkRedeman: I see where you're coming from but I would prefer anyone that uses Studio in a production-like environment to use docker and not the uv + npm setup as it makes debugging harder.

I'm thinking of rewriting this into the following:
```
Docker is the recommended path for most users:

- You do not need to install development tooling such as [uv](https://docs.astral.sh/uv/) or [nodejs](https://nodejs.org/en).
- Docker starts both the backend and UI together.
- The container images include the application runtime prerequisites, no need to install them on your own system.
- Running in containers keeps the application environment separate from your system, which makes setup and debugging more predictable.
```

what do you think of this?

sovrasov: Yeah, I was mostly referring to `should` usage, the arguments are pretty clear already

MarkRedeman: Ok in that case: I've updated that sentence so say "Docker is the recommended path for most users"

sovrasov (approved):

PR #677, opened 2026-06-05
refactor(ui): download models and datasets directly

MarkRedeman:
Open download URLs in a new tab instead of streaming archives through the browser to avoid excessive memory usage.

Before we would download the model / dataset zip and transform the response into a blob that was downloaded then downloaded by the UI. This resulted in the UI requiring to load up the full response in memory which results in issues when trying to download large datasets or models.
Now we avoid this by letting the browser be responsible for downloading it.

## Type of Change


RHeckerIntel (approved):
Awesome, yes, this'll do

PR #681, opened 2026-06-08
fix: move samples inputs to the same device that holds policy

sovrasov:
# Pull Request

## Description

- Explicitly move model inputs to the policy device during export
- Overall, that's experimental changes: might not work with all the devices, policies and export backends we have

## Type of Change

- [x] 🐞 `fix` - Bug fix


MarkRedeman (approved):
Thanks Vlad! I've verified that this is working for all models (with and without `compile_model`) on both XPU and CUDA.

PR #683, opened 2026-06-09
refactor: Solidify environment relations

RHeckerIntel:
Previously the connection between environments and robots and cameras was a json column.
This PR adds join tables to solidify the environment relations

### Why no names

Previously this PR contained the names in the link. Below is the argument for. But I prefer not too. Although the name makes sense, with dataset and model import we need to map cameras to a dataset or model input anyway. Having a locked name inside the environment won't help there. An explicit map needs to be added for that (or atleast cached for ux). Given that we need this map we might as well always use it. New datasets will just use the environment camera names by default.

The only open is what to do with multiple robots. In that case the explicit map comes to the rescue. Once a user creates a new dataset he should see the map and then he can give the prefix names there.

### Why names (outdated and not implemented)

These names are important because camera names do not have to match the input for an environment.
For example, you might have multiple robots connected to your system. Naming multiple right grabber cameras all "right" will be very confusing. I propose we name them a logical name (e.g. inno so101 or D415) and then when creating an environment we give them the name for that setup.

### Next steps
This is the first step in a larger move towards using manifests for matching interoperability between environments, datasets and models.


application/backend/src/alembic/versions/e1f2a3b4c5d6_environment_join_tables.py
```diff
@@ -0,0 +1,171 @@
...
+    with op.batch_alter_table("project_environments", schema=None) as batch_op:
+        batch_op.drop_column("robots")
+        batch_op.drop_column("camera_ids")
+
+    op.create_table(
+        "environment_robots",
+        sa.Column("environment_id", sa.Text(), nullable=False),
+        sa.Column("robot_id", sa.Text(), nullable=False),
+        sa.Column("name", sa.String(length=255), nullable=False),
+        sa.Column("position", sa.Integer(), nullable=False),
```
MarkRedeman: At first I thought `position` could be representing things like "left robot", "right robot" etc but from reading the code I understand this is mainly intended to be used as a orderBy thing.
Maybe it could be renamed to `index` instead, and/or perhaps also add a unique key constraint in combination with `environment_id`?

application/backend/src/repositories/project_environment_repo.py
```diff
@@ -32,70 +28,78 @@ def to_schema(self) -> Callable[[Environment], ProjectEnvironmentDB]:
...
+        result = await self.db.execute(stmt)
+        env = result.scalars().first()
         if env is None:
             return None
 
-        robots_map = await self._fetch_robots_map(env.robots)
-        cameras = await self._fetch_cameras(env.camera_ids)
-        robots_with_teleoperators = self._build_robots_with_teleoperators(env.robots, robots_map)
+        cameras = [
+            ProjectCameraMapper.from_schema(link.camera).model_copy(update={"name": link.name})
```
MarkRedeman: Would this not result into an `N + 1` problem? I.e. suppose you have 4 cameras, would this then do 1 query for the environment and 4 for the cameras, or is alembic smart enough to do eager loading?

RHeckerIntel: The data is already there, just mapping it and updating the name with the link name. This is temporary until we fix the environment with relations and how we build a dataset features.

Basically, imagine your environment has a camera D415 and you named it overview in the environment. The EnvironmentWithRelations would get cameras: [{id: _, name: "overview", ....}].

application/backend/src/alembic/versions/e1f2a3b4c5d6_environment_join_tables.py
RHeckerIntel: Oh good point, no need for order anyway.

application/backend/src/alembic/versions/e1f2a3b4c5d6_environment_join_tables.py
```diff
@@ -0,0 +1,170 @@
...
+        sa.PrimaryKeyConstraint("environment_id", "camera_id"),
+        sa.ForeignKeyConstraint(["environment_id"], ["project_environments.id"], ondelete="CASCADE"),
+        sa.ForeignKeyConstraint(["camera_id"], ["project_cameras.id"], ondelete="CASCADE"),
+    )
+
+    for raw_env_id, robots_raw, camera_ids_raw in environments:
+        env_id = str(raw_env_id)
+
+        for robot_cfg in _load_json(robots_raw):
+            robot_id = str(robot_cfg.get("robot_id"))
```
MarkRedeman, line 84: This is rasing a `'str' object has no attribute 'get'` for me.

I'm trying to debug this and checking if this is perhaps an issue with my local database. But would be good to be careful here.

RHeckerIntel: Fixed, we currently double encode/decode the json string there. Didn't account for that.

MarkRedeman (approved):

PR #684, opened 2026-06-09
feat(backend): Add training summary and hyper parameters to model details endpoint

MarkRedeman:
Extend the `/api/models/{model_id}` endpoint to return a training summary (based on its training job) and hyper parameters based on `hparams.yaml`.
We will use this in the UI to visualize training details.

## Type of Change

- [x] ✨ `feat` - New feature


RHeckerIntel (dismissed):
Awesome. Nice amount of detail too.

RHeckerIntel (approved):

PR #688, opened 2026-06-10
chore: rearrange unit and integration tests

sovrasov:
# Pull Request

## Description

- add a job to run "not slow" integration tests
- mark more tests as "slow"
- move extended lerobot wrappers tests from unit to integration

## Type of Change

- [x] 🧪 `test` - Tests
- [x] 🔧 `chore` - Maintenance

## Limitations

- Libero gym doesn't work on the particular runner we have
- pi* models are not tested, as they require an access token


samet-akcay (approved):
lgtm, thanks for addressing this

PR #703, opened 2026-06-15
feat: refactor camera worker

RHeckerIntel:
First step of work towards #613 

Refactor camera_worker to write to shared memory. This is overkill in the current implementation, but really useful towards the new robot control orchestrator where recording will happen in separate process.  A separate `DataRegistry` will be built where the robots and cameras shared memory and tensor info (joint_names and width, height, fps respectively) will be shared.

Although SharedCamera gives that functionality keeping it all in line will make it easier to maintain.


application/backend/src/api/camera.py
```diff
@@ -86,52 +84,37 @@ async def camera_websocket_openapi(
...
-    registry: CameraRegistryDep,
-    locked_camera_fingerprints: RecordingLockedCamerasDep,
     camera: Annotated[ProjectCamera, Depends(get_camera_from_query)],
 ) -> None:
-    """WebSocket endpoint for camera streaming.
+    """
+    WebSocket endpoint for camera streaming.
 
     Query Parameters:
+        None
```
AlbertvanHouten:
```suggestion
```

application/backend/src/api/camera.py
```diff
@@ -86,52 +84,36 @@ async def camera_websocket_openapi(
...
+    WebSocket endpoint for camera streaming.
 
     Query Parameters:
         camera: JSON serialized ProjectCamera
 
     Protocol:
-        Client sends JSON messages:
-            {"event": "disconnect"} - Request graceful disconnect
-            {"event": "ping"} - Keep-alive check
+        Client sends JSON message on connect:
```
MarkRedeman, line 96: The camera is currently send as a query parameter (`[camera](camera: Annotated[ProjectCamera, Depends(get_camera_from_query)],)`)
Do you want to change this or should we update the docstring?

I think a downside of the current query parameter approach is that this requires us to encode the serialized json so it can be used as a url.
If we get rid of this and use a message to send the config then that probably simplifies the encoding/decoding?

application/backend/src/workers/camera_worker.py
```diff
@@ -1,145 +1,69 @@
...
 
 from schemas.project_camera import Camera
 from utils.camera_factory import build_shared_camera
-from utils.jpeg import encode_jpeg_rgb
-from workers.transport.worker_transport import WorkerTransport
-from workers.transport_worker import TransportWorker, WorkerState, WorkerStatus
+from workers.base import BaseThreadWorker, run_at_frequency
 
 
-class CameraWorker(TransportWorker[Camera]):
```
MarkRedeman, line 16: In a follow up PR we can probably consider removing this `TransportWorker`, perhaps also removing the `WebSocketTransport` and `WorkerTransport` if we don't plan on using these anymore.

My initial idea for adding the `WorkerTransport` interface was that we could easily swap back to using webrtc but I don't expect we will do that.

application/backend/src/workers/camera_worker.py
```diff
@@ -1,145 +1,69 @@
...
-                self.state = WorkerState.ERROR
-                self.error_message = str(exc)
-                logger.error(f"Failed to connect camera {self.config.name}: {exc}")
-                await self.transport.send_json(
-                    WorkerStatus(state=self.state, config=self.config, message=str(exc)).to_json()
-                )
-                return
+    @staticmethod
+    def frame_from_buffer(buffer: Any, width: int, height: int) -> np.ndarray:
+        return np.frombuffer(buffer, dtype=np.uint8).reshape(height, width, 3).copy()
```
MarkRedeman, line 43: Probably not relevant, but suppose we would want to add support for depth images from realsense, would we then need to change this to have more channels?

MarkRedeman (approved):

AlbertvanHouten (approved):

application/backend/src/api/camera.py
```diff
@@ -86,52 +84,36 @@ async def camera_websocket_openapi(
...
-            logger.error(f"Could not close websocket: {e}")
-    except Exception as e:
-        logger.exception(f"Unexpected error in camera websocket: {e}")
-        try:
-            await websocket.close(code=status.WS_1011_INTERNAL_ERROR)
-        except Exception as e:
-            logger.error(f"Could not close websocket: {e}")
+        worker = CameraWorker(camera)
+        worker.start()
+        while True:
```
AlbertvanHouten, line 109: Should this loop keep running if the CamerWorker stops or should it also use `while not camera.should_stop()` similar to the loop in the CameraWorker?

RHeckerIntel: True should be fine with the disconnect except. But indeed if there is an error in the camera worker it will probably keep sending the latest frame instead of sending a error to the user. We send the images as pure bytes so we cannot send error messages, but we could perhaps gracefully handle this better.

application/backend/src/workers/camera_worker.py
RHeckerIntel: Yes, but given that in the last year no depth has been applied, and we're using pretrained models that dont have that either, it's probably not an issue.

PR #707, opened 2026-06-15
refactor:  Robot worker to teleoperate worker

RHeckerIntel:
Second part in #613.

Moves the robot  worker to teleoperate worker. This worker now allows robots to connect, but also with a leader.
Big feature is that it runs in separate worker and shares its data using SharedMemory. This will be combined with the camera worker into a data registry so that observations can be retrieved in a non blocking way.

A pretty nice bonus is that in the preview of environment you are able to teleoperate. Though this might be a bit yolo because it does so automatically. Perhaps adding a toggle to connect might be nice.

Some notes:
1. Made physical ai adapter for robotclient synchronous. Yes this will be blocking, but its the responsibility of the user of the adapter to make sure it's not blocking the fastapi thread. Async tends to bleed thru and the to_thread for every call is unnecessary in a lot of cases.
2. The health api call is rather useless now. A registry should be added back again at the end of these series of refactors to keep track.


application/backend/src/api/robot_control.py
```diff
@@ -55,39 +54,34 @@ async def robot_websocket(  # noqa: PLR0913
...
+        follower_id = get_robot_id(settings["follower_id"])
+        robot_client_factory = RobotClientFactory(robot_manager, calibration_service)
+        follower = await robot_service.get_robot_by_id(project_id, follower_id)
+        follower_client = await robot_client_factory.build(follower)
+
+        leader_client = None
+        if "leader_id" in settings:
+            leader_id = get_robot_id(settings["leader_id"])
+            leader = await robot_service.get_robot_by_id(project_id, leader_id)
+            leader_client = await robot_client_factory.build(leader)
```
MarkRedeman, line 108: I feel like this might be a bit much, or hard to follow if we later add more robots / leaders. Maybe extract some of this to a private function?

application/backend/src/api/robot_control.py
```diff
@@ -55,39 +54,34 @@ async def robot_websocket(  # noqa: PLR0913
...
-    except ValueError as e:
-        logger.error(f"Failed to register worker: {e}")
-        try:
-            await websocket.send_json({"event": "error", "message": str(e)})
-            await websocket.close(code=status.WS_1008_POLICY_VIOLATION)
-        except Exception as close_err:
-            logger.error(f"Could not close websocket after ValueError: {close_err}")
-
+        worker.start()
+        worker.set_action_source(ActionWriteState.FROM_LEADER)
```
MarkRedeman: It's not immediately clear to me  what `set_action_source` is intended to do.

The type of API design I'd prefer is if could allow for something like,
```
# pass leader_client here, instead of in constructor
worker.setTeleoperatorPolicy(ImitationPolicy(leader_client)

# When performaing inference
worker.setTeleoperatorPolicy(ModelPolicy(model))
```

I.e. let the user determine the behaviour from the user in a "Tell don't ask" type of way.

MarkRedeman (commented):
I really like the initial approach, two quick comments/ideas based no a quick scan of the code.
Will do some testing as well

application/backend/src/api/robot_control.py
RHeckerIntel: I hear you. The worker is a separate process, so cant send that info without the teleoperator taking control of the policy running too.

The idea is that the teleoperator worker has two modes. It either writes to actions, or reads from actions. When it writes to actions it reads those actions from the leader. Although these two modes are now the only options have a default "do nothing" might be nice too. Allows for connection (say in environments screen) without moving the robots until the user sets mode to teleoperate. (EDIT: it seems the "NONE" is already there)

I now realize that "set_action_source" makes sense from a follower pov, but not from a design pov.

application/backend/src/api/robot_control.py
RHeckerIntel: I agree, but whatever private function I write will also have to do this and get all the dependencies. Maybe I need to make the robot client factory more powerful or something.

application/backend/src/workers/base.py
```diff
@@ -9,12 +9,13 @@
...
 from abc import ABC, abstractmethod
 from contextlib import asynccontextmanager
+from multiprocessing import Event
 from typing import TYPE_CHECKING
 
 if TYPE_CHECKING:
     from collections.abc import AsyncGenerator, Iterable
     from multiprocessing.queues import Queue
-    from multiprocessing.synchronize import Event
+    from multiprocessing.synchronize import Event as EventClass
```
MarkRedeman, line 18: What's the difference between these two `Event` classes?
Perhaps we can rename `EventClass` to `SynchronizeEvent`?

application/backend/src/api/robot_control.py
```diff
@@ -1,51 +1,88 @@
...
+        logger.error(f"Outgoing task stopped: {e}")
+
+
+async def handle_incoming(websocket: WebSocket, worker: TeleoperateWorker) -> None:
+    """Handle incoming messages from client to teleoperate worker."""
+    try:
+        while not worker.should_stop():
+            data = await websocket.receive_json("text")
+            payload = data.get("data", {})
+            match data["event"]:
```
MarkRedeman, line 60: The previous robot commands allows the UI to enable/disable torque of a robot as well as set the joint state, so that for instance you can try moving individual joints using the sliders shown in the robot page.

Though we don't use this atm (I think it got removed, or maybe I forgot to merge it.. 😁 ), I think though the overall structure of the `commands.py` setup was quite useful, specifically because it parses the commands instead of assuming the payload is always correct. This gives some better error messages when you use a wrong command.
Perhaps we can look into adding those back again?

application/backend/src/workers/teleoperate_worker.py
```diff
@@ -0,0 +1,131 @@
...
+class TeleoperateWorker(BaseProcessWorker):
+    """Robot control and teleoperate worker
+
+    This Worker Class connects to a robot and if provided a leader robot.
+    Data is stored in mp.SharedMemory.
+
+    The worker can be in 3 ActionReadState modes for the follower:
+    - NONE: follower robot does not receive any actions
+    - TELEOPERATION: follower robot position is set from the leader's robot position
+    - FROM_ACTIONS: follower robot position is set from current _output_actions mp.SharedMemory.
```
MarkRedeman, line 30: So essentially this makes it hold its own position?

application/backend/src/workers/teleoperate_worker.py
```diff
@@ -0,0 +1,131 @@
...
+        # Set current actions to current follower state.
+        state = (self.follower.read_state())["state"]
+        self._set_actions([state[key] for key in self.features])
+
+        self.loaded_event.set()
+
+    async def run_loop(self) -> None:
+        try:
+            # Teleoperate loop until unload is requested
+            goal_time = 1 / self.frequency
```
MarkRedeman, line 110: What's the reasoning to later on multiply by 2 or 3 depending on the read state?

MarkRedeman (approved):

application/backend/src/workers/teleoperate_worker.py
RHeckerIntel: Longer goal time interpolates a bit. Though factor 1 should be fine it is more stable to increase it a little. This is what I found in the field.

application/backend/src/api/robot_control.py
RHeckerIntel: Yeah, but basically the action_read_state should handle that. It's just a matter of following that up in the websocket

application/backend/src/workers/base.py
RHeckerIntel: EventClass is just for typing.

PR #713, opened 2026-06-16
fix: don't drop models folder during migration

MarkRedeman:
This PR removes `rmtree(settings.models_dir)` from our migrations.

While testing #706 this migration accidentally removed my models because I had to rerun all database migrations.
I think the original intend was that this would make it so that the user starts with a clean storage dir, and not have any orphaned models.
I think a better long term solution would be that we expose a "storage" page in the UI showing an overview of storage used by the system and an option to detect and remove orphaned files (e.g. those that are no longer being referenced in the database).


application/backend/src/alembic/versions/679bf09bd098_adds_snapshots.py
AlbertvanHouten, line 1: Comment on line 28 is now no longer valid

AlbertvanHouten (approved):

Issue #705, opened 2026-06-15
[Feature]: Import environment not only dataset


intel-ohara:
### Problem Statement

When importing dataset, physical AI studio requests to choose environment. I recorded dataset on a laptop connected a robot and want to train models on desktop that locates in different place. Setting up an environment requires physical connection with robots ans cameras, but it is inconvenient. The capability of exporting and importing environments would be convenient.

### Proposed Solution

The feature should allow users to export environment including robots and camera setups from one PC and import to another PC.

### Component

Multiple components

MarkRedeman:
Thanks for reporting. I agree that the current flow is rather annoying.

We've been talking internally about improving the implicit relations between environments, datasets and models. The main reason we have these is that we need to make sure that these are compatible with each other (e.g. the datasets use the same amount of cameras and resolution and the same type of robots etc)

We plan to improve the application so that these roles will be more explicit and obvious to the user, including warning them that the environment and dataset might not be compatible.

Besides that we've also found that simply exporting your environment and importing it in another project could be very helpful.
I think what we can do first is create a simple environment export + import feature. Later we can also do this for individual robots (+ calibration) and camera settings.

Another thing that I'd like to look at later is to make the selected environment when creating/importing a dataset be optional. But this requires some more work.

PR #715, opened 2026-06-17
fix: Address several UX issues 

AlbertvanHouten:
## Summary

This PR fixes several UX issues within the UI: 

- 3D viewers (robot-viewer.tsx, setup-robot-viewer.tsx): disabled OrbitControls damping to remove drift/inertia after dragging.
- Robot form (robot-form/form.tsx): added an onSubmit handler so pressing Enter triggers the submit button (respecting its disabled state) instead of reloading the page and discarding form state.
- Environment form (form.tsx, submit-new-environment-button.tsx): autofocus the name field, fixed the label casing (name → Name), and re-enabled real validation (disable submit when name is empty) replacing the hardcoded isDisabled = false.
- New dataset form (new-dataset.component.tsx): made Environment a required picker, moved autofocus off the Task field, and disabled Save until both name and environment are set.
- Datasets empty state (datasets/index.tsx): fixed copy "Currently there are no datasets available."
- Training job table (job-table.component.tsx): wrapped Interrupt in a confirmation AlertDialog.
- Make train button unavailable if no dataset is selected yet.

### DIFF
--- a/application/ui/src/features/robots/environment-form/submit-new-environment-button.tsx
+++ b/application/ui/src/features/robots/environment-form/submit-new-environment-button.tsx
@@ -19,7 +19,7 @@ export const SubmitNewEnvironmentButton = () => {
 
     const environment_id = uuidv4();
     const body = useEnvironmentFormBody(environment_id);
-    const isDisabled = false; // body.name.length === 0 || body.robots.length === 0 || body.cameras.length === 0;
+    const isDisabled = body.name.length === 0;
 
     return (
         <Button
--- a/application/ui/src/features/robots/environment-form/submit-new-environment-button.tsx
+++ b/application/ui/src/features/robots/environment-form/submit-new-environment-button.tsx
@@ -19,7 +19,7 @@ export const SubmitNewEnvironmentButton = () => {
 
     const environment_id = uuidv4();
     const body = useEnvironmentFormBody(environment_id);
-    const isDisabled = body.name.length === 0;
+    const isDisabled = body.name.trim().length === 0 || body.robots.length === 0 || body.cameras.length === 0;
 
     return (
         <Button
--- a/application/ui/src/features/robots/environment-form/submit-new-environment-button.tsx
+++ b/application/ui/src/features/robots/environment-form/submit-new-environment-button.tsx
@@ -19,7 +19,7 @@ export const SubmitNewEnvironmentButton = () => {
 
     const environment_id = uuidv4();
     const body = useEnvironmentFormBody(environment_id);
-    const isDisabled = body.name.trim().length === 0 || body.robots.length === 0 || body.cameras.length === 0;
+    const isDisabled = body.name.trim().length === 0;
 
     return (
         <Button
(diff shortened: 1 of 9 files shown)


application/ui/src/features/robots/environment-form/submit-new-environment-button.tsx
```diff
@@ -19,7 +19,7 @@ export const SubmitNewEnvironmentButton = () => {
 
     const environment_id = uuidv4();
     const body = useEnvironmentFormBody(environment_id);
-    const isDisabled = false; // body.name.length === 0 || body.robots.length === 0 || body.cameras.length === 0;
+    const isDisabled = body.name.trim().length === 0 || body.robots.length === 0 || body.cameras.length === 0;
```
MarkRedeman: I think I accidentally merged that commented line.
However it might be best to remove the `body.robots.length` check for now because otherwise you can't create an environment on a remote environment.

The latter is useful if you want to import a dataset and then train a model with it.
We need to improve that UX, see also my comment at #705

application/ui/src/features/robots/robot-form/form.tsx
```diff
@@ -278,6 +280,15 @@ export const RobotForm = ({ heading = 'Add new robot', submitButton = <SubmitNew
     const robotForm = useRobotForm();
     const setRobotForm = useSetRobotForm();
 
+    const submitContainerRef = useRef<HTMLDivElement>(null);
+
+    // Make Enter behave like clicking the submit button. A disabled button ignores
+    // the click, so validation still applies.
+    const handleSubmit = (event: FormEvent<HTMLFormElement>) => {
+        event.preventDefault();
+        submitContainerRef.current?.querySelector('button')?.click();
```
MarkRedeman, line 289: Not a fan of this approach. It would be better if we move the submit logic from the button into this form.
Do you mind if later today I push a change on your branch to do this?

application/ui/src/routes/models/job-table.component.tsx
```diff
@@ -124,9 +138,19 @@ export const TrainingRow = ({
...
+                                    <Button variant='secondary'>Interrupt</Button>
+                                    <AlertDialog
+                                        onPrimaryAction={onInterrupt}
+                                        title='Interrupt training'
+                                        variant='warning'
+                                        primaryActionLabel='Interrupt'
+                                        cancelLabel='Cancel'
+                                    >
+                                        Are you sure you want to interrupt training for {trainJob.payload.model_name}?
+                                        This stops the run and cannot be undone.
```
MarkRedeman: We could extend this comment that we will save and export the model's checkpoint at the current training step.

AlbertvanHouten: Done

application/ui/src/features/robots/robot-form/form.tsx
AlbertvanHouten: Sure, sounds good.

MarkRedeman: Spend a bit of time checking this but it will require some bigger changes to do this properly in idomatic react. I'll keep that in mind and improve this part of the while preparing support for more robots.

Let's merge it as is.

MarkRedeman (approved):

PR #719, opened 2026-06-18
feat: add RoboCasa Kitchen support with isolated CI lane and RoboCasa365 roadmap

eugene123tw:
# Pull Request

## Description

This PR adds RoboCasa Kitchen support to the library with isolated environment setup, CI coverage, tests, and docs.

Current limitation:
* The implementation is Kitchen-only (RoboCasa v1.0 task groups and named kitchen tasks).
Full RoboCasa365 is not included in this PR because asset size and setup/runtime cost are much larger, which would significantly impact regular CI and local iteration speed.

Why this design:
* RoboCasa and LIBERO have conflicting dependency constraints, so RoboCasa is installed in a separate virtual environment and validated in a dedicated CI lane.

Future RoboCasa365 support plan:

1. Add install profiles (kitchen vs robocasa365) with explicit asset download paths.
2. Add separate task catalogs and split mappings for 365 tasks.
3. Extend config surface in a backward-compatible way (default remains kitchen).
4. Add a separate non-blocking CI workflow for RoboCasa365 (nightly or manual trigger).
5. Expand tests from one smoke E2E to broader 365 coverage over time.
6. Document resource requirements and migration guidance.

## Type of Change

- [x] ✨ feat - New feature
- [x] 📚 docs - Documentation
- [x] 🧪 test - Tests
- [x] 🔧 chore - Maintenance

## Related Issues

- Relates to RoboCasa integration work for library gym wrappers and CI environment isolation.

## Breaking Changes

None expected for existing LIBERO or default test paths.

## Changes Made

1. Added RoboCasa gym wrapper and registration:
- robocasa.py
- __init__.py

2. Added dedicated install script for RoboCasa + robosuite pinned SHAs:
- install_robocasa.sh

3. Added tests:
- Unit tests:
  - test_robocasa.py
- Integration tests:
  - test_robocasa_e2e.py

4. Updated CI with a separate RoboCasa integration job and required-check wiring:
-  .github/workflows/library.yml

5. Added and updated docs for gym usage:
- New:
  - robocasa.md
- Updated:
  - README.md

6. Updated pyrefly excludes with rationale for optional dependency environment split:
- pyproject.toml

## Validation

- RoboCasa-specific unit and integration tests were added.
- CI now runs RoboCasa tests in an isolated job/environment.
- Existing integration lane remains separate (no RoboCasa dependency injection into LIBERO path).

## Notes for Reviewers

- The environment split is intentional and required.
- The install script pins upstream SHAs to keep behavior reproducible and aligned with lerobot benchmark setup.
- Docs explicitly explain why RoboCasa is not a pyproject extra in the shared environment.


library/docs/explanation/gyms/robocasa.md
alfieroddan, line 1: My understanding is that the Robocasa365 benchmark would be integrated later, so this comment is just for v1.

Do you plan a follow up PR for the benchmark API for robocasa v1? This gym I think has the capability to be a benchmark with minimal changes.

Perhaps a benchmark folder like liberos `https://github.com/open-edge-platform/physical-ai-studio/tree/main/library/src/physicalai/benchmark/gyms/libero`. Perhaps with an example script for Pi05 / SmolVLA with results?

library/scripts/install_robocasa.sh
```diff
@@ -0,0 +1,81 @@
+#!/usr/bin/env bash
+# Copyright (C) 2025-2026 Intel Corporation
```
alfieroddan: Just 2026?

library/src/physicalai/gyms/robocasa.py
```diff
@@ -0,0 +1,596 @@
+# Copyright (C) 2025 Intel Corporation
```
alfieroddan: 2026

library/src/physicalai/gyms/robocasa.py
```diff
@@ -0,0 +1,596 @@
...
+# SPDX-License-Identifier: Apache-2.0
+
+"""RoboCasa Kitchen benchmark environment wrapper with optional dependencies.
+
+Ported from ``lerobot/src/lerobot/envs/robocasa.py`` with PAS-native types
+and the lerobot ``AsyncVectorEnv`` plumbing removed. The Studio benchmark
+loop drives episodes single-environment-at-a-time and resets explicitly
+between rollouts (see ``LiberoBenchmark``), so the upstream
+auto-reset-on-terminal step is also dropped.
```
alfieroddan, line 10: Why have we removed Async? Just easier?

LIBERO and PushT have async capabilities so it would be inline if it's not too hard. Would be useful for later, when benchmarking.

library/scripts/benchmark/install_robocasa.sh
```diff
@@ -0,0 +1,81 @@
...
+#   uv sync --active --extra cu128       # or --extra cpu / --extra xpu
+#   bash library/scripts/install_robocasa.sh
+
+set -euo pipefail
+
+ROBOCASA_SHA="${ROBOCASA_SHA:-56e355ccc64389dfc1b8a61a33b9127b975ba681}"
+ROBOSUITE_SHA="${ROBOSUITE_SHA:-aaa8b9b214ce8e77e82926d677b4d61d55e577ab}"
+CLONE_ROOT="${CLONE_ROOT:-$HOME/.cache/physicalai/robocasa-src}"
+
+if [[ -z "${VIRTUAL_ENV:-}" ]]; then
```
alfieroddan, line 36: Perhaps we could add a statement to test of libero is installed in current env? To avoid clashing envs?

.github/workflows/library.yml
alfieroddan, line 1: I'm unsure on the tests - should we fix before merge?

library/docs/explanation/gyms/robocasa.md
```diff
@@ -0,0 +1,211 @@
...
+```bash
+uv venv .venv-robocasa
+source .venv-robocasa/bin/activate
+uv sync --extra cu128          # or cpu / xpu — do NOT use --extra all or --extra libero
+
+bash library/scripts/install_robocasa.sh
+
+# Download kitchen assets (~4.4 GB on disk; prompts interactively)
+yes y | python -m robocasa.scripts.download_kitchen_assets \
+    --type tex tex_generative fixtures_lw objs_lw
```
alfieroddan, line 36: This fails on VPN for me. I'm not sure how useful just local benchmark will be. Perhaps we can move assets to our own HF. Unsure.

library/docs/explanation/gyms/robocasa.md
```diff
@@ -0,0 +1,211 @@
...
+  `[libero]` extra needs PyPI `robosuite==1.4.0`. They cannot share a venv.
+- `robocasa`'s dead `tianshou==0.4.10` pin transitively requires `protobuf<3.20`,
+  colliding with `onnx`'s `protobuf>=3.20`.
+
+Install into a dedicated `.venv-robocasa`:
+
+```bash
+uv venv .venv-robocasa
+source .venv-robocasa/bin/activate
+uv sync --extra cu128          # or cpu / xpu — do NOT use --extra all or --extra libero
```
alfieroddan: `uv sync --extra cu128 --active` - just did this and uv defaults to .venv if available.

library/src/physicalai/gyms/__init__.py
alfieroddan: 2025-2026? Unsure.

library/tests/integration/gyms/test_robocasa_e2e.py
```diff
@@ -0,0 +1,152 @@
+# Copyright (C) 2025 Intel Corporation
```
alfieroddan: 2026?

library/tests/unit/gyms/test_robocasa.py
```diff
@@ -0,0 +1,220 @@
+# Copyright (C) 2025 Intel Corporation
```
alfieroddan: 2026?

library/src/physicalai/gyms/robocasa.py
```diff
@@ -0,0 +1,596 @@
...
+        Tuple of ``(list[task_name], split_or_None)``.
+
+    Raises:
+        ValueError: If ``task`` is empty, or if it is a recognized group
+            keyword but missing from the installed robocasa registry.
+    """
+    key = task.strip()
+
+    if key in _TASK_GROUP_SPLITS:
+        _check_robocasa_available()
```
alfieroddan: Up to you but I'd keep the `_check_robocasa_available` only in the public API's.

alfieroddan (commented):
Thanks Eugene!

The actual gym is looking good, just have some comments on surrounding components.

Oddities on my side:
- I can't download assets on VPN, can you?

library/docs/explanation/gyms/robocasa.md
eugene123tw: For RoboCasa v1, yes, I’m planning a follow-up PR to add benchmark support under the same benchmark gym area as the existing integrations. The current PR is focused on the gym scaffolding only, since the benchmark layer is intended to land together with the RLDX-1 integration.

That follow-up PR will add the benchmark entry points and an example evaluation script for running RLDX-1 on RoboCasa. I agree the current gym is already close to benchmark-ready, but I wanted to keep this PR scoped to the environment foundation and layer the benchmark API on top in the next step.

library/docs/explanation/gyms/robocasa.md
eugene123tw: Fixed.

library/docs/explanation/gyms/robocasa.md
eugene123tw: Try downloading the assets off VPN. Once downloaded, they're cached locally so you won't need to re-fetch them. If that's not feasible on your setup, we can explore staging the assets on our own HF hub for the team.

library/tests/unit/gyms/test_robocasa.py
eugene123tw: Updated

library/tests/integration/gyms/test_robocasa_e2e.py
eugene123tw: Updated

library/src/physicalai/gyms/__init__.py
eugene123tw: Updated

library/src/physicalai/gyms/robocasa.py
eugene123tw: Updated

library/scripts/install_robocasa.sh
eugene123tw: Updated

library/src/physicalai/gyms/robocasa.py
eugene123tw: I’ll add async support in the second PR. Since we don’t yet have pretrained RoboCasa weights, adding robocasa benchmark and async in this phase won’t provide much practical benefit.

That said, based on the PushTBenchmark pattern, adding a RoboCasa vectorized path (for example a RoboCasaGym.vectorize-style API) should be straightforward, and I plan to include it in that follow-up.

library/src/physicalai/gyms/robocasa.py
eugene123tw: Removed

library/scripts/benchmark/install_robocasa.sh
eugene123tw: added

alfieroddan (dismissed):
Thanks!

library/scripts/benchmark/install_robocasa.sh
samet-akcay, line 1: is this script to install venv? how about windows users?

eugene123tw: This is the script to install robocasa for developers to train a policy or benchmark policy on robocasa, the reason for a isolated venv is because dependency conflict.

It's not going to work on windows for this PR, but maybe we can discuss that in the future.

samet-akcay (approved):

alfieroddan (approved):

AlexanderBarabanov (approved):

PR #720, opened 2026-06-18
fix: derive dataset storage path from id to prevent data leakage between projects

AlbertvanHouten:
## Description

Episodes recorded into one project's dataset showed up in a different project's dataset that the user never touched. This is caused by datasets deriving their storage paths from the dataset name. If 2 projects use the same dataset name, they will share the same storage folder.

## Fix
The server now owns dataset placement, derived from the globally unique dataset id.

- Dataset.path is a read-only computed field: `settings.datasets_dir / <id>`. Clients no longer send a path; the request body and the schema dropped it.
- The create endpoint and import adapter no longer accept or set a path.
- UI stops computing and sending a path.

The id is unique, so two datasets can never share a folder regardless of name. This matches what the dataset import flow already did.

Includes a migration script with cli command to migrate old datasets with regular names to ID based names.

---

> Supersedes #711, which was auto-closed after an accidental force-push briefly reset this branch to `main`. The branch has been restored; this PR carries the same changes.


MarkRedeman (approved):
Thanks Albert! I tested on my own datasets and migration was smooth.

PR #730, opened 2026-06-22
refactor: Robot Control Orchestrator

RHeckerIntel:
Third PR in series for #613

Implements orchestrator and separate workers for recording, teleoperation and cameras.

General idea is that the orchestrator manages the subworkers. A data registry handles reading the data from the workers.
This implementation has a couple of core advantages:
1. Robustness - workers can be started and stopped easily
2. Teleoperation, recording, inference and ui updates can all run on their own refresh rate.
3. Steady teleoperation - Robots have their own process so nothing can interrupt their loop.

## why not use PhysicalAI Framework's Policy?

The Frameworks policy does not allow for flexibility for multiprocessing or changing modes like teleoperation or human in the loop.


application/ui/src/features/robots/robot-control-provider.tsx
```diff
@@ -150,6 +150,7 @@ export const RobotControlProvider = (props: useRobotControlProps) => {
 
     const onMessage = ({ data }: WebSocketEventMap['message']) => {
         const message = JSON.parse(data) as RobotControlApiJsonResponse<unknown>;
+        console.log(message);
```
MarkRedeman: Leftover ;)

application/backend/src/control/utils.py
```diff
@@ -0,0 +1,105 @@
...
+
+
+def format_observation_for_reporting(observation: dict, manifest: EnvironmentDataRegistry) -> dict:
+    """Format observation for UI."""
+    actions = {i: observation["action"][k] for k, i in enumerate(manifest.robot.features)}
+    state = {i: observation["state"][k] for k, i in enumerate(manifest.robot.features)}
+    camera_images = {}
+    for camera in manifest.cameras:
+        frame = np.ascontiguousarray(observation["images"][camera.id][..., ::-1])  # RGB→BGR for cv2
+        success, imagebytes = cv2.imencode(".jpg", frame)
```
MarkRedeman: Could we use `encode_jpeg_rgb` from `utils.jpeg` here?

application/backend/src/api/record.py
```diff
@@ -63,72 +61,76 @@ async def handle_incoming(
...
     try:
         while True:
-            try:
-                loop = asyncio.get_running_loop()
-
-                message = await loop.run_in_executor(None, queue.get)
-                await websocket.send_json(message)
-            except Empty:
-                await asyncio.sleep(0.05)
+            data = await queue.get()
```
MarkRedeman, line 80: Why remove the try catch here?
If for some reason we can't send the data over websocket (maybe there's a temporary network issue) then that would require the user to reconnect?

application/backend/src/api/record.py
```diff
@@ -63,72 +61,76 @@ async def handle_incoming(
...
         logger.error(f"Outgoing task stopped: {e}")
 
 
+async def observation_update_loop(websocket: WebSocket, robot_control: RobotControlOrchestrator) -> None:
+    """Handle outgoing messages for robot control."""
+    try:
+        while True:
+            async with run_at_frequency(30):
+                try:
+                    observation = robot_control.get_observation()
```
MarkRedeman: nit pick: `get_observation` implies to me we would get a `Observation` object, not a dict. Perhaps we can make this method name more explicit to emphasize that this is expected to be used as a response? (`get_observation_response`?)

application/backend/src/control/utils.py
```diff
@@ -0,0 +1,105 @@
...
+    """Format observation dict into a model-ready Observation object."""
+    from physicalai.data import Observation
+
+    images: dict = {}
+    for camera in manifest.cameras:
+        camera_name = camera.name.lower()
+        # HWC → CHW, float 0..1 range.
+        images[camera_name] = np.ascontiguousarray(
+            observation["images"][camera.id].transpose(2, 0, 1).astype(np.float32)[np.newaxis] / 255
+        )
```
MarkRedeman, line 80: Not specific to this pr, but should this not be done as a preprocessing step inside of the model?

Would be good to add a comment referencing documentation for this.

application/backend/src/control/environment_integration.py
```diff
@@ -1,192 +1,82 @@
...
+
+    manifest: EnvironmentDataRegistry | None = None
+
+    def __init__(
+        self,
+        environment: EnvironmentWithRelations,
+        robot_client_factory: RobotClientFactory,
+        mp_terminate_event: EventClass,
+    ):
+        self.robot = environment.robots[0]
```
AlbertvanHouten, line 25: I have had this line throw errors before when the robots have somehow disappeared from the environment. Not sure what exactly caused this but it then spams the UI with list index out of range errors. Can we add a defensive check here to make sure the robots list is properly initialized?

application/backend/src/control/data_registry.py
```diff
@@ -0,0 +1,29 @@
...
+from typing import Any
+
+
+@dataclass
+class CameraRegistryEntry:
+    id: str
+    name: str
+    width: int
+    height: int
+    frame_data: Any  # mp.Array[c_uint8], shared with CameraWorker
```
AlbertvanHouten: You can use `SynchronizedArray` from `multiprocessing.sharedctypes` for typing here and `Synchronized` for action_read_state in RobotRegistryEntry

application/backend/src/workers/model_integration_worker.py
```diff
@@ -0,0 +1,106 @@
...
+        super().__init__(stop_event=mp_terminate_event, queues_to_cancel=[])
+        self._child_workers: list[BaseProcessWorker] = []
+        self.loaded_event = mp.Event()
+        self.data_manifest = data_manifest
+        self.inference_device = inference_device
+        self.model = model
+        self.event_queue = event_queue
+        self.model_integration = None
+        self.is_running = False
+        self.fps = 30  # TODO FPS
```
AlbertvanHouten: The recording worker has this FPS defined as a global variable. Should we do the same for ModelIntegrationWorker?

application/backend/src/workers/robot_control_orchestrator_worker.py
```diff
@@ -0,0 +1,237 @@
...
+
+MESSAGE_QUEUE_FREQUENCY = 10
+
+
+class RobotControlOrchestrator(BaseThreadWorker):
+    ROLE = "RobotControlOrchestrator"
+
+    recording: RecordingWorker | None = None
+    model_integration: ModelIntegrationWorker | None = None
+    environment_integration: EnvironmentIntegration | None = None
```
AlbertvanHouten: Why class variables over instance variables?

application/backend/src/workers/robot_control_orchestrator_worker.py
```diff
@@ -0,0 +1,237 @@
...
+from schemas import InferenceDevice, Model
+from schemas.dataset import Dataset
+from schemas.environment import EnvironmentWithRelations
+from workers.base import BaseThreadWorker, run_at_frequency
+from workers.model_integration_worker import ModelIntegrationWorker
+from workers.recording_worker import RecordingWorker
+
+
+class RobotControlState(BaseModel):
+    task: str | None = None
```
AlbertvanHouten, line 22: Task is unused

application/backend/src/workers/recording_worker.py
```diff
@@ -0,0 +1,116 @@
...
+from schemas.dataset import Dataset
+from workers.base import BaseProcessWorker, run_at_frequency
+
+RECORDING_FPS = 30
+
+
+class RecordingWorker(BaseProcessWorker):
+    ROLE = "RecordingWorker"
+
+    recording_mutation: RecordingMutation | None = None
```
AlbertvanHouten, line 19: This should also be an instance variable instead of a class variable.

RHeckerIntel:
stale

PR #745, opened 2026-06-26
feat(RT 2/5): add standalone trainer microservice for remote training

AlbertvanHouten:
## Summary

Add a standalone **trainer microservice** (`application/trainer/`) that runs training jobs out-of-process. This is the service the backend offloads to when `TRAINING_MODE=remote`.

## What changed

New self-contained package with its own `pyproject.toml`/`uv.lock`:

- `api.py` — HTTP API (submit/query/cancel jobs, stream artifacts).
- `queue_worker.py` — single-job queue worker.
- `runner.py` — training execution wrapper around `physicalai.train`.
- `store.py` — job state/artifact store.
- `devices.py` — local device discovery.
- `schemas.py`, `settings.py`, `main.py`.
- Tests for devices, queue worker, schemas, and store.

## Why

Decoupling training execution lets recording-only / lightweight backend nodes offload heavy VLA training to a dedicated host with GPUs.

## Notes

- 2nd PR in the remote-training stack. **Base: `albert/rt-1-lib-callback`** (uses the centralized callback).
- The large insertion count is mostly the generated `uv.lock`.

## Test plan

- [ ] `uv run pytest application/trainer/tests`

### DIFF
--- a/application/trainer/README.md
+++ b/application/trainer/README.md
@@ -1,0 +1,63 @@
+# Physical AI Trainer
+
+Standalone remote training service for Physical AI Studio. Runs the heavy
+torch/`physicalai` training stack on a GPU server so recording nodes stay
+lightweight.
+
+## How it fits together
+
+1. The studio backend (`TRAINING_MODE=remote`) pushes the dataset snapshot to an
+   ephemeral private HuggingFace dataset repo and submits a job here.
+2. This service queues the job, pulls the snapshot at a pinned commit SHA,
+   trains, exports, and zips the model.
+3. The backend polls progress, downloads the archive, imports it as a model,
+   then deletes the ephemeral repo.
+
+## Install
+
+```bash
+cd application/trainer
+uv sync --extra cuda   # or --extra cpu / --extra xpu
+```
+
+The `cpu` and `cuda` extras include `executorch`, enabling the ExecuTorch
+export backend. The `xpu` extra omits it: executorch conflicts with the xpu
+torch build, so ExecuTorch export is skipped on xpu installs.
+
+## Configure
+
+Set environment variables (or an `.env` file):
+
+| Variable                     | Required | Description                                  |
+| ---------------------------- | -------- | -------------------------------------------- |
+| `HF_TOKEN`                   | yes      | **Read** access to the snapshot repos. The Studio backend that pushes them needs **write** access. See [token permissions](../backend/docs/huggingface_integration.md#required-token-permissions). |
+| `STORAGE_DIR`                | no       | Working directory for jobs and artifacts.    |
+| `TRAINER_MAX_CONCURRENT_JOBS`| no       | Queue concurrency (default 1).               |
+| `TRAINER_DEVICE`             | no       | Force `cuda`/`xpu`/`cpu` (auto if unset).    |
+| `PORT`                       | no       | Listen port (default 8001).                  |
+
+Never commit `HF_TOKEN`. Store it in a secret manager or local `.env`.
+
+## Run
+
+```bash
+uv run python -m trainer.main
+```
+
+## API
+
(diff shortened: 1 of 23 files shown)


ArendJanKramer (approved):

application/trainer/README.md
```diff
@@ -0,0 +1,63 @@
...
+uv sync --extra cuda   # or --extra cpu / --extra xpu
+```
+
+The `cpu` and `cuda` extras include `executorch`, enabling the ExecuTorch
+export backend. The `xpu` extra omits it: executorch conflicts with the xpu
+torch build, so ExecuTorch export is skipped on xpu installs.
+
+## Configure
+
+Set environment variables (or an `.env` file):
```
AlexanderBarabanov, line 29: As there is no authentication by design, I propose to warn consumer, something like:

> [!IMPORTANT]  
> The trainer service has no built-in authentication. Any machine that can reach its port can submit jobs, cancel jobs, and download trained model artifacts. 
> Deploy it on a private network and allow only the Physical Studio backend IP address to reach it. Do not expose this port to the internet.

application/trainer/src/trainer/runner.py
```diff
@@ -0,0 +1,191 @@
...
+        *,
+        should_stop: StopFn,
+        report: ProgressFn,
+    ) -> None:
+        from lightning.pytorch.callbacks import ModelCheckpoint
+        from lightning.pytorch.loggers import CSVLogger
+        from physicalai.data import LeRobotDataModule
+        from physicalai.train import ProgressReportingCallback, Trainer
+
+        payload = request.payload
```
AlexanderBarabanov, line 89: Wondering if we need to add some validation for this payload, at least for this input values as this may potentially cause denial of service:
- `max_steps` - potentially runs training indefinitely
- `batch_size` - out of memory (?)
- `num_workers` - spawns hundreds of processes
- `precision` - may produce confusing Lightning errors (?)

application/trainer/README.md
AlbertvanHouten: Will add it in PR 3/5

application/trainer/src/trainer/runner.py
AlbertvanHouten: I think this will be a bit hard to validate as these values differ depending on the hardware. A batch size of x on GPU A might work without issue but can be OOM on GPU B. Same for max_steps/num_workers.

library/src/physicalai/train/callbacks.py
```diff
@@ -4,12 +4,179 @@
...
+                None,
+                {"val_event": "batch", "val_batch": current, "val/loss_step": self._extract_loss(outputs)},
+            )
+        self._check_stop(trainer)
+
+    def on_validation_epoch_end(self, trainer: L.Trainer, _pl_module: L.LightningModule) -> None:
+        """Report the validation summary with aggregated loss and elapsed time."""
+        val_loss = trainer.callback_metrics.get("val/loss")
+        val_loss_val = val_loss.item() if val_loss is not None else None
+        elapsed = time.monotonic() - self._val_start_t if self._val_start_t is not None else 0.0
```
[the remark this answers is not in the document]
bsgrigorov: Yes — before this change Phase 1 would run through approval + prod builds and fail in `release-create-gh-release.sh` on tag/release SHA conflict; with `skip_publish`, Phases 2–3 could pull assets from the wrong release.

Fixed in e0c1cae79a: when `execute_store_phases=true` and `v{version}` targets a different commit than the workflow SHA, Phase 0 fails fast with `::error::`. With store phases off (default), it remains a warning only.

Runbook: `.github/workflows/runway-extension-release-and-submit.md`.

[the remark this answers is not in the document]
AlbertvanHouten: ??

PR #746, opened 2026-06-26
refactor(RT 3/5): introduce pluggable training backend abstraction

AlbertvanHouten:
## Summary

Introduce a pluggable **training backend abstraction** and refactor the existing in-process training onto it. Behavior-preserving for local training.

## What changed

- New `services/training_backends/` package: `base.py` (contract: `TrainingBackend`, `TrainingContext`, `ProgressReporter`, `TrainingCanceledError`), `local.py` (in-process torch/Lightning backend), `_log_format.py` (shared job-log formatting), `__init__.py` (`get_training_backend()` selector — local-only here; remote selection arrives in the next PR).
- Refactor `training_service.py` and `training_worker.py` to drive the backend interface.
- Add `TRAINING_MODE` and trainer settings to `settings.py`.
- Clean tqdm/ANSI progress output in `core/logging/handlers.py`.
- Move tests to `test_local_training_backend.py`, add `test_handlers.py`; remove the old `test_training_service.py`.
- Added github action workflow for tests and code quality checks 

## Why

A single seam for "where training runs" keeps the worker agnostic and lets the remote backend slot in without touching call sites.

## Notes

- 3rd PR in the stack. **Base: `albert/rt-2-trainer-service`.**
- `__init__.py` ships a local-only selector to stay self-consistent; the remote branch is added in PR 4.

## Test plan

- [ ] `uv run pytest application/backend/tests/services/test_local_training_backend.py application/backend/tests/test_settings.py application/backend/tests/workers/test_training_worker.py application/backend/tests/core/logging/test_handlers.py`


application/backend/src/settings.py
```diff
@@ -111,6 +111,28 @@ def log_dir(self) -> Path:
...
+    trainer_hf_namespace: str | None = Field(default=None, alias="TRAINER_HF_NAMESPACE")
+    # Seconds to wait for trainer HTTP requests (excludes long-poll/SSE streams).
+    trainer_request_timeout_s: float = Field(default=30.0, alias="TRAINER_REQUEST_TIMEOUT_S")
+    # Seconds to wait between chunks while streaming the model artifact. A stalled
+    # transfer (e.g. a proxy holding the connection open) must fail instead of
+    # hanging the job forever; this is a per-read gap, not a total transfer cap.
+    trainer_download_read_timeout_s: float = Field(default=120.0, alias="TRAINER_DOWNLOAD_READ_TIMEOUT_S")
+
+    @model_validator(mode="after")
+    def validate_remote_training_config(self) -> "Settings":
```
AlexanderBarabanov, line 130: Can we additionally validate that provided `trainer_url` has `http://` or `https:// `scheme with host?
Perhaps we can utilize validation via `from pydantic import AnyHttpUrl`

AlbertvanHouten: Fixed

AlexanderBarabanov (dismissed):

application/backend/src/services/training_service.py
```diff
@@ -83,147 +29,30 @@ def __init__(self, job_id: UUID, event_queue: mp.Queue, interrupt_event: Event):
...
-    def on_fit_start(self, trainer: "pl.Trainer", pl_module: "pl.LightningModule") -> None:  # noqa ARG002
-        """Resolve logging interval once trainer values are available."""
-        self.every_n_steps = self._auto_every_n_steps(trainer.max_steps)
-
-        logger.info(
-            f"Training log cadence configured: every_n_steps={self.every_n_steps}, max_steps={trainer.max_steps}"
+        while await self._drain_one():
+            pass
+
+    async def _drain_one(self) -> bool:
```
MarkRedeman, line 37: nit picky: `_drain_one` doesn't feel as descriptive to me, maybe `_update_training_status()` would be nicer? (perhaps move the `asyncio.sleep` into here?

application/backend/src/services/training_service.py
```diff
@@ -83,147 +29,30 @@ def __init__(self, job_id: UUID, event_queue: mp.Queue, interrupt_event: Event):
...
-        self.every_n_steps = 1
-        self._val_start_t: float | None = None
-
-    def on_fit_start(self, trainer: "pl.Trainer", pl_module: "pl.LightningModule") -> None:  # noqa ARG002
-        """Resolve logging interval once trainer values are available."""
-        self.every_n_steps = self._auto_every_n_steps(trainer.max_steps)
-
-        logger.info(
-            f"Training log cadence configured: every_n_steps={self.every_n_steps}, max_steps={trainer.max_steps}"
+        while await self._drain_one():
```
MarkRedeman, line 34: Why is there another drain here?

application/backend/src/services/training_backends/local.py
```diff
@@ -0,0 +1,151 @@
...
+        shutil.move(str(cache_path), str(output_dir))
+
+        export_policy = policy
+        if payload.compile_model and context.model.policy in ("act", "smolvla"):
+            try:
+                logger.info("Reloading non-compiled policy for export")
+                export_policy = load_policy(context.model, compile_model=False)
+            except Exception as exc:
+                logger.warning("Failed to reload non-compiled policy for export; using trained policy")
+                logger.exception(exc)
```
MarkRedeman, line 103: Just for context (no need to address this): the reason we do this here is that both SmolVLA and Act where throwing errors when we'd try to export their compiled model to OpenVINO, ONNX and ExecuTorch.
The issue didn't happen for pi05.

application/backend/src/services/training_backends/local.py
```diff
@@ -0,0 +1,151 @@
...
+
+    from lightning.pytorch.callbacks import Callback
+
+    from services.training_backends.base import TrainingContext
+
+
+class LocalTrainingBackend:
+    """Train in the worker process with Lightning."""
+
+    async def train(self, context: TrainingContext) -> None:
```
MarkRedeman, line 32: Perhaps "train" is a bit misleading here since we also export at the end of the training.
Not a big deal but something we could look into later.

One thing I've been thinking of is that we could add a list of exports to `TrainingContext`. So for instance a user can fill in a form in the UI so that they export to Torch, OpenVINO FP16 and OpenVINO Int8 after training (but skip the others..?)

No need to address this here, mainly wanted to share the idea.

application/trainer/.pre-commit-config.yaml
```diff
@@ -0,0 +1,37 @@
+# Copyright (C) 2026 Intel Corporation
+# SPDX-License-Identifier: Apache-2.0
+
+# Trainer-specific pre-commit configuration for prek workspace mode
+# This config applies only to files in application/trainer/
+
+repos:
```
MarkRedeman, line 12: Can this be configured to extend the other pre commit config? Having pyrefly in here would be nice.

MarkRedeman (dismissed):

MarkRedeman (approved):

application/backend/src/services/training_service.py
AlbertvanHouten: Its a guard against race conditions to guarantee no queued update is lost when the dispatcher is asked to stop.

AlexanderBarabanov (approved):

PR #747, opened 2026-06-26
feat: add remote training backend and device discovery

AlbertvanHouten:
## Summary

Add the **remote training backend** and remote device discovery, wiring the backend service to the trainer microservice. Also make the backend install lightweight so recording-only nodes skip heavy training deps.

## What changed

- `training_backends/remote.py` — offloads jobs to the trainer service: snapshot dataset upload, progress mapping, artifact streaming with per-read timeouts, cancellation.
- `__init__.py` selector now returns the remote backend when `TRAINING_MODE=remote`.
- Remote device discovery with local fallback: `system_service.py`, `api/system.py`.
- Lazy-import heavy deps (`models/utils.py`, `workers/model_worker.py`) and split `physicalai-train[...]`/`executorch` behind a `train` extra in `pyproject.toml`.
- `run.sh` startup for the remote mode; `.dockerignore`; log-level tweak in `event_processor.py`.
- Tests: remote backend, backend selection, log-format mirroring, system service.

## Why

Completes the offloaded training path and keeps the recording/remote install free of in-process torch training weights.

## Notes

- 4th PR in the stack. **Base: `albert/rt-3-training-backend-abstraction`** (also depends conceptually on the trainer service in PR 2).

## Test plan

- [ ] `uv run pytest application/backend/tests/services/test_remote_training_backend.py application/backend/tests/services/test_training_backends.py application/backend/tests/services/test_training_log_format.py application/backend/tests/services/test_system_service.py`


application/backend/run.sh
```diff
@@ -2,15 +2,198 @@
...
+#   UV_CMD                 Launch command (default: "uv run --no-sync").
+#
+# Requirements:
+# - 'uv' CLI installed and available in PATH.
 # -----------------------------------------------------------------------------
 
-export PYTHONUNBUFFERED=1
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+
+# Load KEY=VALUE pairs from a .env file into the environment. Variables already
```
AlexanderBarabanov: Minor. 

If `.env` is world-readable (e.g., `chmod 644`), any user on the same host can read `HF_TOKEN`. I propose to add permission check and warning message to consumer.
Something like: 

```bash
load_env_file() {
    local env_file="$1"
    [[ -f "$env_file" ]] || return 0
    local perms
    perms=$(stat -c "%a" "$env_file" 2>/dev/null || stat -f "%p" "$env_file" 2>/dev/null | tail -c 4)
    if [[ "$perms" != "600" && "$perms" != "400" ]]; then
        echo "Warning: ${env_file} is readable by others (mode ${perms}). Consider: chmod 600 ${env_file}" >&2
    fi
    ...
}
```

application/backend/src/services/training_backends/remote.py
```diff
@@ -0,0 +1,509 @@
...
+
+    def _apply_state(self, context: TrainingContext, state: dict[str, Any]) -> bool:
+        """Mirror a job state into the local job; return True if completed.
+
+        Raises ``TrainingCanceledError`` / ``RemoteTrainingError`` on terminal
+        cancellation / failure.
+        """
+        status = state.get("status")
+        remote_progress = self._coerce_progress(state.get("progress"))
+        raw_extra = state.get("extra_info")
```
AlexanderBarabanov, line 390: Can this `extra_info` dict exhaust memory in the dispatcher queue or disk space in the backend job database?
If yes, maybe we can add some size guard.

application/backend/src/services/training_backends/remote.py
```diff
@@ -0,0 +1,509 @@
...
+        """Push snapshot, submit job, mirror progress, and ingest the model."""
+        repo_id: str | None = None
+        try:
+            # Sub-step 1: push the snapshot to an ephemeral private dataset repo (0-10%).
+            context.progress(0, message="Uploading dataset snapshot")
+            repo_id, revision = await self._push_snapshot(context)
+            context.progress(SNAPSHOT_UPLOAD_PROGRESS, message="Snapshot uploaded")
+
+            # Sub-step 2: submit and wait for the remote job (10-95%).
+            remote_job_id = await self._submit_job(context, repo_id=repo_id, revision=revision)
```
AlexanderBarabanov, line 157: Can we add some pattern validation here for `remote_job_id ` value ?

application/backend/src/services/training_backends/remote.py
```diff
@@ -0,0 +1,509 @@
...
+                cached = await self._probe_proxy()
+                self._trust_env = cached
+        return cached
+
+    async def _probe_proxy(self) -> bool:
+        """Return True if /health is reachable with proxy env vars honored."""
+        try:
+            # AsyncClient honors HTTP_PROXY/HTTPS_PROXY by default (trust_env).
+            async with httpx.AsyncClient(timeout=httpx.Timeout(5.0)) as client:
+                response = await client.get(f"{self._base_url}/health")
```
AlexanderBarabanov, line 109: Backend picks up `HTTP_PROXY` / `HTTPS_PROXY` from the host env automatically, which means whoever controls those variables on the machine also controls where model artifacts go. 

I propose to add a note under security section in `application/docs/08-remote-training-server.md` ([PR](https://github.com/open-edge-platform/physical-ai-studio/pull/748/)) to highlight this risk for consumer, something like:

> - The Physical AI Studio backend inherits proxy settings (`HTTP_PROXY` / `HTTPS_PROXY`) from its host environment. When those variables are set, all trainer communication (including model artifact download) routes through the configured proxy.  Only deploy the backend in a fully trusted environment where proxy variables cannot be set by other users. Do not run it on a shared or multi-tenant host.

application/backend/src/services/training_backends/remote.py
AlbertvanHouten: added

application/backend/src/services/training_backends/remote.py
AlbertvanHouten: message is capped at 16 kb now

application/backend/run.sh
MarkRedeman, line 1: Do we want to keep this in `run.sh`, or move it into subcommands of the `physicalai-studio` cli?
I'd personally prefer to see it as a separate cli command

AlbertvanHouten: Yeah good point. I updated it to be a cli command

PR #755, opened 2026-06-29
feat: disable conflicting resource deletion

maxxgx:
# Pull Request

## Description

Deleting a robot or camera that was still referenced by an environment silently corrupted the environment. The DB cascade removed the join row (and, for a teleoperator leader, the follower's row too), leaving the environment pointing at things that no longer existed. This surfaced later as a 500 in the robot control websocket ("Robot could not be found").

This PR makes robots and cameras that are in use by an environment non-deletable, and gives the user clear feedback instead of a crash.

### Backend

- **Schema**: changed the robot/camera foreign keys on the environment join tables from `ON DELETE CASCADE` to `ON DELETE NO ACTION` (`environment_robots.robot_id`, `environment_robots.tele_operator_robot_id`, `environment_cameras.camera_id`). The `environment_id` FKs keep `CASCADE`, so deleting an environment still removes its links.
- **Migration** (`a7c1e9f4b2d3`): rebuilds both join tables with the new FK actions, preserving existing data. SQLite cannot alter a foreign key in place, so the tables are recreated. Reversible.
- **Delete guards**: `delete_robot` and `delete_camera` now return a 409 `ResourceInUseError` listing the environments that use the resource, instead of letting the DB error bubble up as a 500.

### Frontend

- Robot and camera delete now surface the backend message via toast.
- The `409` in-use case shows an informational toast (blue), since it is an expected and recoverable state, not an application failure. Genuine errors still show a negative toast (red).

## Breaking Changes

---

## Examples/Screenshots

Deleting a robot that is currently referenced by an environment will result in the following message:
image


application/backend/src/alembic/versions/a7c1e9f4b2d3_restrict_env_robot_camera_fks.py
```diff
@@ -0,0 +1,106 @@
...
+    cameras = conn.execute(sa.text("SELECT environment_id, camera_id FROM environment_cameras")).fetchall()
+
+    # Disable FK enforcement during the rebuild: dropping/recreating the child tables would otherwise
+    # trip constraint checks mid-operation.
+    conn.execute(sa.text("PRAGMA foreign_keys = OFF"))
+
+    op.drop_table("environment_robots")
+    op.drop_table("environment_cameras")
+    _create_environment_robots(robot_ondelete)
+    _create_environment_cameras(camera_ondelete)
```
MarkRedeman, line 73: Would it not be possible to update the foreign key constraints individually instead of recreating the tables?

application/backend/src/repositories/project_environment_repo.py
```diff
@@ -99,3 +99,33 @@ def _build_robots_with_teleoperators(robot_links: list[EnvironmentRobotDB]) -> l
             robots_with_teleoperators.append(RobotWithTeleoperator(robot=robot, tele_operator=tele_operator))
 
         return robots_with_teleoperators
+
+    async def find_environment_names_using_robot(self, robot_id: UUID) -> list[str]:
+        """Return names of environments in this project that reference the robot (as robot or leader)."""
+        rid = str(robot_id)
+        stmt = (
+            select(ProjectEnvironmentDB.name)
```
MarkRedeman, line 107: What if there's an environment with a duplicated name, should we instead use the environment id here?

MarkRedeman: Nvm, I see that this is only used as a way to tell the user which environments are in use.

application/backend/src/alembic/versions/a7c1e9f4b2d3_restrict_env_robot_camera_fks.py
MarkRedeman: @copilot is it possible to instead only remove the foreign key constraints?

maxxgx: Issue with Alembic drop_constraint -> create_foreign_key:

> it's not clearly better for this case, for one concrete reason: your FKs are unnamed (the original `create_table` didn't name them), so `drop_constraint` has nothing to target, and you'd have to fight SQLite reflection or add naming conventions. For changing FK actions specifically, the explicit manual drop/recreate/reinsert you have is actually clearer and more reliable.

It seems it also recreates the table under the hood, so I think we keep this migration

MarkRedeman (approved):

PR #757, opened 2026-07-01
fix: fix teleoperation from a robot without velocities

MarkRedeman:
Before when teleoperating a follower robot that records `.vel` states while using a leader that only uses `.pos` states the teleoperate worker would crash due to a missing index.

This specific scenario occurred while testing the `rebot-b601-dm` (follower) arm with a `stararm` (leader). The rebot arm does record velocities but the stararm does not.

### DIFF
--- a/application/backend/src/workers/teleoperate_worker.py
+++ b/application/backend/src/workers/teleoperate_worker.py
@@ -88,7 +88,7 @@ class TeleoperateWorker(BaseProcessWorker):
         with self._action_read_state.get_lock():
             self._action_read_state.value = value
 
-    def _feature_values(
+    def _align_feature_values(
         self,
         source_state: dict[str, float],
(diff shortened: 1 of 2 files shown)


application/backend/src/workers/teleoperate_worker.py
```diff
@@ -88,6 +88,26 @@ def set_action_read_state(self, value: ActionReadState) -> None:
         with self._action_read_state.get_lock():
             self._action_read_state.value = value
 
+    def _feature_values(
```
ArendJanKramer: maybe call it _align_feature_values, or worse, _pad_feature_values

MarkRedeman: Renamed to `_align_feature_values`

ArendJanKramer (approved):

PR #759, opened 2026-07-02
chore: remove unused calibrations endpoint

MarkRedeman:
This endpoint hasn't been used for some time, calibrations are stored per robot not per file (see #194 for the initial implementation)


AlbertvanHouten (approved):

application/backend/src/utils/calibration.py
AlbertvanHouten, line 1: 0 tests apparently :/

MarkRedeman: Yeah we will need to up our game here. A lot of the initial work on studio was us discovering how the UX etc should be. Now we're focusing on extending the system and should also focus on stability.

PR #760, opened 2026-07-02
fix: use connection string (usb port) as a fallback if robot has no serial number

MarkRedeman:
Some robots don't have a serial number, in which case we use its connection string (usb port) as a fallback.

In particular this will fix connection issues with virtual usbs, with the rebot and stararm as well as other so101 arms that use a control board that our team didn't test before.

### DIFF
--- a/application/backend/src/api/robot_setup.py
+++ b/application/backend/src/api/robot_setup.py
@@ -5,7 +5,8 @@ from typing import Annotated
 from fastapi import APIRouter, Depends, WebSocket, status
 from loguru import logger
 
-from api.dependencies import get_project_id
+from api.dependencies import RobotConnectionManagerDep, get_project_id
+from schemas import SerialPortInfo
 from schemas.robot import RobotType
 from workers.robots.so101_setup_worker import SO101SetupWorker
 from workers.transport.websocket_transport import WebSocketTransport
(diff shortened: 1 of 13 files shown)


MarkRedeman:
Note that one issue that this does bring is that if the user uses a robot without serial number that they then may not be able to reconnect to the robot after a reboot or after unplugging and plugging back in the usb port.

This is because the server will only store the usb port if the robot has no serial number. And the usb port might change on reboot/re-plugging in the usb port.
What we will need to do as a follow up is allow users to edit their robot's configuration. I will create a ticket for this with some ideas.

application/backend/src/api/robot_setup.py
```diff
@@ -16,15 +17,18 @@
 @router.websocket("/setup/ws")
 async def robot_setup_websocket(
     _project_id: Annotated[str, Depends(get_project_id)],
+    robot_manager: RobotConnectionManagerDep,
     websocket: WebSocket,
     robot_type: str,
-    serial_number: str,
+    serial_number: str = "",
+    connection_string: str = "",
```
AlbertvanHouten, line 24: It is cleaner to set these values to None instead of using an empty string (requires updating the below check too)


```suggestion
    serial_number: str | None= None,
    connection_string: str | None = None
```

AlbertvanHouten (approved):

application/backend/src/api/robot_setup.py
MarkRedeman: I think this needs a minor change in how we enable the websocket connection (because atm we always send all query parameters). Will set this up in a small follow up pr.

PR #764, opened 2026-07-02
refactor: move urdf file serving from frontend to backend

MarkRedeman:
This PR adds two new API endpoints:
- `/api/robots/catalog/{robot_type}/urdf`
- `/api/robots/catalog/{robot_type}/{asset_path:Path}`

The first is used to return a URDF file for a specific robot type (i.e. `SO101_Leader`), the UI fetches this file and then fetches the related meshes using the second endpoint.

To make the backend serve these files I've moved the `npm run clone-so101` script (and widowx) into the backend via a new cli command.
When the server starts it will check if the so101 and widowx files are available to the backend, if not we will download these.

Later when we add robot plugins we will add extra discoverability in these endpoints so that they can search for urdf files inside of a package.


application/backend/src/robots/catalog/sync_robot_assets.py
```diff
@@ -0,0 +1,56 @@
...
+
+import shutil
+import subprocess
+import tempfile
+from pathlib import Path
+
+from robots.catalog.assets import get_builtin_robot_assets_root
+
+SO101_REPO_URL = "https://github.com/TheRobotStudio/SO-ARM100.git"
+WIDOWX_REPO_URL = "https://github.com/TrossenRobotics/trossen_arm_description.git"
```
ArendJanKramer, line 12: Didn't we patch the joint names in Trossen URDF?

MarkRedeman: No we only clone them from those repos. In the UI we map the feature names to joint names using https://github.com/open-edge-platform/physical-ai-studio/blob/main/application/ui/src/features/robots/robots-configuration.ts#L39

After this PR I'll start moving all of this frontend logic into the backend so that it will be fully responsible for maintaining this. 
That will then allow us to more easily add plugins as we only need to do this in the backend.

application/backend/src/cli/serve.py
```diff
@@ -40,3 +45,15 @@ def serve(host: str, port: int) -> None:
...
+def _sync_missing_robot_assets() -> None:
+    if builtin_robot_assets_are_available():
+        return
+
+    click.echo("Robot assets are missing; syncing them now...")
+    try:
+        sync_robot_assets()
+    except (subprocess.CalledProcessError, OSError) as error:
+        click.echo(f"✗ Failed to sync robot assets: {error}")
+        sys.exit(1)
```
AlbertvanHouten, line 59: Why is this a CLI command? Shouldn't we always sync these assets when the app starts or when the user requests these specific assets?

MarkRedeman: I mainly figured I would follow the same setup we had before, but moved to python instead of using npm.

The way it's setup now is that we do check this when `serve` is run, but only if the assets are missing.

In the [rebot-601 plugin](https://github.com/MarkRedeman/physicalai-rebot-b601-plugin/tree/main/urdf) I added the urdf files inside of the package which get distributed together with the source code.
I expect we will do the same for all other robots that we will add so we this CLI command is mainly a unique thing for SO101 and WidowX AI. We could later on also move those to a separate pacakge and then remove this command?

application/docker/Dockerfile
```diff
@@ -268,6 +256,7 @@ RUN uv pip install --python /app/application/backend/.venv/bin/python --no-deps
 
 WORKDIR /app/application/backend
 USER appuser
+RUN physicalai-studio sync-robot-assets
 CMD ["./run.sh"]
```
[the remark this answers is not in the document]
MarkRedeman: Fixed by now cloning specific revision ids.

application/ui/src/features/robots/robots-configuration.ts
```diff
@@ -18,14 +19,9 @@ export const isLeader = (robot: SchemaRobot) => {
...
-        return '/widowx/urdf/generated/stationary_ai.urdf';
-    }
-
-    if (robotType !== undefined && robotType.toLowerCase().includes('trossen')) {
-        return '/widowx/urdf/generated/wxai/wxai_follower.urdf';
-    }
-    return '/SO101/so101_new_calib.urdf';
+    return fetchClient.PATH('/api/robots/catalog/{robot_type}/urdf', {
+        params: { path: { robot_type: robotType } },
+    });
```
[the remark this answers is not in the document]
MarkRedeman: Fixed by hardcoding the packages for widowx/trossen. In a follow up PR we'll be adding a `/api/robots/catalog` endpoint that provides this info per robot.

application/backend/src/cli/serve.py
MarkRedeman: Another specific reason why I figured having it as CLI would be nice is that we can then pre-download these files in the docker images.
I think we could do the same for the `physicalai-studio` python package?

AlbertvanHouten (dismissed):

AlbertvanHouten (approved):

PR #765, opened 2026-07-02
fix: use query to load urdf models

MarkRedeman:
Before we were using a mutation which would get called multiple times when viewing a robot. This made loading robots slow as we'd fetch the same urdf file multiple times, resulting in both http fetches as well as rerenders.

Test this by looking at the network tab in the developer tools.


application/ui/src/features/robots/robot-models-context.tsx
```diff
@@ -89,57 +89,61 @@ export const useRobotModels = () => {
...
-            setModel(pathRef.current, model);
-        },
-        meta: { skipInvalidation: true },
+export const useLoadModelQuery = (path: string) => {
+    const { getModel, setModel } = useRobotModels();
+
+    const cachedModel = getModel(path);
+    const query = useQuery({
+        queryKey: ['robotModel', path],
+        queryFn: () => cachedModel ?? loadURDFModel(path),
```
maxxgx, line 98: Nitpick: isn't `queryFn` only executed if `enabled: true`? Then `cachedModel ??` is unnecessary

maxxgx (approved):

application/ui/src/features/robots/robot-models-context.tsx
MarkRedeman: Yeah good point, it doesn't hurt to have it there (I'd need to double check but it might be possible that `queryFn` still gets called if you'd manually trigger the query via it's `queryKey`, I'm not sure though)
I will keep it as is for now and maybe revisit it again after some other refactor work.

PR #768, opened 2026-07-03
feat: RLDX-1 Policy

eugene123tw:
# Pull Request

## Description

Adds **RLDX-1** (RLWRLD's Qwen3-VL + flow-matching action-expert VLA) as a new policy family in `physicalai-train`, and adds **RoboCasa Kitchen** as a new benchmark gym, which serves as the primary parity target for the RLDX-1 `FT-ROBOCASA` checkpoint.

Key changes:

- **New policy: `library/src/physicalai/policies/rldx1/`** — config/model/policy/preprocessor split following the standard Physical AI Studio contract (`Policy(L.LightningModule)`, `predict_action_chunk`, `Observation`). Includes:
  - Qwen3-VL-based vision-text backbone adapter (`components/backbone/`) with VTC multi-frame video support (4 frames/step, strides `[-6,-4,-2,0]`).
  - MSAT flow-matching action head + attention/blocks (`components/action_model/`).
  - Native, PAS-first preprocessing pipeline (`preprocessing.py`, `augmentations.py`) with a bit-exact `AspectAreaResizeAndCrop` image transform, replacing the vendored RLDX-1 processor as the live default path (vendored code retained only as a parity oracle under `tests/`).
  - LoRA (PEFT) fine-tuning support and lazy imports for optional heavy deps (`diffusers`, `albumentations`) to keep base import cost low.
  - Config-driven `embodiment_id` (int or embodiment-tag string) with validation, defaulting to `general_embodiment` (slot 0), matching how all released `FT-*` checkpoints were trained.
- **New benchmark gym: `physicalai.benchmark.gyms.robocasa`** — `RoboCasaBenchmark`/`RoboCasaGym`, configurable state/action ordering, plus a new `library/configs/benchmark/robocasa.yaml` and install script (`bash library/scripts/benchmark/install_robocasa.sh`) since RoboCasa requires a dedicated virtualenv.
- **Shared components refactor**: extracted reusable NN building blocks from `policies/groot/components/nn.py` into `policies/shared/components/` so RLDX-1 (and future policies) can reuse them instead of duplicating GR00T-specific code.
- **Packaging/CI**: new `rldx1` extra in `library/pyproject.toml` (transformers, safetensors, Pillow, albumentations, scipy, accelerate, peft, diffusers), added to the `all` extra, and wired into `uv sync` steps in `.github/workflows/library.yml` (CPU + CUDA unit test jobs). Pre-commit `pyrefly-check` now runs after an explicit `uv sync --group dev` since it uses `--no-sync`.
- **Docs**: `library/README.md` benchmark section now documents RoboCasa; `library/docs/explanation/gyms/robocasa.md` and a new `library/docs/getting-started/first-benchmark.md` guide added/updated. Agent skill docs (`skills/library/**`) touched for consistency.
- **Tests**: new unit tests for the RoboCasa gym/benchmark (`library/tests/unit/gyms/test_robocasa.py`, `library/tests/unit/benchmark/test_benchmark.py`) and an extensive RLDX-1 test suite (`library/tests/unit/policies/test_rldx1_preprocessing.py`, `test_rldx1_video_window.py`) with the vendored RLDX-1 reference implementation kept under `library/tests/unit/policies/rldx1_vendored/` as a parity oracle only (not shipped in production code).

## Related Issues

No linked issues — this is a net-new policy/benchmark integration tracked internally as the RLDX-1 integration effort.

## Breaking Changes

None expected for existing users. Notable non-breaking but notable changes:

- `physicalai.policies.groot.components.nn` symbols moved to `physicalai.policies.shared.components` (re-exported/updated import paths within `groot`); external code importing directly from the old `groot/components/nn.py` path would need to update imports.
- CI unit test jobs (CPU + CUDA) now also install the new `rldx1` extra, increasing CI dependency install time slightly.
- RLDX-1 ships under RLWRLD's research/non-commercial model license for the pretrained weights (code itself is Apache-2.0) — commercial users should confirm licensing before using released `RLWRLD/RLDX-1-*` checkpoints.

---

## Examples

Training/using the new RoboCasa benchmark (from `library/README.md`):

```python
from physicalai.benchmark.gyms import RoboCasaBenchmark
from physicalai.policies import ACT

policy = ACT.load_from_checkpoint("experiments/lightning_logs/version_0/checkpoints/last.ckpt")
policy.eval()

benchmark = RoboCasaBenchmark(task="atomic_seen", num_episodes=20)
results = benchmark.evaluate(policy)

print(results.summary())
results.to_json("results.json")
```

Installing the new extras:

```bash
# RLDX-1 policy dependencies
uv sync --extra rldx1

# RoboCasa benchmark (dedicated venv)
bash library/scripts/benchmark/install_robocasa.sh
```


library/docs/getting-started/first-benchmark.md
```diff
@@ -164,6 +164,56 @@ benchmark = PushTBenchmark(num_episodes=50)
...
+**Python API:**
+
+```python test="skip" reason="requires robocasa dedicated venv"
+from physicalai.benchmark.gyms import RoboCasaBenchmark
+from physicalai.policies import ACT
+
+policy = ACT.load_from_checkpoint(
+    "experiments/lightning_logs/version_0/checkpoints/last.ckpt"
+)
+policy.eval()
```
alfieroddan, line 195: Nice example, could we use RLDX-1 though? I imagine it has pre-trained weights that will give some success %?

alfieroddan: Perhaps like in your notebook? Or another model?

library/src/physicalai/eval/rollout/functional.py
```diff
@@ -293,6 +305,13 @@ def run_rollout_loop(  # noqa: PLR0914
     done_mask = torch.zeros(batch_size, dtype=torch.bool)
     success_mask = torch.zeros(batch_size, dtype=torch.bool)
 
+    # Some gyms (e.g. RoboCasa) only know the task description after reset(),
+    # so refresh the caption here rather than relying on a value captured earlier.
+    if video_recorder is not None:
+        description = getattr(env, "task_description", None)
+        if description:
+            video_recorder.caption = description
```
alfieroddan, line 317: If I manually set all captions to be `VideoRecorder(caption="Custom caption")` for a `Gym` (not a benchmark, fine for a benchmark I guess) - wouldn't this silently lose my custom caption?

library/src/physicalai/policies/rldx1/components/action_model/attention.py
```diff
@@ -0,0 +1,199 @@
...
+    """Single transformer block used by the MSAT self-attention stack.
+
+    The block applies layer norm, self-/cross-attention, residual connection,
+    then feed-forward + residual connection.
+    """
+
+    pos_embed: SinusoidalPositionalEmbedding | None
+    final_dropout: nn.Dropout | None
+
+    def __init__(  # noqa: PLR0913, PLR0917
```
alfieroddan:
```python
def __init__(
    self,
    dim: int,
    num_attention_heads: int,
    attention_head_dim: int,
    *,
    dropout: float = 0.0,
    cross_attention_dim: int | None = None,
    activation_fn: str = "geglu",
    attention_bias: bool = False,
    upcast_attention: bool = False,
    norm_elementwise_affine: bool = True,
    norm_type: str = "layer_norm",
    norm_eps: float = 1e-5,
    final_dropout: bool = False,
    attention_type: str = "default",  # noqa: ARG002
    positional_embeddings: str | None = None,
    max_seq_length: int | None = None,
    ff_inner_dim: int | None = None,
    ff_bias: bool = True,
    attention_out_bias: bool = True,
) -> None:
```

The * can get rid of these noqa's

library/src/physicalai/policies/rldx1/components/action_model/blocks.py
```diff
@@ -0,0 +1,1284 @@
...
+            ModulationOut(*out[:3]),
+            ModulationOut(*out[3:]) if self.is_double else None,
+        )
+
+
+# Models ================================================
+class SingleStreamBlock(nn.Module):
+    """Flux-style SingleStreamBlock with parallel linear layers."""
+
+    def __init__(  # noqa: PLR0913, PLR0917
```
alfieroddan, line 67: * again, if you fancy using it - will leave others to your disgression.

library/src/physicalai/policies/rldx1/components/backbone/modeling_vtc.py
alfieroddan, line 1: modelling_vtc.py*

library/src/physicalai/policies/rldx1/preprocessor.py
alfieroddan, line 1: pre-processing is all done in numpy - could be worth only implementing in the physicalai runtime repo. Would save you having to implement twice.

library/src/physicalai/policies/rldx1/model.py
```diff
@@ -0,0 +1,1221 @@
...
+            # Keep the trainable vlln in fp32. Its weight inits to 1.0, where the
+            # bf16 ULP is 2**-7 ~= 0.0078 -- larger than a typical AdamW step
+            # (~lr), so in-place bf16 updates round back to 1.0 and the gain
+            # never moves. fp32 storage lets small updates accumulate.
+            self.vlln.to(torch.float32)
+
+        print(f"[MSAT] Tune action model projector: {self.tune_projector}")
+        print(f"[MSAT] Tune action model diffusion model: {self.tune_diffusion_model}")
+        print(f"[MSAT] Tune action model vlln: {self.tune_vlln}")
+        print(f"[MSAT] Action model LoRA: {use_lora}")
```
alfieroddan, line 267: Wanted or leftover?

library/pyproject.toml
```diff
@@ -238,6 +251,8 @@ project-excludes = [
     "src/physicalai/gyms/robocasa.py",
     "src/physicalai/policies/pi05/pi_gemma.py",
     "src/physicalai/policies/groot/components/backbone.py",
+    "src/physicalai/policies/rldx1/augmentations.py",
+    "src/physicalai/policies/rldx1/model.py",
```
alfieroddan: Why excluded?

library/.pre-commit-config.yaml
alfieroddan, line 1: maybe different PR?

library/src/physicalai/policies/rldx1/model.py
```diff
@@ -0,0 +1,1221 @@
...
+    return [(i - (video_length - 1)) * video_stride for i in range(video_length)]
+
+
+def compute_action_delta_indices(chunk_size: int) -> list[int]:
+    """Return the action-chunk indices relative to the current timestep.
+
+    Returns:
+        ``list(range(chunk_size))``.
+    """
+    return list(range(chunk_size))
```
alfieroddan, line 42: Do these need to be there own functions?

library/src/physicalai/policies/rldx1/policy.py
```diff
@@ -0,0 +1,794 @@
...
+    All hyperparameters are explicit in the signature for discoverability.
+    Supports the dual-path initialization shared across Studio policies:
+
+    - **Lazy path**: ``Rldx1()`` + ``trainer.fit()`` - model built in ``setup()``
+      once dataset features are known.
+    - **Eager path**: ``Rldx1.load_from_checkpoint()`` or ``Rldx1(env_action_dim=...)``
+      - model built immediately.
+
+    Args:
+        chunk_size: Number of action predictions per forward pass (action horizon).
```
alfieroddan: not in constructor?

library/src/physicalai/policies/rldx1/policy.py
```diff
@@ -0,0 +1,794 @@
...
+                # Image augmentation (train only)
+                color_jitter_params=color_jitter_params,
+                image_min_area=image_min_area,
+                # Normalization
+                clip_outliers=clip_outliers,
+                action_horizon=n_action_steps,
+                embodiment_tag=embodiment_tag,
+            )
+
+        self.save_hyperparameters(ignore=["config", "pretrained_name_or_path", "compile_model"])
```
alfieroddan: Why not pretrained path? Is this so that when we "load_from_checkpoint" we don't load weights? Would that not skip loading certain config options?

library/src/physicalai/policies/rldx1/policy.py
```diff
@@ -0,0 +1,794 @@
...
+        Raises:
+            RuntimeError: If model or preprocessor has not been initialized.
+        """
+        if self.model is None or self._preprocessor is None:
+            msg = "Model not initialized. Call trainer.fit() or pass env_action_dim."
+            raise RuntimeError(msg)
+        preprocessed = self._preprocessor(batch)
+        return self.model.compute_val_loss(preprocessed)
+
+    def on_load_checkpoint(self, checkpoint: dict) -> None:  # noqa: PLR6301
```
alfieroddan: is this backwards compatibility needed - considering this is v1 in studio?

library/src/physicalai/policies/rldx1/policy.py
```diff
@@ -0,0 +1,794 @@
...
+            msg = "Model not initialized. Call trainer.fit() or pass env_action_dim."
+            raise RuntimeError(msg)
+
+        trainable_params = [p for p in self.model.parameters() if p.requires_grad]
+        optimizer = self._build_optimizer(trainable_params)
+
+        total_steps = 10000
+        if hasattr(self, "trainer") and self.trainer is not None:
+            total_steps = int(getattr(self.trainer, "estimated_stepping_batches", total_steps))
+        warmup_steps = max(1, int(total_steps * self.config.warmup_ratio))
```
alfieroddan: Should we assume that our Lightning Policy is always attached to a lightning trainer? To avoid silently creating wrong schedules?

.github/workflows/library.yml
```diff
@@ -185,7 +190,7 @@ jobs:
       - name: Install dependencies
         working-directory: library
         run: |
-          uv sync --frozen --extra tests --extra pi05 --extra libero --extra smolvla --extra cu128
+          uv sync --frozen --extra tests --extra pi05 --extra libero --extra smolvla --extra rldx1 --extra cu128
```
samet-akcay, line 194: maybe not related to this pr...

would it be an idea to have an optional dependencies section in `pyproject.toml` file, sth like `policies` that installs our 1st party policies. 

eg.,

```bash
uv pip install physicalai[policies]
```

or

```bash
uv sync --frozen --extra tests --extra policies --extra libero --extra cu128
```

application/backend/pyproject.toml
```diff
@@ -11,7 +11,7 @@ dependencies = [
   'harvesters; sys_platform != "darwin"',
   "fastapi[standard]<1",
   "physicalai[robots, capture, transport]",
-  "physicalai-train[pi0,smolvla]",
+  "physicalai-train[pi0,smolvla,rldx1]",
```
samet-akcay, line 16: and same as above, this would become `physicalai[policies]`?

library/docs/explanation/gyms/robocasa.md
```diff
@@ -55,9 +56,9 @@ next_obs, reward, terminated, truncated, info = gym.step(action)
 gym.close()
 
 # Task-group keyword (18 atomic tasks in robocasa v1.0)
-from physicalai.gyms import create_robocasa_gyms
+from physicalai.gyms import RoboCasaTaskGroup, create_robocasa_gyms
 
-gyms = create_robocasa_gyms(tasks="atomic_seen")
+gyms = create_robocasa_gyms(tasks=RoboCasaTaskGroup.ATOMIC_SEEN)
```
samet-akcay, line 61: both str, and enum not possible ? like `str | RoboCasaTaskGroup`

library/pyrefly-baseline.json
alfieroddan, line 1: -1700 lines is quite a lot? is this normal?

library/README.md
```diff
@@ -142,6 +142,25 @@ print(results.summary())
...
+
+# Run benchmark
+benchmark = RoboCasaBenchmark(task="atomic_seen", num_episodes=20)
+results = benchmark.evaluate(policy)
+
+# View results
+print(results.summary())
+results.to_json("results.json")
+```
+
```
alfieroddan: Unsure if needed here.

skills/library/physicalai-train-adding-a-policy/SKILL.md
alfieroddan, line 1: another pr with the other uv --no-sync?

skills/library/physicalai-train-benchmarking-a-policy/SKILL.md
alfieroddan, line 1: Again, another PR with --no-sync fixes for pre-commits etc ther'es more examples in the skills files - up to you.

library/docs/getting-started/first-benchmark.md
```diff
@@ -164,6 +164,56 @@ benchmark = PushTBenchmark(num_episodes=50)
...
+
+# Headless servers
+export MUJOCO_GL=egl
+```
+
+**CLI:**
+
+```bash
+physicalai benchmark \
+    --benchmark physicalai.benchmark.gyms.RoboCasaBenchmark \
```
samet-akcay: definitely not related to this pr, but would it not make more sense to have the following import path 
```diff
+ physicalai.benchmark.RoboCasaBenchmark
- physicalai.benchmark.gyms.RoboCasaBenchmark
```
@alfieroddan @sovrasov

alfieroddan (commented):
Thanks Eugene! Looks very good :smile:. Have done a first pass.

library/src/physicalai/benchmark/gyms/robocasa/__init__.py
```diff
@@ -0,0 +1,42 @@
...
+    - ``composite_unseen`` — composite tasks, target split
+    - ``pretrain50`` — 50-task pretrain split
+    - ``pretrain100`` — 100-task pretrain split
+    - ``pretrain200`` — 200-task pretrain split
+    - ``pretrain300`` — 300-task pretrain split
+
+Example:
+    Run a full atomic_seen benchmark and print per-task results:
+
+        >>> from physicalai.benchmark.gyms.robocasa import RoboCasaBenchmark
```
samet-akcay, line 21: same as above, and a bit nitpick. until we agree, maybe we could do relative imports like?

```python
from physicalai.benchmark import RoboCasaBenchmark
```

library/src/physicalai/policies/groot/components/__init__.py
```diff
@@ -26,20 +26,20 @@
...
-    swish,
-)
 from physicalai.policies.groot.components.transformer import (
     AdaLayerNorm,
     BasicTransformerBlock,
     TimestepEncoder,
     get_dit_class,
     get_self_attention_transformer_class,
 )
+from physicalai.policies.shared.components.nn import (
```
samet-akcay: isn't `shared` redundant here? You could just have `physicalai.policies.components`? 

and that would be inline with `physicalai.policies.rldx1.components`?

library/src/physicalai/policies/rldx1/components/action_model/blocks.py
```diff
@@ -0,0 +1,1284 @@
...
+
+    shift: torch.Tensor
+    scale: torch.Tensor
+    gate: torch.Tensor
+
+
+class Modulation(nn.Module):
+    """Flux-style modulation for AdaLN."""
+
+    def __init__(self, dim: int, double: bool, remove_bias: bool = False) -> None:  # noqa: FBT001, FBT002
```
samet-akcay: this would get rid of the boolean type hint issues ? 
```suggestion
    def __init__(self, dim: int, *, double: bool, remove_bias: bool = False) -> None:
```

library/src/physicalai/policies/rldx1/components/action_model/msat.py
```diff
@@ -0,0 +1,1247 @@
+# Copyright (C) 2026 Intel Corporation
+# SPDX-License-Identifier: Apache-2.0
+# Vendored from RLWRLD/RLDX-1 (Apache-2.0)
+# ruff: noqa: T201
```
samet-akcay: perhaps we could use a logger instead of print ?

library/src/physicalai/policies/rldx1/components/action_model/physics_head.py
```diff
@@ -0,0 +1,499 @@
+# Copyright (C) 2026 Intel Corporation
+# SPDX-License-Identifier: Apache-2.0
+# Vendored from RLWRLD/RLDX-1 (Apache-2.0)
+# ruff: noqa: T201
```
samet-akcay: same as above, maybe your agent can replace prints with a logger :)

library/src/physicalai/policies/rldx1/augmentations.py
```diff
@@ -0,0 +1,408 @@
...
+
+import cv2
+import numpy as np
+import torch
+
+from physicalai.policies.rldx1.preprocessing import compute_aspect_area_resize_crop
+
+if TYPE_CHECKING:
+    # Only for static type checking; runtime import is lazy via _import_albumentations().
+    import albumentations as A  # noqa: N812
```
samet-akcay: albumentations is now a public archieve. are we certain that this pipeline will always work?

samet-akcay: https://github.com/albumentations-team/albumentations

library/src/physicalai/policies/rldx1/preprocessor.py
samet-akcay: this is something we will need to discuss as a team

library/src/physicalai/policies/shared/components/__init__.py
```diff
@@ -0,0 +1,22 @@
+# Copyright (C) 2026 Intel Corporation
+# SPDX-License-Identifier: Apache-2.0
+
+"""Shared neural network components for policy modules."""
+
+from physicalai.policies.shared.components.nn import (
```
samet-akcay: Thoughts?
```suggestion
from physicalai.policies.components.nn import (
```

library/tests/unit/policies/rldx1_vendored/__init__.py
samet-akcay, line 1: any reason why unit tests for this policy are named as `library/tests/unit/policies/rldx1_vendored/`

why not library/tests/unit/policies/rldx1`?

samet-akcay (commented):
thanks @eugene123tw, massive effort. added some feedback. for some others I see that @alfieroddan already has given them, so I didn't add mine

library/docs/getting-started/first-benchmark.md
eugene123tw: Changed to Rldx1 with `policy = Rldx1(pretrained_name_or_path="RLWRLD/RLDX-1-FT-ROBOCASA")`

application/backend/pyproject.toml
eugene123tw: yeah, I agree. I think that would be less error prone. Otherwise, we have to update this every time adding a new policy.

library/pyproject.toml
eugene123tw: `peft` and `albumentations` are only declared under the rldx1 optional extra in pyproject.toml (pip install physicalai-train[rldx1]), not the base dependency set. Pyrefly type-checks against the main .venv, which doesn't include that extra by default, so it can't resolve those imports and reports spurious reportMissingImports errors, even though both are lazy, guarded imports at runtime (function-local try/except ImportError with a clear install message), so nothing actually breaks if the packages are absent.

Same pattern already exists for pi_gemma.py (Pi0.5) and groot/components/backbone.py, excluded for identical reasons before this PR.

library/src/physicalai/policies/rldx1/policy.py
```diff
@@ -0,0 +1,799 @@
...
+            truncate task-critical extremes (e.g. PushT).
+        env_action_dim: Environment action dimension. If provided, enables eager init.
+        dataset_stats: Dataset normalization statistics for eager init.
+    """
+
+    def __init__(  # noqa: PLR0913
+        self,
+        # Model architecture
+        n_action_steps: int = 16,
+        max_state_dim: int = 64,
```
sovrasov, line 141: Are max_dim parameter that frequently modified so we place them on top?
In most cases, changing them can break compatibility with initial pre-trained weights. Besides, 64 is enough with a good reserve for the future high-dim states

library/src/physicalai/policies/rldx1/policy.py
eugene123tw: Aligned with pi05's convention, dropped the guard and hardcoded fallback:
```
total_steps = int(self.trainer.estimated_stepping_batches)
warmup_steps = max(1, int(total_steps * self.config.warmup_ratio))
```

Now fails loudly with AttributeError if `configure_optimizers()` is ever called without a trainer attached

library/src/physicalai/eval/rollout/functional.py
eugene123tw: Good point. Changed the refresh to only trigger when `video_recorder.caption` is falsy, so an explicitly-set custom caption is preserved. Benchmarks (e.g. RoboCasa) still get their task description filled in post-reset since their initial caption is empty.

library/src/physicalai/policies/rldx1/components/action_model/attention.py
eugene123tw: Applied — made everything after attention_head_dim keyword-only and dropped the FBT001/FBT002/PLR0917 noqa suppressions

library/src/physicalai/policies/rldx1/components/action_model/blocks.py
eugene123tw: Applied — double and remove_bias are now keyword-only, noqa removed.

library/src/physicalai/policies/rldx1/components/backbone/modeling_vtc.py
eugene123tw: Let's stay with the American spelling :))

library/src/physicalai/policies/rldx1/components/action_model/msat.py
eugene123tw: Switched to logger (logging.getLogger(__name__)) and removed the # ruff: noqa: T201 file-level suppression.

library/src/physicalai/policies/rldx1/components/action_model/physics_head.py
eugene123tw: Switched to logger (logging.getLogger(__name__)) and removed the # ruff: noqa: T201 file-level suppression.

library/src/physicalai/policies/rldx1/components/action_model/blocks.py
eugene123tw: Applied as suggested. All call sites already used keyword args for double/remove_bias, so no other changes needed.

library/README.md
eugene123tw: Removed

library/src/physicalai/policies/rldx1/policy.py
sovrasov: Looks like we don't need to look at pretrained_path after we have all other hyper parameters set from the latest training: they describe the full model we'd like to build

library/src/physicalai/policies/rldx1/policy.py
eugene123tw: ah, yeah, sorry, remove the leftover.

library/src/physicalai/policies/rldx1/policy.py
eugene123tw: good eye 👁️  removed.

library/src/physicalai/policies/rldx1/augmentations.py
sovrasov: We can potentially migrate to https://github.com/albumentations-team/AlbumentationsX
but agpl-3.0 is not good news for us

library/src/physicalai/policies/rldx1/model.py
eugene123tw: Pulling them out lets us test the offset/index math in isolation with plain ints, without constructing a full `Rldx1Model` (backbone, safetensors, etc.). Inlining them into the `action_delta_indices`/`observation_delta_indices` properties would make this untestable without a full model instance.

library/src/physicalai/policies/rldx1/policy.py
eugene123tw: I removed it from ignore.

library/pyrefly-baseline.json
eugene123tw: It's pyrefly's baseline of pre-existing type errors, I only update the pyrefly's basefile with `uv run pyrefly check -c pyproject.toml --baseline="pyrefly-baseline.json" --update-baseline` otherwise the CI would fail. This is stated in the README.

library/tests/unit/policies/rldx1_vendored/__init__.py
eugene123tw: This is intentional. This directory only holds **parity tests** from the vendored upstream utility code (data/state-action processing copied from RLWRLD/RLDX-1), not the native Studio-integrated policy. There's no `rldx1/` test dir yet since native policy tests haven't been added; happy to rename if that distinction isn't useful though.

library/src/physicalai/policies/shared/components/__init__.py
eugene123tw: Fixed

library/src/physicalai/policies/groot/components/__init__.py
eugene123tw: good suggestion. Changed to `physicalai.policies.components`

library/docs/getting-started/first-benchmark.md
alfieroddan: I think possibly @sovrasov moved them to `benchmark.gyms` to avoid a naming collision with the runtime package - https://github.com/openvinotoolkit/physicalai/tree/main/src/physicalai/benchmark/performance

library/src/physicalai/policies/rldx1/policy.py
eugene123tw: Good question. I default to keeping `max_state_dim`/`max_action_dim` configurable (rather than hardcoding 64) because I found training converges better when they're set to the dataset's actual dimensions instead of always padding to 64.

I compared fine-tuning on PushT with max_state_dim=max_action_dim=2 (real dims) vs. =64 (padded), and the fixed-at-2 run converged to noticeably better accuracy.

My working theory is [CategorySpecificLinear](library/src/physicalai/policies/components/nn.py):
```py
selected_w = self.W[cat_ids]
selected_b = self.b[cat_ids]
return torch.bmm(x, selected_w) + selected_b.unsqueeze(1)
```

When state/action is zero-padded from 2→64, the extra 62 input columns are always exactly zero, so the corresponding 62 rows of `W` never receive a nonzero input and effectively never get gradient signal, they just sit at their random init the whole run, diluting the parameter's signal to noise ratio compared to a `W` sized exactly to the real input (dim=2 for instance).

So there are two options: always derive `max_state_dim`/`max_action_dim` from the dataset (no padding at all), or keep the vendored upstream default of always padding to 64. Vendored RLDX-1 always pads to 64, so I made it configurable rather than changing the default, I still want to run the same comparison on SO101 before committing either way in phase 2. If fixed-to-dataset-dim holds up there too, I'm happy to drop the configurability and always derive it from the dataset instead of exposing it as a parameter.

library/src/physicalai/policies/rldx1/augmentations.py
eugene123tw: Removed albumentations augmentations

skills/library/physicalai-train-benchmarking-a-policy/SKILL.md
eugene123tw: Fair point, but I think it's fine to include here, the change is minimal, and it's confined to SKILL.md docs, not application code, so there's no risk of it "contaminating" the codebase or interacting with anything this PR touches functionally. Happy to split it out if you'd still prefer a separate PR, but given the small blast radius I don't think it needs one.

.github/workflows/library.yml
eugene123tw: I agree, let's improve this in another PR.

library/docs/getting-started/first-benchmark.md
eugene123tw: Let's address this in another PR.

library/src/physicalai/benchmark/gyms/robocasa/__init__.py
eugene123tw: yes, but I think there's still some import path issue? Let's check this in another PR.

library/docs/explanation/gyms/robocasa.md
eugene123tw: The signature is tasks: RoboCasaTaskGroup | Sequence[str], not str | RoboCasaTaskGroup. That said, both `RoboCasaTaskGroup | Sequence[str]` is possible

.github/workflows/library.yml
samet-akcay: https://github.com/open-edge-platform/physical-ai-studio/issues/990

samet-akcay (dismissed):
thanks for the effort

skills/library/physicalai-train-adding-a-policy/SKILL.md
AlexanderBarabanov: Could you please update `https://github.com/openvinotoolkit/physicalai/blob/main/docs/how-to/config/use-from-config.md` -- currently this is 404.

AlexanderBarabanov (dismissed):

skills/library/physicalai-train-adding-a-policy/SKILL.md
eugene123tw: Done

AlexanderBarabanov (approved):

PR #769, opened 2026-07-03
fix: clear settings cache before serving studio

MarkRedeman:
This PR is a follow up of #764 that makes it so that our build script includes the URDF files of SO101 and Widowx AI arms in our python distribution.

It also includes an important fix (last commit) that makes it so that we clear our settings cache before serving studio. This is important as we overwrite `STATIC_FILES_DIR`, `ALEMBIC_CONFIG_PATH` and `ALEMBIC_SCRIPT_LOCATION` to target files relative to the packaged application.
Without these we'd get an error when serving the application outside of the `application/backend` directory.
This broke when `settings = get_settings()` was moved to the outer scope of `cli/serve.py`.

### DIFF
--- a/application/docker/Dockerfile
+++ b/application/docker/Dockerfile
@@ -176,6 +176,20 @@ RUN uv pip uninstall --python .venv/bin/python \
     expecttest hypothesis sortedcontainers kgb parameterized
 
 # ===========================================================================
+# Stage 4d: Robot assets builder
+#
+# Sync backend-owned robot assets once so all runtime targets can copy the
+# exact same files before the backend wheel install step.
+# ===========================================================================
+FROM builder-base AS robot-assets
+
+COPY --link --from=app-source /app/application/backend/src /app/application/backend/src
+
+RUN PYTHONPATH=/app/application/backend/src \
+    /app/application/backend/.venv/bin/python -c \
+    "from robots.catalog.sync_robot_assets import sync_robot_assets; sync_robot_assets()"
+
+# ===========================================================================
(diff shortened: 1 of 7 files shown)


application/docker/Dockerfile
```diff
@@ -252,7 +252,7 @@ ARG APP_GID=1000
 COPY --link --from=builder-cpu --chown=${APP_UID}:${APP_GID} \
     /app/application/backend/.venv /app/application/backend/.venv
 COPY --link --from=app-source --chown=${APP_UID}:${APP_GID} /app /app
-RUN uv pip install --python /app/application/backend/.venv/bin/python --no-deps /app/application/backend
+RUN uv pip install --python /app/application/backend/.venv/bin/python --no-deps -e /app/application/backend
```
AlbertvanHouten: Why editable install? This normally shouldn't be used in production images

MarkRedeman: When I didn't do this the install would trigger the `hatch_build.py` script, but because it wouldn't have been able to run `uv run physicalai-studio sync-robot-assets` yet this will fail.

I don't mind reverting this and getting a better proper fix for it.

MarkRedeman: I've updated the code:
1. ~~Removed `src/static/robot-assets` from `build_data["force_include"]`, since these will be included as they are available from `src`~~ (actually was needed because hatch will ignore files from `.gitignore`?)
2. Added a new build stage where we run the `sync-robot-assets` command directly. We then clone the assets from this stage.

AlbertvanHouten (approved):

PR #770, opened 2026-07-03
fix: do allow setup of SO101 if there's no serial number

MarkRedeman:
We were still doing a check on the serial number during the SO101 setup. I don't have a SO101 without serial number so forgot to verify this.

### DIFF
--- a/application/ui/src/routes/robots/new.tsx
+++ b/application/ui/src/routes/robots/new.tsx
@@ -1,14 +1,10 @@
 import { Suspense } from 'react';
 
-import { Button, Flex, Grid, Loading, minmax, View } from '@geti-ui/ui';
-import { useNavigate } from 'react-router';
+import { Flex, Grid, Loading, minmax, View } from '@geti-ui/ui';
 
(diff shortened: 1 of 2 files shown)


application/ui/src/routes/robots/new.tsx
```diff
@@ -49,6 +38,23 @@ const NewRobotSubmitButton = () => {
...
+ * - SO101 types: navigates to the setup wizard (sibling route under same layout)
+ * - Other types (Trossen): directly creates the robot via POST (default behavior)
+ */
+const NewRobotSubmitButton = () => {
+    const { activeType } = useRobotForm();
+
+    const isSO101 = activeType === 'SO101_Follower' || activeType === 'SO101_Leader';
+
+    if (isSO101) {
+        return <BeginSO101SetupButton activeType={activeType} />;
```
MarkRedeman: Note: we add `activeType` here to help with type inference. By passing `activeType` (whose type is restricted to be one of the so101 follower types) we can infer that `useRobotForm(activeType)` will return a SO101 payload.

maxxgx (approved):

PR #771, opened 2026-07-03
chore(ci): add `skill-scan`

AlexanderBarabanov:
This PR adds a new “SkillSpector” scan to the Skills CI workflow and refactors workflow triggering to be path-aware via an initial `check-paths` gate.


ivanzati (approved):
Thank you, Alexander!

PR #774, opened 2026-07-06
feat: allow editing robots

### LINKED ISSUE #761: Allow users to edit robots
At the moment we only allow users to add or delete robots, without an option to edit them.
However this becomes problematic when we merge #760 as the user will not have an option to update the usb port used by their robot.

To fix this we should add a new edit page where the user can change the usb port, and perhaps other fields as well.
We may also reuse this same page to export and import calibration files.

MarkRedeman:
This is important now that we allow adding robots without a unique serial number.
For such cases we only store the usb port that was selected when adding the robot. But since this port can change when reconnecting the robot or when rebooting a system the user could end up with their robot living on a different port, and thus showing up as disconnected.

Note: we already had some basic support for editing robots but it was disabled because it was allowing users to change the type of a robot, which can then lead to inconsistent data.
This PR changes the form logic so that instead of having a single `RobotForm` we now have a `CreateRobotForm` and `UpdateRobotForm` where each is responsible for rendering its own form fields.
A nice advantage of this approach is that we could get rid of,

```
    const handleSubmit = (event: FormEvent<HTMLFormElement>) => {
        event.preventDefault();
        submitContainerRef.current?.querySelector('button')?.click();
    };

```
and instead of an actual submit handler that is passed to the form (and thus buttons etc)

image


application/ui/src/features/robots/robot-form/update-form.tsx
```diff
@@ -0,0 +1,66 @@
+import type { FormEvent } from 'react';
```
camiloHimura, line 1: Will you add unit tests later?

camiloHimura (approved):

application/ui/src/features/robots/robot-form/update-form.tsx
MarkRedeman: Will add some in a follow up pr. I expect we need to make a few more behavioural changes after finishing the  robot plugins (#775 and #776 are a start, will need a few more prs after that as well)

For proper integration and end to end tests we also need to make some improvements to our architecture.

PR #775, opened 2026-07-06
feat(robot-catalog): Add robot catalog definitions API

MarkRedeman:
Add a new `list_robot_catalog` endpoint that returns definitions of all robots supported by PhsycialAI Studio.
These definitions will be used by the UI to feature names to urdf joint names, determine the URDF path (and package map) to show the 3D model and simply list all supported models when adding a new robot.

At the moment the `RobotCatalogRegistry` is hardcoded to only include the current SO101 and WidowX AI robots. The registry will be extended to allow registring new definitions from external packages.

We will also move the build methods from `RobotClientFactory` into these `RobotCatalogDefinition` and give the factory class access to the registry.

Example response from the new endpoint:
```json
[
    {
        "type": "SO101_Follower",
        "display_name": "SO101 Follower",
        "role": "follower",
        "urdf_path": "/api/robots/catalog/SO101_Follower/urdf",
        "package_map": {
            "SO101": "/api/robots/catalog/SO101_Follower"
        },
        "joint_map": {
            "shoulder_pan.pos": ["shoulder_pan"],
            "shoulder_lift.pos": ["shoulder_lift"],
            "elbow_flex.pos": ["elbow_flex"],
            "wrist_flex.pos": ["wrist_flex"],
            "wrist_roll.pos": ["wrist_roll"],
            "gripper.pos": ["gripper"]
        }
    },
    {
        "type": "SO101_Leader",
        "display_name": "SO101 Leader",
        "role": "leader",
        "urdf_path": "/api/robots/catalog/SO101_Leader/urdf",
        "package_map": {
            "SO101": "/api/robots/catalog/SO101_Leader"
        },
        "joint_map": {
            "shoulder_pan.pos": ["shoulder_pan"],
            "shoulder_lift.pos": ["shoulder_lift"],
            "elbow_flex.pos": ["elbow_flex"],
            "wrist_flex.pos": ["wrist_flex"],
            "wrist_roll.pos": ["wrist_roll"],
            "gripper.pos": ["gripper"]
        }
    },
    {
        "type": "Trossen_WidowXAI_Follower",
        "display_name": "Trossen WidowX AI Follower",
        "role": "follower",
        "urdf_path": "/api/robots/catalog/Trossen_WidowXAI_Follower/urdf",
        "package_map": {
            "trossen_arm_description": "/api/robots/catalog/Trossen_WidowXAI_Follower"
        },
        "joint_map": {
            "shoulder_pan.pos": ["joint_0"],
            "shoulder_lift.pos": ["joint_1"],
            "elbow_flex.pos": ["joint_2"],
            "wrist_flex.pos": ["joint_3"],
            "wrist_yaw.pos": ["joint_4"],
            "wrist_roll.pos": ["joint_5"],
            "gripper.pos": ["left_carriage_joint", "right_carriage_joint"]
        }
    },
    {
        "type": "Trossen_WidowXAI_Leader",
        "display_name": "Trossen WidowX AI Leader",
        "role": "leader",
        "urdf_path": "/api/robots/catalog/Trossen_WidowXAI_Leader/urdf",
        "package_map": {
            "trossen_arm_description": "/api/robots/catalog/Trossen_WidowXAI_Leader"
        },
        "joint_map": {
            "shoulder_pan.pos": ["joint_0"],
            "shoulder_lift.pos": ["joint_1"],
            "elbow_flex.pos": ["joint_2"],
            "wrist_flex.pos": ["joint_3"],
            "wrist_yaw.pos": ["joint_4"],
            "wrist_roll.pos": ["joint_5"],
            "gripper.pos": ["left_carriage_joint", "right_carriage_joint"]
        }
    },
    {
        "type": "Trossen_Bimanual_WidowXAI_Follower",
        "display_name": "Trossen Bimanual WidowX AI Follower",
        "role": "follower",
        "urdf_path": "/api/robots/catalog/Trossen_Bimanual_WidowXAI_Follower/urdf",
        "package_map": {
            "trossen_arm_description": "/api/robots/catalog/Trossen_Bimanual_WidowXAI_Follower"
        },
        "joint_map": {
            "left_shoulder_pan.pos": ["follower_left_joint_0"],
            "left_shoulder_lift.pos": ["follower_left_joint_1"],
            "left_elbow_flex.pos": ["follower_left_joint_2"],
            "left_wrist_flex.pos": ["follower_left_joint_3"],
            "left_wrist_yaw.pos": ["follower_left_joint_4"],
            "left_wrist_roll.pos": ["follower_left_joint_5"],
            "left_gripper.pos": [
                "follower_left_left_carriage_joint",
                "follower_left_right_carriage_joint"
            ],
            "right_shoulder_pan.pos": ["follower_right_joint_0"],
            "right_shoulder_lift.pos": ["follower_right_joint_1"],
            "right_elbow_flex.pos": ["follower_right_joint_2"],
            "right_wrist_flex.pos": ["follower_right_joint_3"],
            "right_wrist_yaw.pos": ["follower_right_joint_4"],
            "right_wrist_roll.pos": ["follower_right_joint_5"],
            "right_gripper.pos": [
                "follower_right_left_carriage_joint",
                "follower_right_right_carriage_joint"
            ]
        }
    },
    {
        "type": "Trossen_Bimanual_WidowXAI_Leader",
        "display_name": "Trossen Bimanual WidowX AI Leader",
        "role": "leader",
        "urdf_path": "/api/robots/catalog/Trossen_Bimanual_WidowXAI_Leader/urdf",
        "package_map": {
            "trossen_arm_description": "/api/robots/catalog/Trossen_Bimanual_WidowXAI_Leader"
        },
        "joint_map": {
            "left_shoulder_pan.pos": ["follower_left_joint_0"],
            "left_shoulder_lift.pos": ["follower_left_joint_1"],
            "left_elbow_flex.pos": ["follower_left_joint_2"],
            "left_wrist_flex.pos": ["follower_left_joint_3"],
            "left_wrist_yaw.pos": ["follower_left_joint_4"],
            "left_wrist_roll.pos": ["follower_left_joint_5"],
            "left_gripper.pos": [
                "follower_left_left_carriage_joint",
                "follower_left_right_carriage_joint"
            ],
            "right_shoulder_pan.pos": ["follower_right_joint_0"],
            "right_shoulder_lift.pos": ["follower_right_joint_1"],
            "right_elbow_flex.pos": ["follower_right_joint_2"],
            "right_wrist_flex.pos": ["follower_right_joint_3"],
            "right_wrist_yaw.pos": ["follower_right_joint_4"],
            "right_wrist_roll.pos": ["follower_right_joint_5"],
            "right_gripper.pos": [
                "follower_right_left_carriage_joint",
                "follower_right_right_carriage_joint"
            ]
        }
    }
]
```

### DIFF
--- a/application/backend/src/api/robot_catalog.py
+++ b/application/backend/src/api/robot_catalog.py
@@ -1,24 +1,64 @@
+from pathlib import Path
+from typing import Literal
+
 from fastapi import APIRouter
 from fastapi.responses import FileResponse
+from pydantic import BaseModel, Field
 
+from api.dependencies import RobotCatalogServiceDep
 from robots.catalog.assets import resolve_robot_asset_path, resolve_robot_urdf_path
+from robots.catalog.types import RobotCatalogDefinition
 from schemas.robot import RobotType
+
+
+class RobotCatalogDefinitionResponse(BaseModel):
+    type: RobotType = Field(..., description="Stable backend robot type identifier")
+    display_name: str = Field(..., description="Human-readable robot type label")
+    role: Literal["follower", "leader"] = Field(..., description="Default robot role")
+    urdf_path: str = Field(description="URDF URL used by the UI model loader")
+    package_map: dict[str, str] = Field(default_factory=dict, description="URDF package name to URL prefix map")
+    joint_map: dict[str, list[str]] = Field(
+        description="Observation joint name to URDF joint(s) mapping",
+    )
+
+
+def _to_response(definition: RobotCatalogDefinition) -> RobotCatalogDefinitionResponse:
+    return RobotCatalogDefinitionResponse(
+        type=definition.type,
+        display_name=definition.display_name,
+        role=definition.role,
+        urdf_path=definition.urdf_path,
+        package_map=definition.package_map,
+        joint_map=definition.joint_map,
+    )
+
 
 router = APIRouter(prefix="/api/robots/catalog", tags=["Robot Catalog"])
 
 
+@router.get("")
+async def list_robot_catalog(catalog_service: RobotCatalogServiceDep) -> list[RobotCatalogDefinitionResponse]:
+    """List robot catalog definitions exposed to the UI."""
+    return [_to_response(definition) for definition in catalog_service.list_entries()]
+
+
 @router.get("/{robot_type}/urdf")
-async def get_robot_catalog_urdf(robot_type: RobotType) -> FileResponse:
+async def get_robot_catalog_urdf(catalog_service: RobotCatalogServiceDep, robot_type: RobotType) -> FileResponse:
     """Return the URDF file for a catalog robot type."""
-    resolved_path = resolve_robot_urdf_path(robot_type=robot_type)
+    definition = catalog_service.get_definition(robot_type)
+
+    resolved_path = resolve_robot_urdf_path(definition)
     return FileResponse(resolved_path)
 
 
 @router.get("/{robot_type}/{asset_path:path}")
 async def get_robot_catalog_asset(
+    catalog_service: RobotCatalogServiceDep,
     robot_type: RobotType,
-    asset_path: str,
+    asset_path: Path,
 ) -> FileResponse:
     """Return an asset file referenced by a catalog robot URDF."""
-    resolved_path = resolve_robot_asset_path(robot_type=robot_type, asset_path=asset_path)
+    definition = catalog_service.get_definition(robot_type)
+
+    resolved_path = resolve_robot_asset_path(definition, asset_path=asset_path)
     return FileResponse(resolved_path)
--- a/application/backend/src/robots/catalog/registry.py
+++ b/application/backend/src/robots/catalog/registry.py
@@ -1,0 +1,24 @@
+from schemas.robot import RobotType
+
+from . import so101, widowxai
+from .types import RobotCatalogDefinition
+
+
+class RobotCatalogRegistry:
+    def __init__(self) -> None:
+        self._definitions: dict[RobotType, RobotCatalogDefinition] = {}
+
+        for definition in so101.get_definitions() + widowxai.get_definitions():
+            self.register(definition)
+
+    def list_definitions(self) -> list[RobotCatalogDefinition]:
+        return list(self._definitions.values())
+
+    def get_definition(self, robot_type: RobotType) -> RobotCatalogDefinition | None:
+        return self._definitions.get(robot_type)
+
+    def register(self, definition: RobotCatalogDefinition) -> None:
+        if definition.robot_type in self._definitions:
+            raise ValueError(f"Duplicate robot catalog registration for type: {definition.robot_type}")
+
+        self._definitions[definition.robot_type] = definition
--- a/application/backend/src/robots/catalog/types.py
+++ b/application/backend/src/robots/catalog/types.py
@@ -1,0 +1,24 @@
+from __future__ import annotations
+
+from typing import Literal
+
+from pydantic import BaseModel, Field
+
+from schemas.robot import RobotType
+
+
+class RobotCatalogDefinition(BaseModel):
+    type: RobotType = Field(..., description="Stable backend robot type identifier")
+    display_name: str = Field(..., description="Human-readable robot type label")
+    role: Literal["follower", "leader"] = Field(..., description="Default robot role")
+    urdf_path: str = Field(description="URDF URL used by the UI model loader")
+    package_map: dict[str, str] = Field(default_factory=dict, description="URDF package name to URL prefix map")
+    joint_map: dict[str, list[str]] = Field(
+        default_factory=dict,
+        description="Observation joint name to URDF joint(s) mapping",
+    )
+    urdf_relative_path: str = Field(..., description="Relative path to the robot URDF asset")
+
+    @property
+    def robot_type(self) -> RobotType:
+        return self.type
--- a/application/backend/src/services/robot_catalog_service.py
+++ b/application/backend/src/services/robot_catalog_service.py
@@ -1,0 +1,22 @@
+from exceptions import ResourceNotFoundError, ResourceType
+from robots.catalog.registry import RobotCatalogRegistry
+from robots.catalog.types import RobotCatalogDefinition
+from schemas.robot import RobotType
+
+
+class RobotCatalogService:
+    def __init__(self) -> None:
+        self._registry: RobotCatalogRegistry = RobotCatalogRegistry()
+
+    def list_entries(self) -> list[RobotCatalogDefinition]:
+        return self._registry.list_definitions()
+
+    def get_definition(self, robot_type: RobotType) -> RobotCatalogDefinition:
+        definition = self._registry.get_definition(robot_type)
+        if definition is None:
+            raise ResourceNotFoundError(
+                resource_type=ResourceType.ROBOT,
+                resource_id=robot_type,
+                message="Robot type is not part of the catalog.",
+            )
+        return definition
--- a/application/backend/src/api/dependencies.py
+++ b/application/backend/src/api/dependencies.py
@@ -24,6 +24,7 @@ from services.event_processor import EventProcessor
 from services.job_service import JobService
 from services.log_service import LogService
 from services.robot_calibration_service import RobotCalibrationService
+from services.robot_catalog_service import RobotCatalogService
 from services.system_service import SystemService
 from settings import get_settings
 from utils.serial_robot_tools import RobotConnectionManager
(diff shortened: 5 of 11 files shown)


MarkRedeman:
Note: you can find an initial example of how to integrate an external plugin here: https://github.com/MarkRedeman/physicalai-rebot-b601-plugin/blob/main/src/physicalai_rebot_b601_plugin/studio_catalog.py

However that integration is based on a more complex API which I've tried to simplify in this PR.
There are a few follow up PRs that I will start after this gets merged:
1. UI integration that uses the new catalog API
2. Move build definition from `RobotClientFactory` into each of the catalog definitions (`so101.py`, `widowx.py`)
3. Extend registry with ability to load external packages (by loading them via a module entrypoint)
4. Extend definition with capability to determine if a robot is online
5. Extend `RobotType` to allow any string so that we can support external plugins. Same for `Robot` (`RobotAdapter`), and `RobotWithConnectionState` (`RobotWithConnectionStateAdapter`)
6. Generate UI forms based on JSON schema. (less important, for now we can hardcode this in the UI)

application/backend/src/robots/catalog/types.py
```diff
@@ -0,0 +1,24 @@
+from __future__ import annotations
+
+from typing import Literal
+
+from pydantic import BaseModel, Field
+
+from schemas.robot import RobotType
+
+
+class RobotCatalogDefinition(BaseModel):
```
MarkRedeman, line 10: This is expected to be the general shape of how a robot catalog (plugin) definition will look like. Maybe renaming this to `RobotCatalogPlugin` might be nicer?

We will likely need to extend this with a few more fields when integrating an actual plugin, specifically: (see https://github.com/MarkRedeman/physicalai-rebot-b601-plugin/blob/main/src/physicalai_rebot_b601_plugin/studio_catalog.py#L244-L248)
- `asset_root_resolver`: which will return the relative path to the urdf of the robot, might replace `urdf_relative_path`.
- `discover_devices`: used to determine which robots are online and to know what robots to show in the robot picker when adding/editing a robot
- `robot_builder`: factory method that will be used by `ClientRobotFactory`
- `payload_model`: the `payload` when adding/updating a robot of this type. This payload will also be used by `robot_builder`.
- `adapter_options`: extra arguments for `robot_builder`, might not be needed

I figured though that it would be better to keep this first integration simple so we can get more feedback on this design.

application/backend/src/services/robot_catalog_service.py
```diff
@@ -0,0 +1,22 @@
+from exceptions import ResourceNotFoundError, ResourceType
+from robots.catalog.registry import RobotCatalogRegistry
+from robots.catalog.types import RobotCatalogDefinition
+from schemas.robot import RobotType
+
+
+class RobotCatalogService:
+    def __init__(self) -> None:
+        self._registry: RobotCatalogRegistry = RobotCatalogRegistry()
+
```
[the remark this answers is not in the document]
MarkRedeman: Will be added once needed.

application/backend/src/api/robot_catalog.py
```diff
@@ -1,24 +1,64 @@
...
+from robots.catalog.types import RobotCatalogDefinition
 from schemas.robot import RobotType
 
+
+class RobotCatalogDefinitionResponse(BaseModel):
+    type: RobotType = Field(..., description="Stable backend robot type identifier")
+    display_name: str = Field(..., description="Human-readable robot type label")
+    role: Literal["follower", "leader"] = Field(..., description="Default robot role")
+    urdf_path: str = Field(description="URDF URL used by the UI model loader")
+    package_map: dict[str, str] = Field(default_factory=dict, description="URDF package name to URL prefix map")
```
AlbertvanHouten, line 19: urdf_path and package_map seem to follow the same pattern. You might be able to define them as properties instead:

```
@property
def urdf_path(self):
    return "/api/robots/catalog/{self.type}/urdf"
```

AlbertvanHouten (approved):

application/backend/src/api/robot_catalog.py
AlbertvanHouten: why is package map a dict?

MarkRedeman: The tricky part about this is that for trossen we'd need to define the package_map as:
```
{
    "trossen_arm_description": "/api/robots/catalog/Trossen_WidowXAI_Follower"
}
```

This is necessary because internally the URDF we use for Trossen's robots use: `package://trossen_arm_description/meshes/wxai/base_link.stl`. So our URDF loader needs to know specifically for the `trossen_arm_description` how to load this.

We could probably omit the `urdf_path` from the API, as that would also be something that we can constrcut with `fetchClient.PATH(...)` (a typesafe way to generate paths that are used by the OpenAPI spec), but I haven't found a proper way to avoid this for the package map.

application/backend/src/robots/catalog/types.py
maxxgx: I feel like callables such as `asset_root_resolver, discover_devices, robot_builder` should not be part of the definition. 
- Is the plugin developer responsible for implementing them?
- Wouldn't you need to write custom functions for each robot plugin?

application/backend/src/robots/catalog/registry.py
```diff
@@ -0,0 +1,24 @@
...
+from . import so101, widowxai
+from .types import RobotCatalogDefinition
+
+
+class RobotCatalogRegistry:
+    def __init__(self) -> None:
+        self._definitions: dict[RobotType, RobotCatalogDefinition] = {}
+
+        for definition in so101.get_definitions() + widowxai.get_definitions():
+            self.register(definition)
```
maxxgx, line 12: What's the plan for dynamically finding the plugin definitions?

I was thinking that we could have a list of plugin definitions in a JSON file, and that would be our curated list of supported robots. Each plugin would define its python package and we could install them at runtime.

application/backend/src/robots/catalog/registry.py
maxxgx, line 1: I think `physicalai` framework must also support the same set of robot plugins. Does the current design address that?

application/backend/src/robots/catalog/types.py
MarkRedeman: Yes my intention is that the plugin author knows how to construct the robot for instance:
```
async def _build_rebot_arm102_driver(
    robot: _PayloadContainer[ReBotArm102Payload],
    factory: _PortFinder,
) -> PhysicalAIRobot:
    raw = robot.payload
    if isinstance(raw, ReBotArm102Payload):
        validated = raw
    elif isinstance(raw, dict):
        validated = ReBotArm102Payload.model_validate(raw)
    else:
        validated = ReBotArm102Payload.model_validate(raw.model_dump(mode="json"))

    serial_number = validated.serial_number
    port = await factory.find_port_by_serial(serial_number)
    if port is None:
        msg = f"Robot not found: {serial_number}"
        raise RuntimeError(msg)

    return ReBotArm102Leader(
        port=port,
        baudrate=validated.baudrate,
        unlock_on_connect=validated.unlock_on_connect,
        reset_multi_turn_on_connect=validated.reset_multi_turn_on_connect,
        zero_on_connect=validated.zero_on_connect,
    )
```

This would be the function that is implemented by the rebot plugin for `robot_builder`.

Similarly for the `asset_root_resolver`, so that the package can resolve to files internally to how it build its assets.

Otherwise we would need to update our own factory class everytime we add a new plugin / robot etc.

application/backend/src/robots/catalog/registry.py
MarkRedeman: In my initial desgin I'm using the package's entrypoints to do this. So for the rebot plugin it has,
```
[project.entry-points."physicalai.studio.catalog_plugins"]
rebot-b601 = "physicalai_rebot_b601_plugin.studio_catalog:register_physicalai_studio_plugin"
```
in its [pyrpoject.toml](https://github.com/MarkRedeman/physicalai-rebot-b601-plugin/blob/main/pyproject.toml#L32-L33).

and then in the `RobotCatalogRegistry` class I'm doing something along the lines of:
```python
CATALOG_PLUGIN_ENTRYPOINT_GROUP = "physicalai.studio.catalog_plugins"


    def _load_external_plugins(self) -> None:
        discovered_entry_points = list(entry_points(group=CATALOG_PLUGIN_ENTRYPOINT_GROUP))

        for discovered_entry_point in discovered_entry_points:
            register_plugin = discovered_entry_point.load()
            if not callable(register_plugin):
                raise ValueError(
                    f"Catalog plugin entry point '{discovered_entry_point.name}' must load a callable "
                    "register_physicalai_studio_plugin(registry)"
                )

            plugin_callable: RegisterPluginCallable = register_plugin
            plugin_callable(self)
```

So then if you simply add the package `uv add physicalai-rebot-b601-plugin` we will automatically detect this and add the robot to studio.

I think one important thing here to add is have an allowlist so that we can give warnings when users add plugins that might not be official.
The way I've thought about this is that we can later extend this as well with a UI that allows users to install plugins via the UI, similar to vscode extensions.

application/backend/src/robots/catalog/types.py
maxxgx: So this function answers the question: `how to build the robot for physicalai studio?`

Where do you define it, studio repo or plugin repo? I think if you define it in the plugin repo, you'd have to depend on physical-ai-studio. The dependency flow is not that clear to me

maxxgx (approved):
Approving after IRL discussion

PR #779, opened 2026-07-06
refactor: update hyperparameters for ACT

sovrasov:
# Pull Request

## Description

The update slightly improves convergence, and therefore requires less steps to get a working model


library/configs/physicalai/act.yaml
```diff
@@ -29,7 +31,7 @@ data:
                 - 1.3
 
 trainer:
-  max_epochs: 100
+  max_steps: 70000
```
MarkRedeman, line 34: In the application we are setting the `max_steps` from the UI side, using 10.000 by default.
Do we have any guestimates on what the correct values should be?

Right now we always use 10000 as a default (but do allow users to change this) regardless of which model is selected.
Should we update the UI so that we adjust the max steps based on the selected model, and perhaps dataset size?

sovrasov: Definitely, that setting should be per model. For ACT I'll make a couple more experiments and tell you the number (now I think of 30k with batch 16). lerobot's baseline is 100k with 8 batch size FYI

MarkRedeman: That sounds good, thanks!

We also have an `BatchSizeFinder` that's enabled by default in the UI. So in that case I guess we might need to add some extra logic that updates the max steps based on that? (or use epochs, but I was getting conflicting info on whether we should use epochs vs steps?)

sovrasov: This finder is cuda only as far as I remember, and we have only a tiny bit of information how all these heuristics affect the VLAs training. In Geti we could conduct hundreds of experiments on ~10 datasets per model to figure out an automatic strategy that pushes hardware limits. Here I'd rather go with fail-safe approach: if 50k with batch 8 is good enough, I'd use it, rather than implementing and testing extra schedule scaling logic (which we don't have now despite having batch size finder enabled)

samet-akcay (approved):
lgtm thanks a lot for the effort!

PR #780, opened 2026-07-07
feat: update to lerobot 0.6.0

MarkRedeman:
Update to latest lerobot 0.6.0.
Once merged we can enable more video hardware acceleration options via https://github.com/huggingface/lerobot/pull/3455


MarkRedeman:
Note: currently set to draft. I will do some more extensive end to end testing. I've also included a torch update from #727.

At the moment I've updated,`exclude-newer-package` to exclude lerobot as it's only been released today. Ideally we should revert that change, or perhaps change `exclude-newer` to be smaller?

AlbertvanHouten (approved):

sovrasov (approved):
Thanks Mark!
Confirm that a model trained from this branch works IRL, which means data pipeline stayed correct

PR #793, opened 2026-07-09
feat: add remote training backend with HTTP dataset transfer and resilient, resumable runs

AlbertvanHouten:
# Pull Request

## Description

This PR builds out end-to-end **remote training**: it adds the backend-side `RemoteTrainingBackend` that offloads a training job to the standalone trainer microservice, and makes those long-running remote runs robust and easy to operate. The default way to hand a dataset to the trainer is now a direct **HTTP ZIP upload** (`TRAINER_DATASET_TRANSFER=http`), so remote training no longer depends on an ephemeral private HuggingFace dataset repo.

On top of the transfer work, remote runs — which can take many hours — are now resilient to dropped connections and Studio restarts, so a long job is no longer lost to a transient network blip, a proxy killing an SSE stream, or the backend being restarted.

`TRAINING_MODE=local` (in-process torch/Lightning training) remains the default and is unaffected.

### Remote training backend
- New `RemoteTrainingBackend` (`services/training_backends/remote.py`), selected by `get_training_backend()` when `TRAINING_MODE=remote`. It submits the job to the trainer, transfers the dataset, mirrors progress, and downloads the resulting model artifact. `LocalTrainingBackend` is unchanged and still the default.

### HTTP dataset transfer
- Stream the dataset ZIP straight to the trainer (`PUT /jobs/{id}/dataset`, resumable via `Content-Range`) as the default transfer.
- Trainer-side upload handling with archive-safety validation and an `awaiting_dataset → queued` lifecycle.
- Archive-safety guards on both sides (`services/archive_safety.py` and trainer `archive_safety.py`): file-count / uncompressed-size (zip-bomb) limits, path-traversal and symlink checks, disk-headroom checks, and single-root flattening that ignores macOS junk files.
- Throttled byte-progress heartbeats (`_transfer_progress.py`) for both the snapshot upload and the model download, mirrored into the job progress bar.

### Resilient trainer event stream
- The trainer keeps running a job regardless of the client connection, so a dropped SSE stream is now recoverable: reconnect with exponential backoff (capped by `TRAINER_STREAM_RECONNECT_BACKOFF_MAX_S`) instead of a fixed short retry.
- Replaced the fragile fixed reconnect-count limit with a time budget: keep retrying while the trainer is unreachable for up to `TRAINER_STREAM_RECONNECT_MAX_S` (default 15 min) before giving up, so a multi-hour job survives short outages.
- Added a plain `GET /jobs/{id}` poll fallback for environments where a middlebox/proxy breaks long-lived SSE streaming but still serves short requests.

### Reattach to in-flight remote jobs after a Studio restart
- Persist the trainer's `remote_job_id` on the job payload so a restart can reconnect to the still-running (or completed) remote job instead of failing it.
- Restart reconciliation is now reattach-aware: a RUNNING remote job with a recorded `remote_job_id` is requeued and reattached (progress mirrored, model ingested) rather than marked FAILED; anything not resumable is still failed.
- Distinguish application shutdown from user cancellation: on shutdown the remote job is left running for reattach (new `TrainingSuspendedError`); an explicit user cancel still cancels the remote job.
- On reattach the dataset snapshot is not re-created or re-uploaded — the run resumes straight into progress mirroring and model download.
- The HTTP transfer path persists `remote_job_id` after the upload lands, so the default transfer can actually reattach instead of re-uploading the whole dataset on restart.

### Device discovery & remote availability (backend + UI)
- `GET /api/system/devices/training` now returns a `TrainingDevices` payload (`mode`, `remote_available`, `devices`). In remote mode the devices reflect the trainer's hardware; when the trainer is unreachable, `remote_available` is false and no devices are returned so callers can block training instead of silently falling back to local CPU.
- Train dialog polls availability while the remote trainer is unavailable, shows a "Remote trainer unavailable" status and warning, disables the **Train** button, and re-checks availability just before submitting.
- Job status/table now show the job's live `message` (e.g. upload/training progress) instead of a static label.

### CLI
- New `physicalai-studio remote` launcher (`cli/run.py`): loads the matching `.env`, runs `uv sync` for the requested device (`cpu`/`cuda`/`xpu`), warns on world/group-readable env files that may hold `HF_TOKEN`, and starts the backend in remote mode via a shared `start_server()` helper reused by `serve`.
- New `physicalai-trainer` CLI (trainer `cli.py`) to launch the trainer microservice.

## Tests
- Remote-backend tests (`test_remote_training_backend.py`): reconnect/poll fallback, unreachable-budget abort, reattach without re-submitting, shutdown-suspends-without-cancel, and an HTTP `remote_job_id` persistence regression guard.
- Worker tests (`test_training_worker.py`): suspend → requeue, reattach wiring, id persistence; and training-service reattach reconciliation tests (`test_training_service_reattach.py`).
- Device-availability tests (`test_system_service.py`), backend archive-safety tests, remote CLI tests (`test_run.py`), and training-log-format tests.
- Trainer-side tests for resumable upload (`test_api_upload.py`), archive safety, schemas, store, and CLI.

## Configuration
- `TRAINER_DATASET_TRANSFER` (`http` default, `hf` optional)
- `TRAINER_STREAM_RECONNECT_MAX_S` (default 900s), `TRAINER_STREAM_RECONNECT_BACKOFF_MAX_S` (default 30s)
- `.env.example` files added for backend and trainer; `.dockerignore`, Dockerfile, `run.sh`, and README updates for remote deployment.


application/backend/src/services/training_backends/remote.py
```diff
@@ -145,7 +149,39 @@ async def get_training_devices(self) -> list[DeviceInfo]:
...
 
     async def train(self, context: TrainingContext) -> None:
-        """Push snapshot, submit job, mirror progress, and ingest the model."""
+        """Deliver the snapshot, submit the job, mirror progress, ingest the model."""
+        if self._dataset_transfer == "hf":
+            await self._train_via_hf(context)
+        else:
+            await self._train_via_http(context)
+
+    async def _train_via_http(self, context: TrainingContext) -> None:
```
MarkRedeman: This feels like something that could be hidden away via an interface.
```
class TrainingMethod:
    def train(self, context: TrainingContext):
        ...

class HttpTrainingMethod(TrainingMethod):
    def train(self, context: TrainingContext):
        ...

class HfTrainingMethod(TrainingMethod):
    def train(self, context: TrainingContext):
        ...
```

AlbertvanHouten: I agree, fixed now.

application/backend/src/cli/run.py
```diff
@@ -0,0 +1,156 @@
...
+_EXPORT_PREFIX_RE = re.compile(r"^export\s+")
+
+
+def _backend_dir() -> Path:
+    """Return the backend project directory (contains pyproject.toml and .venv)."""
+    # src/cli/run.py -> src/cli -> src -> backend
+    return Path(__file__).resolve().parents[2]
+
+
+def load_env_file(env_file: Path) -> None:
```
MarkRedeman, line 33: Could this not be done via our `Settings` object? It should automatically fill environment variables based on `.env`

application/backend/src/cli/run.py
MarkRedeman, line 1: perhaps rename this file to `remote.py` since it only exposes a `remote` command?

application/backend/src/cli/run.py
```diff
@@ -0,0 +1,156 @@
...
+        )
+
+    # Remote/recording nodes offload training, so cpu torch suffices; a GPU extra
+    # only matters for local torch-backend GPU inference.
+    if device_explicit and resolved_device != "cpu":
+        click.echo(
+            f"Note: DEVICE={resolved_device} on a remote node only affects local torch-backend "
+            "GPU inference; training is offloaded, so cpu torch suffices for recording.",
+        )
+    maybe_sync(_backend_dir(), resolved_device, sync=sync)
```
MarkRedeman, line 147: Why do we need to sync here? Shouldn't this only matter for the actual trainer process?

The user may still use cpu, xpu or cuda for inference, which might not be the same environment as the training environment with this new feature.

application/backend/src/services/training_backends/__init__.py
```diff
@@ -4,19 +4,33 @@
...
+from services.training_backends.base import (
+    ProgressReporter,
+    TrainingBackend,
+    TrainingCanceledError,
+    TrainingContext,
+    TrainingSuspendedError,
+)
 
 
 def get_training_backend() -> TrainingBackend:
```
MarkRedeman, line 20: If I understand correctly we now set the training backend as a default for the current server instance.

One idea I initially had was that each job would be responsible for resolving its own training backend. So for instance if later we have a `AWSTrainingBackend`, `HuggingFaceJobTrainingBackend` etc. then the job could include the targetted training backend in its payload (or the scheduler determines this automatically) and we then resolve the used backend based on that.

I think the same concept could also work if we'd have multiple of the same backends e.g.,
```
RemoteTrainingBackend("http://10.10.120.101:8080/"), 
RemoteTrainingBackend("http://10.10.120.103:8080/"),  # both of these stored in our local database as references
```

and then have the backend be resolved based on the job's payload:
```
if job.payload.training_backend == "remote":
    return RemoteTrainingBackend(job.payload.ip) # or maybe load it from db etc..?
```

No need to address this now, mainly sharing ideas here.

application/backend/src/services/training_backends/_training_methods.py
```diff
@@ -0,0 +1,84 @@
...
+
+    def __init__(self, backend: RemoteTrainingBackend) -> None:
+        self._backend = backend
+
+    @abstractmethod
+    async def train(self, context: TrainingContext) -> None:
+        """Deliver the snapshot, submit the job, mirror progress, ingest the model."""
+
+
+class HttpTrainingMethod(TrainingMethod):
```
MarkRedeman, line 33: Would having a `LocalTrainingMethod` make any sense here as well?

application/backend/src/services/training_backends/remote.py
```diff
@@ -0,0 +1,790 @@
...
+# Only these patterns are pulled by the trainer; mirrors snapshot_download allowlists.
+_SNAPSHOT_ALLOW_PATTERNS = ["*.safetensors", "*.json", "*.txt", "*.md", "*.parquet", "*.mp4", "*.png", "*.jpg"]
+_EVENT_WAIT_TIMEOUT_S = 3.0
+_RECONNECT_BACKOFF_S = 2.0
+_TERMINAL_STATES = {"completed", "failed", "canceled"}
+_UPLOAD_CHUNK_SIZE = 8 * 1024 * 1024  # 8 MB
+_DOWNLOAD_CHUNK_SIZE = 8 * 1024 * 1024  # 8 MB
+_TRANSFER_RETRY_LIMIT = 3
+
+_REMOTE_JOB_ID_PATTERN = re.compile(r"^[A-Za-z0-9_-]{1,128}$")
```
MarkRedeman: This feels like something that should be type guarded, if we give `remote_job_id: UUID` then we don't need to check this and instead make the creator of this job be responsible for constructing a UUID

application/backend/src/services/training_backends/_training_methods.py
```diff
@@ -0,0 +1,84 @@
...
+        archive_path: Path | None = None
+        if context.snapshot is None:
+            raise ValueError("HTTP dataset transfer requires a dataset snapshot")
+        try:
+            # Archive and upload the snapshot (0-10%).
+            context.progress(0, message="Preparing dataset snapshot")
+            archive = await asyncio.to_thread(
+                DatasetDownloadService().create_dataset_archive, Path(context.snapshot.path)
+            )
+            archive_path = archive
```
MarkRedeman: nit picky:

```suggestion
            archive_path = await asyncio.to_thread(
                DatasetDownloadService().create_dataset_archive, Path(context.snapshot.path)
            )
```

application/backend/src/cli/run.py
AlbertvanHouten: I will make a follow up PR removing this CLI completely which then reuses the serve command for starting the backend.

application/backend/src/services/training_backends/_training_methods.py
```diff
@@ -0,0 +1,84 @@
...
+            context.progress(SNAPSHOT_UPLOAD_PROGRESS, message="Dataset uploaded, starting training")
+
+            remote_job_id = await backend.submit_job(context, dataset_transfer="hf", repo_id=repo_id, revision=revision)
+
+            # Persist the id so a restart can reattach to the remote job.
+            if context.on_remote_job_id is not None:
+                await context.on_remote_job_id(remote_job_id)
+
+            await backend.await_and_ingest(context, remote_job_id)
+        finally:
```
MarkRedeman: I think you mentioned before that we don't upload the model using the hugging face api, so we don't need to delete it here. But why do we not use that?

application/backend/src/services/training_backends/remote.py
```diff
@@ -0,0 +1,790 @@
...
+from settings import get_settings
+
+if TYPE_CHECKING:
+    from collections.abc import AsyncIterator, Iterator
+
+    from services.training_backends._training_methods import TrainingMethod
+    from services.training_backends.base import TrainingContext
+
+# Only these patterns are pulled by the trainer; mirrors snapshot_download allowlists.
+_SNAPSHOT_ALLOW_PATTERNS = ["*.safetensors", "*.json", "*.txt", "*.md", "*.parquet", "*.mp4", "*.png", "*.jpg"]
```
MarkRedeman: Would we not also need `.ckpt` when retraining an existing model? Or I guess this only matters for the dataset snapshot?

application/backend/src/services/training_backends/remote.py
```diff
@@ -0,0 +1,790 @@
...
+        self._hf_token = os.environ.get("HF_TOKEN")
+        # Resolved once by _resolve_trust_env(): True honors proxy env vars,
+        # False bypasses them. None means "not yet probed".
+        self._trust_env: bool | None = None
+        self._trust_env_lock = asyncio.Lock()
+        # Suppress duplicate consecutive progress lines (e.g. the trainer
+        # re-emitting the final training state while it optimizes/exports).
+        self._last_progress_log: str | None = None
+
+    async def _resolve_trust_env(self) -> bool:
```
MarkRedeman, line 92: I think a lot of these safety checks could later be simplified/removed when we allow users to add training backends from the UI.
In that case we can do these checks before adding the backend, guaranteeing that any backend that we've added won't need these checks anymore.

application/backend/src/services/training_backends/remote.py
```diff
@@ -0,0 +1,790 @@
...
+        :meth:`_mirror_upload_progress`. Returns the repo id and the commit SHA.
+        """
+        from huggingface_hub import HfApi
+
+        if context.snapshot is None:
+            raise RemoteTrainingError("HuggingFace dataset transfer requires a dataset snapshot")
+
+        api = HfApi(token=self._hf_token)
+        repo_name = f"pais-snapshot-{uuid.uuid4().hex[:12]}"
+        requested_repo_id = f"{self._namespace}/{repo_name}" if self._namespace else repo_name
```
MarkRedeman: This feels like logic that should live inside of the related `training_method` since a http method wouldn't need to use the hf api.

application/backend/src/services/system_service.py
```diff
@@ -5,17 +5,24 @@
...
+if TYPE_CHECKING:
+    from services.training_backends.remote import RemoteTrainingBackend
 
 
 class SystemService:
     """Service to discover and report available compute hardware."""
 
+    # Cached remote backend, reused across calls so one-time setup (e.g. the
+    # trainer proxy probe in RemoteTrainingBackend) isn't repeated per request.
+    _remote_backend: ClassVar["RemoteTrainingBackend | None"] = None
```
MarkRedeman, line 24: This feels like something that should use dependency injection, i.e. have the fastapi lifecycle build the remote training backend and inject it into the `SystemService`'s constructor.

application/backend/README.md
```diff
@@ -55,12 +55,25 @@ Please check out [this document](docs/video_hardware_acceleration_intel.md) for
...
 
+The launch flows are exposed as `physicalai-studio` subcommands: `serve`
+(in-process/local training, the default) and `remote` (training offloaded to a
+remote trainer service). `remote` loads the matching `.env`, runs `uv sync`
+for the requested `DEVICE`, runs migrations, and starts the backend.
+`run.sh` simply forwards to these subcommands.
+
+The remote trainer service itself is a separate component, launched from
+`application/trainer` via its own `physicalai-trainer` command (see
+[`application/trainer/README.md`](../trainer/README.md)).
```
MarkRedeman: The docs above don't mention the remote training yet. I would try to rewrite this a bit. The part about subcommands doesn't seem relevant here.

application/trainer/README.md
```diff
@@ -32,27 +40,58 @@ Set environment variables (or an `.env` file):
 > The trainer service has no built-in authentication. Any machine that can reach its port can submit jobs, cancel jobs, and download trained model artifacts.
 > Deploy it on a private network reachable only by the Physical AI Studio backend IP address – never expose this port to the internet.
 
+> [!WARNING]
+> The Physical AI Studio backend inherits proxy settings (`HTTP_PROXY` / `HTTPS_PROXY`) from its host environment. When those variables are set, all trainer communication (including model artifact download) routes through the configured proxy, so whoever controls those variables controls where model artifacts flow. Only deploy the backend in a fully trusted environment where proxy variables cannot be set by other users. Do not run it on a shared or multi-tenant host.
```
MarkRedeman: Do we need this extra warning given that we already warn about not having authentication above?

application/backend/src/services/training_backends/_training_methods.py
AlbertvanHouten: local training is already under a separate LocalTrainingBackend which doesn't need different strategies as far as I am aware so adding it would just add unnecessary abstraction

application/backend/src/services/training_backends/remote.py
AlbertvanHouten: only for dataset indeed

application/trainer/src/trainer/archive_safety.py
MarkRedeman, line 1: Could we somehow reuse the same code from the main application? Or would that add a lot of maintenance overhead?

AlbertvanHouten: We could move it to the library and share it from there

application/backend/src/services/system_service.py
AlbertvanHouten: I agree, but this cache will most likely be removed or changed in a follow up PR where we can configure multiple remote trainers. I will address it there if it makes sense

application/trainer/src/trainer/archive_safety.py
AlbertvanHouten: moved to lib now

application/trainer/README.md
AlbertvanHouten: merged into one section now

MarkRedeman (approved):

PR #794, opened 2026-07-09
fix: torch export preprocessing

sovrasov:
# Pull Request

## Description

To support inputs coming directly from cameras (hwc, uint8), an extra preprocessor is required.
Now torch export adds than extra node, so `InferenceModel` with torch under the hood can process both fp32 and uint8 inputs

## Related Issues

https://github.com/openvinotoolkit/physicalai/pull/185


MarkRedeman (approved):

ivanzati (approved):
Thank you, Vladislav!

PR #799, opened 2026-07-10
refactor: move robot factory builder methods

MarkRedeman:
Prepare for robot plugin design by moving builders from a complicated `RobotClientFactory` into their own files. The goal is that each robot file (`so101.py`, `widowx.py`) will resemble a plugin. 
Plugin authors need to implement a similar interface to get their robot into studio.

Next step:
- Move the payload and robot interface into the plugin file
- Move discovery methods into plugin file

### DIFF
--- a/application/backend/src/robots/catalog/types.py
+++ b/application/backend/src/robots/catalog/types.py
@@ -13,6 +13,7 @@ if TYPE_CHECKING:
     from uuid import UUID
 
     from schemas.calibration import Calibration
+    from schemas.robot import SO101Robot
 
 
 @dataclass(frozen=True)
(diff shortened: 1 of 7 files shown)


application/backend/src/robots/catalog/types.py
```diff
@@ -1,11 +1,54 @@
...
+
+    from schemas.calibration import Calibration
+    from schemas.robot import SO101Robot
+
+
+@dataclass(frozen=True)
+class RobotAdapterOptions:
+    include_velocities: bool = False
+    goal_time_scale: float = 1.0
+    external_effort_gain: float | None = 0.1
```
AlbertvanHouten, line 28: Could we have some comments explaining what these options do?

AlbertvanHouten (approved):

PR #809, opened 2026-07-14
feat: MolmoAct2 first party implementation

alfieroddan:
# Pull Request

## Description

This PR integrates MolmoACT2 as a first party policy.

Useful links:
- https://allenai.org/blog/molmoact2
- https://huggingface.co/docs/lerobot/en/molmoact2
- https://arxiv.org/abs/2605.02881

## Policy Design Changes

Implemented following the new policy design PR - https://github.com/open-edge-platform/physical-ai-studio/pull/1040

## Future PRs
- [ ] Enable RTC
- [ ] Enable depth modality in PAS - then we can enable MolmoAct2-Think

## Unrelated Changes

- Gymnasium
  - In order to adapt LIBERO benchmark to the pretrained policy, we've added a camera re-mapping. I've extended this to all gyms.
  - We've also added a to device functionality, this ensures the output of the benchmark is compatible with the policy.

## Runtime

There is a sister PR in the other package - https://github.com/openvinotoolkit/physicalai/pull/234


library/configs/physicalai/molmoact2.yaml
```diff
@@ -0,0 +1,36 @@
...
+model:
+  class_path: physicalai.policies.MolmoAct2
+  init_args:
+    # Pretrained checkpoint
+    repo_id: allenai/MolmoAct2-LIBERO
+    norm_tag: libero
+    load_weights: true
+    # Input / output structure
+    n_obs_steps: 30
+    n_action_steps: 10
```
alfieroddan: These aren't args of the model

library/src/physicalai/inference/adapters/pytorch.py
alfieroddan, line 1: These changes should be another PR - leave for now so users can test MolmoAct2 on edge devices

library/src/physicalai/policies/molmoact2/model/vision.py
```diff
@@ -0,0 +1,331 @@
...
+
+if TYPE_CHECKING:
+    from physicalai.policies.molmoact2.config import MolmoAct2AdapterConfig, MolmoAct2VitConfig
+
+
+class VisionMultiHeadAttention(nn.Module):
+    """Multi-head (optionally grouped) attention used by the ViT and the pooler.
+
+    A single SDPA path replaces the upstream eager/sdpa/flash branches. When
+    ``float32_attention`` is set, q/k/v are promoted to float32 for the scaled
```
alfieroddan: Do we need to promote to `float32`? We should keep whatever dtype training set.

library/src/physicalai/policies/molmoact2/backbone.py
```diff
@@ -0,0 +1,372 @@
...
+        *,
+        input_ids: torch.Tensor,
+        attention_mask: torch.Tensor | None = None,
+        token_type_ids: torch.Tensor | None = None,
+        images: torch.Tensor | None = None,
+        token_pooling: torch.Tensor | None = None,
+        action_dim_is_pad: torch.Tensor | None = None,
+        action_horizon: int,
+        num_steps: int | None = None,
+        sample_noise: bool = False,
```
alfieroddan, line 238: Rename to match SmolVLA's sample noise generation option.

library/src/physicalai/policies/molmoact2/processors/factory.py
```diff
@@ -0,0 +1,42 @@
...
+        raise ValueError(msg)
+
+    preprocessor = MolmoAct2Preprocessor(config=config)
+    postprocessor = MolmoAct2Postprocessor(
+        output_features=config.output_features,
+        adapt_to_so101=config.adapt_to_so101,
+        joint_signs=config.joint_signs,
+        joint_offsets=config.joint_offsets,
+    )
+    return preprocessor, postprocessor
```
alfieroddan: Refactor to have preprocessor config and post-processor config.

library/src/physicalai/policies/molmoact2/processors/preprocess_steps.py
```diff
@@ -0,0 +1,377 @@
...
+        """
+        if image.dtype == torch.uint8:
+            pixels = image
+        else:
+            pixels = image.to(torch.float32)
+            if float(pixels.max()) <= 1.0:
+                pixels *= 255.0
+            pixels = pixels.clamp(0, 255).to(torch.uint8)
+        resized = tv_functional.resize(pixels, [self.height, self.width], antialias=False)
+        return resized.to(torch.float32) / 255.0
```
alfieroddan: offload resize to a different pre-processing step.

library/src/physicalai/policies/molmoact2/processors/tokenizers.py
```diff
@@ -0,0 +1,265 @@
...
+        # ``tokenizer.json`` (the vocab) from the Hub.
+        repo_id = repo_or_path
+        try:
+            tokenizer_json_path = hf_hub_download(repo_id, _TOKENIZER_JSON_FILENAME)
+        except RemoteEntryNotFoundError as exc:
+            msg = (
+                f"Could not resolve MolmoAct2 tokenizer: no local '{_TOKENIZER_JSON_FILENAME}' "
+                f"and '{_TOKENIZER_JSON_FILENAME}' not found in repo '{repo_id}'."
+            )
+            raise FileNotFoundError(msg) from exc
```
alfieroddan: Don't like case 2, but without the external vocab json we can't use FAST tokenizer. Possibly we could remove with defaults and not use FAST.

library/src/physicalai/policies/molmoact2/__init__.py
```diff
@@ -0,0 +1,31 @@
+# Copyright 2026 The Allen Institute for Artificial Intelligence and The HuggingFace Inc. team.
+
+# Copyright (C) 2026 Intel Corporation
+# SPDX-License-Identifier: Apache-2.0
+
+"""MolmoAct2 Policy."""
+
+from physicalai.utils.hf_utils import HuggingfacePolicyContainer
```
alfieroddan: Don't need to export HuggingfacePolicyContainer

library/src/physicalai/policies/molmoact2/config.py
alfieroddan, line 1: Remove un-used config options for 1st party code. Then we can unify approach.

library/src/physicalai/policies/molmoact2/policy.py
alfieroddan, line 1: Add `from_config` method. We should also work out how to load model that's been pre-trained but exported - it shouldn't reload pre-trained weights for loading.

library/src/physicalai/policies/utils/features.py
```diff
@@ -0,0 +1,29 @@
+# Copyright 2026 The Allen Institute for Artificial Intelligence and The HuggingFace Inc. team.
```
alfieroddan: Remove copyright.

library/src/physicalai/policies/utils/normalization.py
alfieroddan, line 1: Remove changes to normaliztion. Add own normalization method if we need the changes for Molmo specifically.

library/tests/unit/policies/molmoact2/conftest.py
alfieroddan, line 1: remove hf_container tests and move to it's own unit test.

library/src/physicalai/policies/molmoact2/config.py
```diff
@@ -0,0 +1,601 @@
...
+    # Component sub-configs
+    vit_config: MolmoAct2VitConfig = field(default_factory=MolmoAct2VitConfig)
+    adapter_config: MolmoAct2AdapterConfig = field(default_factory=MolmoAct2AdapterConfig)
+    text_config: MolmoAct2TextConfig = field(default_factory=MolmoAct2TextConfig)
+    action_expert_config: MolmoAct2ActionExpertConfig | None = field(default_factory=MolmoAct2ActionExpertConfig)
+
+    # Input and rollout structure
+    n_obs_steps: int = 30
+    chunk_size: int = 30
+    n_action_steps: int = 30
```
alfieroddan: n_action_steps align with Policy init and the rest of the policies.

library/pyproject.toml
```diff
@@ -108,6 +108,14 @@ smolvla = [
     "safetensors>=0.4.3,<1.0.0",
 ]
 
+# MolmoAct2
+# Supports LoRA fine-tuning for efficient adaptation
+molmoact2 = [
+   "transformers>=5.3.0,<5.4.0",
```
alfieroddan: Seems to have conflict wth uv.lock on main.

AGENTS.md
alfieroddan, line 1: Revert.

library/configs/physicalai/molmoact2.yaml
```diff
@@ -0,0 +1,37 @@
...
+    chunk_size: 30
+    # Action mode (only "continuous" is currently supported)
+    action_mode: continuous
+    # SO-100/101 joint calibration correction (needed for zero-shot / finetuning
+    # from the pre-#777 LeRobot calibration checkpoint)
+    adapt_to_so101: false
```
sovrasov: Would it be useful to have a standalone zero-shot config?

library/pyproject.toml
eugene123tw: @alfieroddan could you share more detail about the conflict? Current `uv.lock` looks fine, but I'm just curious about the python version downgrading from  ">=3.14" to  ">=3.13"

library/src/physicalai/benchmark/gyms/libero/libero.py
```diff
@@ -83,6 +83,7 @@ def __init__(
         self,
         task_suite: str = "libero_10",
         task_ids: list[int] | None = None,
+        camera_name_mapping: dict[str, str] | None = None,
```
eugene123tw, line 95: camera_name_mapping is missing from the Args: section in the class-level docstring. Could you add it with a brief description and an example?

An example would also help. 

```py
>>> # Rename the wrist camera key to match what MolmoAct2 expects
>>> benchmark = LiberoBenchmark(
...     task_suite="libero_10",
...     camera_name_mapping={"image2": "wrist_image"},
... )
```

Also worth considering: pusht benchmarks don't currently expose camera_name_mapping, but if the underlying gyms support it, threading it through at the Benchmark level (or via **kwargs forwarded to the gym factory) could make this pattern reusable rather than LIBERO-specific. Is there a reason to scope it only here?

library/src/physicalai/data/observation.py
```diff
@@ -481,6 +481,7 @@ class NormalizationParameters:
     max: list[float] | float | None = None
     q01: list[float] | float | None = None
     q99: list[float] | float | None = None
+    mask: list[bool] | None = None
```
eugene123tw, line 496: The new `mask` field on `NormalizationParameters` has no docstring, and its semantics are not obvious compared to the other fields (which are statistics). Could you add a field-level doc comment explaining what it represents? 

Also worth noting: `NormalizationParameters` is a shared dataclass, but this field is only consumed by MolmoAct2's postprocessor. Is mask intended to be a general-purpose feature of the normalization contract, or is it MolmoAct2-specific? If the latter, it might be worth a comment flagging that the base FeatureNormalizeTransform silently ignores it, so other policies won't accidentally rely on it.

library/src/physicalai/policies/molmoact2/action_expert.py
```diff
@@ -0,0 +1,547 @@
...
+    """Precomputed per-step context shared across denoising steps."""
+
+    kv_contexts: Sequence[KVContext]
+    cross_mask: torch.Tensor | None
+    self_mask: torch.Tensor | None
+    valid_action: torch.Tensor | None
+    rope_cache: tuple[torch.Tensor, torch.Tensor] | None
+
+
+class ActionExpertRMSNorm(nn.Module):
```
eugene123tw: Could we reuse `library/src/physicalai/policies/pi0/components/attention.py::AdaRMSNorm`? If it's possible we could move it to a shared folder, and refactor import for both files.

library/src/physicalai/policies/molmoact2/text.py
```diff
@@ -0,0 +1,310 @@
...
+        Repeated values of the KV values to match the number of heads.
+    """
+    batch, num_key_value_heads, slen, head_dim = hidden_states.shape
+    if n_rep == 1:
+        return hidden_states
+    hidden_states = hidden_states[:, :, None, :, :].expand(batch, num_key_value_heads, n_rep, slen, head_dim)
+    return hidden_states.reshape(batch, num_key_value_heads * n_rep, slen, head_dim)
+
+
+class MolmoAct2RMSNorm(nn.Module):
```
eugene123tw: Could you check if ActionExpertRMSNorm and MolmoAct2RMSNorm are the same? If they are the same we could potentially remove one. That said, it's not a stopper.

library/src/physicalai/utils/hf_utils.py
```diff
@@ -0,0 +1,196 @@
...
+    "resume_download",
+    "proxies",
+    "token",
+    "revision",
+    "local_files_only",
+}
+
+
+@dataclass
+class HuggingfacePolicyContainer:
```
eugene123tw: A few specific questions:

1. Why a new shared module rather than the existing pattern? Pi05Policy._from_hf and SmolVLAPolicy._from_hf download their artifacts inline and return a plain tuple. MolmoAct2 instead externalises the download logic into a shared util and wraps the result in HuggingfacePolicyContainer. What drove that change, was it the sharded-weights support, the extra fields (tokenizer_config, norm_stats), or something else? 

2. Is HuggingfacePolicyContainer the right abstraction boundary? If the goal is just to avoid a long tuple return from load_hf_pretrained_container, a TypedDict or a private named tuple scoped to molmoact2/from_hf.py carries the same information without claiming space in the shared namespace. Moving it to shared utils/ sets a precedent that every new policy can add its own dataclass there.

3. If it is genuinely reusable, can pi05 and smolvla be migrated to it? If HuggingfacePolicyContainer is meant to be the shared loading contract going forward, it should either be adopted by the existing policies in this PR or tracked as a follow-up issue. A utility used by exactly one policy in utils/ is harder to justify than one with two or more adopters.

I think the concern isn't with the implementation itself, the sharded-weight handling is useful but with scoping. Could this live in molmoact2/from_hf.py for now, with a migration to utils/ happening when a second policy needs it?

library/src/physicalai/policies/molmoact2/from_hf.py
```diff
@@ -0,0 +1,731 @@
...
+                "local_files_only",
+            }
+        }
+        (
+            config_file,
+            weights_file,
+            preprocessor_file,
+            preprocessor_dir,
+            norm_stats_file,
+        ) = download_policy_artifacts_from_hub(
```
eugene123tw: `_ensure_processor_assets_downloaded` exists solely because `download_policy_artifacts_from_hub` doesn't download `processor_config.json`, but `load_hf_pretrained_container` needs it 10 lines later. I don't think we need a dedicated _ensure_processor_assets_downloaded function, it's just passing `processor_config.json` to `download_policy_artifacts_from_hub` directly:

```py
config_file, weights_file, preprocessor_file, preprocessor_dir, norm_stats_file, processor_config_file = \
    download_policy_artifacts_from_hub(
        ...,
        extra_files=["processor_config.json"],  # or an additional param
    )
```

library/src/physicalai/policies/molmoact2/from_hf.py
```diff
@@ -0,0 +1,731 @@
...
+        try:
+            downloaded = hf_hub_download(candidate, "tokenizer_config.json")
+        except RemoteEntryNotFoundError:
+            continue
+        with Path(downloaded).open(encoding="utf-8") as f:
+            return json.load(f)
+    return None
+
+
+def _resolve_local_weights_path(checkpoint_dir: Path) -> Path:
```
eugene123tw: Could we remove `_resolve_local_weights_path` entirely and inline the two-line check directly in `load_hf_pretrained_container`:

```py
if is_local:
    weights_file = path / SAFE_WEIGHTS_NAME
    if not weights_file.is_file():
        weights_file = path / SAFE_WEIGHTS_INDEX_NAME
    if not weights_file.is_file():
        raise FileNotFoundError(...)
```

That's 4 lines instead of a dedicated function, and it keeps local and Hub resolution visibly co-located in the same place that knows whether it's dealing with a local path or a repo ID. The function only earns its keep if the same check needs to be reused elsewhere and currently it isn't.

library/src/physicalai/policies/molmoact2/from_hf.py
```diff
@@ -0,0 +1,731 @@
...
+        [feature for feature in unmatched_pretrained if id(feature) not in consumed_pretrained_ids],
+    )
+    resolved_features.extend(
+        [feature for feature in unmatched_overrides if id(feature) not in consumed_override_ids],
+    )
+
+    return resolved_features
+
+
+def build_config_from_hf_config(  # noqa: PLR0913
```
eugene123tw: `from_hf.py` is hard to follow because three separate concerns are mixed without clear boundaries: artifact loading, HF→Studio config translation, and feature/normalization resolution. Some specific details:

1. `build_config_from_hf_config` has 20 parameters and works by building a config_data dict through a long sequence of manual key assignments, some from hf_config, some from caller arguments, some hardcoded. There's no single place that shows you the complete shape of the output config. A comment at the top of the function explaining the three stages (extract from HF, build features, apply caller overrides) would make it much easier to orient.

2. `_resolve_feature_overrides` is the most difficult part. It does name-based matching first, then falls back to type+shape matching for camera alias remaps — using id() to track consumed items. The intent (preserve pretrained normalization stats when a user renames a camera key like image2 → wrist_image) is not obvious from the code. This needs at least a diagram or a concrete example in the docstring. It's also non-trivial logic that currently has no unit tests. What happens if two cameras have the same shape?

3. The overall flow is only visible if you read load_hf_pretrained_container top-to-bottom and mentally trace through 10 helper functions. A module-level comment block (like a mini-README) showing the pipeline:

library/src/physicalai/policies/molmoact2/README.md
```diff
@@ -0,0 +1,234 @@
...
+        record_mode="all",
+        video_dir="./videos/",
+    )
+
+    results = benchmark.evaluate(policy)
+    summary = results.summary()
+
+    block_header = f"\n{'=' * 28} libero_10 {'=' * 28}\n"
+    print(block_header, end="")
+    print(summary)
```
eugene123tw: Could you share the libero result in the README?

library/src/physicalai/policies/molmoact2/policy.py
```diff
@@ -0,0 +1,755 @@
...
+        gradient_checkpointing: bool = False,
+    ) -> None:
+        """Initialize a MolmoAct2 policy wrapper.
+
+        Args:
+            input_features: Optional observation feature schema.
+            output_features: Optional action feature schema.
+            repo_id: Optional pretrained checkpoint identifier or path. Ignored
+                if ``config`` is provided.
+            norm_tag: Optional normalization tag for pretrained checkpoints.
```
eugene123tw: What does `norm_tag` do? And what are the options for `norm_tags`?

library/src/physicalai/policies/molmoact2/README.md
```diff
@@ -0,0 +1,234 @@
...
+policy_resumed = MolmoAct2.load_from_checkpoint(checkpoint_path)
+
+# Force the (normally skipped) pretrained weight load anyway, e.g. if you
+# specifically need the original pretrained weights loaded first for some
+# custom reason before the checkpoint's own weights get applied on top:
+policy_resumed_force_pretrained = MolmoAct2.load_from_checkpoint(
+    checkpoint_path,
+    load_weights=True,
+)
+```
```
eugene123tw, line 314: As molmoact2 supports zero-shot prediction. Would it be possible to show that in README or a notebook?

eugene123tw (commented):
Thanks for the PR, Alfie! The integration of MolmoAct2 policy, training integration, and benchmark are solid. A few things to address before merge:

**Bug to fix**
- `_load_tokenizer_config` calls `hf_hub_download` without forwarding `hub_kwargs`, so `token`, `revision`, etc. are silently dropped. This will break loading from gated repos.

**Decisions needed (team input)**
- `MolmoAct2Config` uses nested HF-style sub-configs while all other policies use flat configs. Is this intentional? If yes, please document the rationale. If no, flatten it or open a follow-up tracking the migration.
- `HuggingfacePolicyContainer` and `download_policy_artifacts_from_hub` live in shared `utils/` but are only used here. Move them to `molmoact2/from_hf.py` for now — we can promote to shared once a second policy needs them.
- `NormalizationParameters.mask` is shared but only MolmoAct2 consumes it, and the base normalizer silently ignores it. Please add a docstring clarifying the intent and whether this is expected to be general.
- Clarification about `norm_tag`

**Readability asks**
- from_hf.py is hard to follow as a whole. Please add a module-level pipeline comment showing the loading flow, stage comments inside `build_config_from_hf_config`, and a concrete example in `_resolve_feature_overrides` — the `id()`-based type/shape fallback matching especially needs explanation.
- `_ensure_processor_assets_downloaded` loops over a one-item list and duplicates the hub_kwargs allowlist. Simplify to a direct call.
- `camera_name_mapping` is missing from `LiberoBenchmark`'s `Args` docstring — add description + example.

**Test coverage**
- `_resolve_feature_overrides` has non-trivial fallback logic with edge cases (e.g. two cameras with identical shapes). Please add unit tests covering at least: name match, type/shape fallback match, and the duplicate-shape ambiguous case.

**README**

- Would be nice to add libero benchmark number
- Would be nice to have a zero-shot tutorial or guide for Molmoact2 as it's a highlight feature

library/pyproject.toml
sovrasov: We would use `>=5.5` only for all models due to safety reasons.
Perhaps a wrapper for transformers dep can be introduced, as we need to enforce the same version constraints for all models, like lerobot does

library/src/physicalai/policies/molmoact2/policy.py
```diff
@@ -0,0 +1,755 @@
...
+        Returns:
+            A constructed :class:`MolmoAct2` policy.
+
+        Raises:
+            TypeError: If ``config`` is not a dataclass instance (required
+                for ``dataclasses.replace``), or if ``overrides`` contains
+                keys that are not fields on ``config``.
+        """
+        if overrides:
+            if not dataclasses.is_dataclass(config):
```
sovrasov: Isn't the config object always a dataclass according to the annotation?

library/src/physicalai/policies/molmoact2/policy.py
```diff
@@ -0,0 +1,755 @@
...
+    def extra_export_args(self) -> dict[str, ExportParameters]:
+        """Extra backend export args for inference-time pre/post graph components."""
+        output_names = [feature.name for feature in (self.outputs_schema or [])]
+
+        return {
+            "torch": TorchExportParameters(
+                input_names=[TOKENIZED_PROMPT, TOKENIZED_PROMPT_MASK, IMAGES, IMAGE_MASKS, STATE],
+                output_names=output_names,
+                preprocessors_specs=[ComponentSpec(type="to_float_tensor")],
+            ),
```
sovrasov, line 1380: It looks like we need an optional chunk trimmer here (depends on config, like in other policies)

library/src/physicalai/policies/molmoact2/policy.py
```diff
@@ -0,0 +1,755 @@
...
+        return outputs
+
+    @property
+    def extra_export_args(self) -> dict[str, ExportParameters]:
+        """Extra backend export args for inference-time pre/post graph components."""
+        output_names = [feature.name for feature in (self.outputs_schema or [])]
+
+        return {
+            "torch": TorchExportParameters(
+                input_names=[TOKENIZED_PROMPT, TOKENIZED_PROMPT_MASK, IMAGES, IMAGE_MASKS, STATE],
```
sovrasov: Input names are ignored here by torch adapter: it doesn't try to analyze inputs, just forwards it to the underlying policy object

library/src/physicalai/policies/molmoact2/policy.py
```diff
@@ -0,0 +1,755 @@
...
+        norm_tag: str | None = None,
+        n_obs_steps: int = 30,
+        n_action_steps: int = 30,
+        chunk_size: int = 30,
+        action_mode: Literal["continuous", "discrete", "both"] = "continuous",
+        *,
+        config: MolmoAct2Config | None = None,
+        use_random_input_noise: bool = False,
+        adapt_to_so101: bool = False,
+        torch_compile: bool = False,
```
alfieroddan: Change to compile_model. This is in-line with other models.

library/src/physicalai/policies/molmoact2/policy.py
```diff
@@ -0,0 +1,755 @@
...
+            self._checkpoint_location = self.hf_container.checkpoint_location if self.hf_container is not None else None
+
+        # SO-101 joint calibration correction (must be set before processors build).
+        # Applied uniformly regardless of which branch above built the config.
+        # https://huggingface.co/docs/lerobot/v0.6.0/en/molmoact2#joint-frame-transform-so-100101-zero-shot
+        self.config.adapt_to_so101 = adapt_to_so101
+
+        # Keep repo_id in checkpoint hparams so load_from_checkpoint reconstructs
+        # the same pretrained source during inference adapter reload.
+        self.save_hyperparameters(ignore=["config", "load_weights"])
```
alfieroddan: Don't save compile model as well. Perhaps don't save input and output features?

library/src/physicalai/policies/molmoact2/policy.py
```diff
@@ -0,0 +1,755 @@
...
+        input_features: list[Feature] | None = None,
+        output_features: list[Feature] | None = None,
+        repo_id: str | Path | None = "allenai/MolmoAct2",
+        norm_tag: str | None = None,
+        n_obs_steps: int = 30,
+        n_action_steps: int = 30,
+        chunk_size: int = 30,
+        action_mode: Literal["continuous", "discrete", "both"] = "continuous",
+        *,
+        config: MolmoAct2Config | None = None,
```
samet-akcay: we have both `args`/`kwargs` and `config`. what is the rule to follow here ?

library/src/physicalai/policies/molmoact2/README.md
```diff
@@ -0,0 +1,234 @@
...
+
+## Quick Start: Forward Pass
+
+```python
+import torch
+
+from physicalai.data import Feature, FeatureType, Observation
+from physicalai.policies import MolmoAct2
+
+DEVICE = "cuda"
```
samet-akcay: maybe use `get_device()` in `physicalai.devices.utils`, we have a util function for this?

library/src/physicalai/policies/molmoact2/README.md
```diff
@@ -0,0 +1,234 @@
...
+from physicalai.train import Trainer
+
+if __name__ == "__main__":
+    # dataset sets input / output features
+    policy = MolmoAct2()
+    policy.config.use_random_input_noise = True
+    policy.config.train_action_expert_only = True
+
+    dm = LeRobotDataModule(
+        repo_id="MarkRedeman/dice-cleanup-combined",
```
samet-akcay: should we use this repo_id in the docs?

library/src/physicalai/policies/utils/features.py
```diff
@@ -0,0 +1,27 @@
...
+
+from __future__ import annotations
+
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from physicalai.data import Feature, FeatureType
+
+
+def feature_by_type(features: list[Feature], feature_type: FeatureType) -> Feature | None:
```
samet-akcay: how about `get_feature_by_type` or `get_feature(features, type: FeatureType)`?

samet-akcay (commented):
gone through the first round of review. great pr, just some minor comments..

library/src/physicalai/policies/molmoact2/policy.py
```diff
@@ -0,0 +1,755 @@
...
+                    ),
+                )
+            schema.append(
+                InferenceFeature(
+                    ftype=InferenceFeatureType.LANGUAGE,
+                    shape=(),
+                    name=TASK,
+                    dtype=InferenceFeatureDtype.STRING,
+                ),
+            )
```
daniil-lyakhov: Opus 4.8 just found this: should we really append a text feature after each non text feature here?

library/src/physicalai/policies/molmoact2/policy.py
alfieroddan: This is tough to explain in a init. But I've tried my best in the README.

library/configs/physicalai/molmoact2.yaml
alfieroddan: I think it would but we don't have CLI export enabled yet. 

I think it would make sense to make a zero-shot export for SO-101 and others.

library/src/physicalai/policies/molmoact2/policy.py
alfieroddan: https://github.com/open-edge-platform/physical-ai-studio/blob/511046fa2dfdfbfef668e70a9ecd827ce216486b/library/src/physicalai/policies/pi05/policy.py#L755-L760

I think it is necessary for models with text input for OV etc

library/configs/physicalai/molmoact2.yaml
alfieroddan: I'll add it to docs for now until export CLI works.

library/src/physicalai/policies/molmoact2/action_expert.py
alfieroddan: Unfortunately not - I've re-used it within molmo

eugene123tw (dismissed):

library/src/physicalai/gyms/gymnasium_gym.py
```diff
@@ -27,6 +27,51 @@
 logger = logging.getLogger(__name__)
 
 
+def validate_camera_name_mapping(
```
samet-akcay: these util functions look like they could be generic util functions? are these only used in gyms?

library/src/physicalai/policies/molmoact2/processors/postprocessor.py
```diff
@@ -0,0 +1,81 @@
...
+
+class MolmoAct2Postprocessor(torch.nn.Module):
+    """Convert normalized MolmoAct2 outputs back into action space."""
+
+    def __init__(
+        self,
+        *,
+        output_features: list[Feature],
+        normalization_mode: str = "QUANTILES",
+        adapt_to_so101: bool = False,
```
samet-akcay, line 34: is this arg scalable ? say we have many robots in the future? will we need to add different args for each?

library/src/physicalai/policies/molmoact2/processors/postprocessor.py
```diff
@@ -0,0 +1,81 @@
...
+        Args:
+            batch: Batch containing ACTION or "actions" tensor.
+
+        Returns:
+            Batch with ACTION denormalized and clamped.
+
+        Raises:
+            ValueError: If no action tensor is present in the batch.
+        """
+        batch = dict(batch)
```
samet-akcay, line 67: batch is annotated as dict, and converted to dict here. what else could it be? Observation? or anything else?

library/src/physicalai/policies/molmoact2/processors/preprocessor.py
```diff
@@ -0,0 +1,344 @@
...
+)
+from .tokenizers import MolmoAct2Tokenizers
+
+if TYPE_CHECKING:
+    from transformers import Qwen2Tokenizer
+
+    from physicalai.policies.molmoact2.config import MolmoAct2Config
+
+
+def _image_input_size(config: MolmoAct2Config) -> tuple[int, int]:
```
samet-akcay: going through the pr, and I'm a bit confused about the boundary regarding where to pass `config` vs `args`? 

Can you define it in your proposal?

library/src/physicalai/gyms/gymnasium_gym.py
alfieroddan: I think only gym utils. Feature remapping is pretty custom to each policy at the moment. This is more a data util focused for gyms.

library/src/physicalai/policies/molmoact2/processors/postprocessor.py
alfieroddan: this is just a shallow copy - to make sure we don't mutate batch when it's used somewhere else

library/src/physicalai/policies/molmoact2/processors/postprocessor.py
alfieroddan: I see what you mean.

for lots and lots of embodiments the solution is probably something like:

```python
from dataclasses import dataclass

@dataclass(frozen=True)
class RobotFrameConfig:
    name: str = "so101"
    joint_signs: list[float] | None = None
    joint_offsets: list[float] | None = None

class MolmoAct2Postprocessor(torch.nn.Module):
    def __init__(
        self,
        *,
        output_features: list[Feature],
        normalization_mode: str = "QUANTILES",
        robot_frame: RobotFrameConfig | None = None,
    ) -> None:
        super().__init__()
        self.robot_frame = robot_frame or RobotFrameConfig(name="identity")
        self._joint_transform = (
            JointFrameTransform(
                self.robot_frame.joint_signs or [],
                self.robot_frame.joint_offsets or [],
            )
            if self.robot_frame.name == "so101"
            else None
        )
...
```
But I doubt we'd need this much. I'd probably go for something simpler like:

```python
def __init__(
    self,
    *,
    output_features: list[Feature],
    normalization_mode: str = "QUANTILES",
    so101_config: dict[str, list[float]] | None = None,
) -> None:
    ...
    self._joint_transform = (
        JointFrameTransform(
            so101_config.get("joint_signs", []),
            so101_config.get("joint_offsets", []),
        )
        if so101_config is not None
        else None
    )
```

Would you be happy with so101_config?

library/src/physicalai/policies/molmoact2/processors/preprocessor.py
alfieroddan: Have defined in proposal

library/src/physicalai/policies/molmoact2/policy.py
```diff
@@ -0,0 +1,895 @@
...
+        tokenizer_dir = Path(self.config.tokenizer_name_or_path)
+        if (tokenizer_dir / "tokenizer.json").is_file() and self.config.tokenizer_config is not None:
+            return
+        resolved_dir, tokenizer_config = resolve_tokenizer_assets(self.config.tokenizer_name_or_path)
+        self.config.tokenizer_name_or_path = resolved_dir
+        self.config.tokenizer_config = tokenizer_config
+
+        # parameter setting based on config
+        if self.config.train_action_expert_only:
+            self._freeze_non_action_expert_parameters()
```
alfieroddan: Move to _init model - has been seperated by _ensure_tokenizer_assets.

library/src/physicalai/policies/molmoact2/policy.py
```diff
@@ -0,0 +1,1402 @@
...
+class MolmoAct2(ExportablePolicyMixin, Policy):  # noqa: PLR0904
+    """MolmoAct2 policy wrapper for loading pretrained checkpoints and configs."""
+
+    def __init__(  # noqa: PLR0913, PLR0915
+        self,
+        # Input and output features
+        input_features: list[Feature] | None = None,
+        output_features: list[Feature] | None = None,
+        # Pretrained model and normalization tag
+        pretrained_name_or_path: str | Path | None = "allenai/MolmoAct2",
```
sovrasov, line 123: In other policies this param is none by default to prevent loading the mentioned weights by default when loading already trained checkpoint. Are the specified here weights loaded in all possible initialization paths?

library/src/physicalai/policies/molmoact2/policy.py
```diff
@@ -0,0 +1,1402 @@
...
+        self.optimizer_grad_clip_norm = optimizer_grad_clip_norm
+        self.scheduler_warmup_steps = scheduler_warmup_steps
+        self.scheduler_decay_steps = scheduler_decay_steps
+        self.scheduler_decay_lr = scheduler_decay_lr
+
+        # initialize super
+        super().__init__(n_action_steps=self.n_action_steps)
+
+        # ignore input and output features, subject to change
+        self.save_hyperparameters(ignore=["input_features", "output_features", "compile_model"])
```
sovrasov, line 250: Same here about pre-trained weights, should we exclude them? If model checkpoint is saved, it already contains all the weights, we don't need to store the 3rd party weights source

library/src/physicalai/policies/molmoact2/policy.py
alfieroddan: I have changed the lightning loading to remove it. If you prefer to keep in-line with other policies, I'm happy to change it. Not sure users would ever use without a pre-trained path - especially given custom tokenizers.

```python
    @classmethod
    def load_from_checkpoint(
        cls,
        checkpoint_path: str | Path | IO[bytes],
        map_location: torch.device | str | int | Callable | dict | None = None,
        hparams_file: str | Path | None = None,
        strict: bool | None = None,  # noqa: FBT001
        weights_only: bool | None = None,  # noqa: FBT001
        **kwargs: Any,  # noqa: ANN401
    ) -> MolmoAct2:
        """Load a trained policy without resolving pretrained model weights.

        The checkpoint config rebuilds the model before Lightning restores its state dict.
        Tokenizer assets referenced by the config must remain available locally.

        Returns:
            The restored policy in the checkpoint's saved training mode.
        """
        kwargs["pretrained_name_or_path"] = None
        return super().load_from_checkpoint(
            checkpoint_path,
            map_location=map_location,
            hparams_file=hparams_file,
            strict=strict,
            weights_only=weights_only,
            **kwargs,
        )
```

library/src/physicalai/policies/molmoact2/policy.py
```diff
@@ -0,0 +1,1402 @@
...
+        if user_eager or pretrained_eager:
+            self.initialize_model()
+
+    @classmethod
+    def from_config(  # noqa: PLR0913
+        cls,
+        config: MolmoAct2Config,
+        *,
+        compile_model: bool = False,
+        openvino_compress_to_fp16: bool = False,
```
sovrasov, line 271: Is this arg used further in the method?

library/src/physicalai/policies/molmoact2/policy.py
```diff
@@ -0,0 +1,1402 @@
...
+                image_use_col_tokens=config.image_use_col_tokens,
+                use_single_crop_col_tokens=config.use_single_crop_col_tokens,
+                use_single_crop_start_token=config.use_single_crop_start_token,
+                image_token_ids=image_token_ids,
+            ),
+        ]
+        return {
+            ExportBackend.TORCH: TorchExportParameters(
+                preprocessors_specs=[ComponentSpec(type="to_float_tensor")],
+            ),
```
sovrasov, line 1380: Should we apply chunk trimmer to comply with `n_action_steps` or it's done somewhere else?

sovrasov (commented):
Thanks Alfie! Great PR overall. Have an extra minor comment: I don't see any changes in integration tests. Were they able to pick up the new policy without modifications?

library/src/physicalai/policies/molmoact2/policy.py
alfieroddan: Refactored to use chunk trimmer in https://github.com/open-edge-platform/physical-ai-studio/pull/809/commits/74bae433d302fe3f10608c902c8c42f806f0c931

alfieroddan:
> Thanks Alfie! Great PR overall. Have an extra minor comment: I don't see any changes in integration tests. Were they able to pick up the new policy without modifications?

I've added integration tests in https://github.com/open-edge-platform/physical-ai-studio/pull/809/commits/a8d854ba82a173795e22a560e7b31e49e3e0d6ce, can't use the base e2e class as we have to download the tokenizer (which are packaged with the base weights). But otherwise fine.

library/src/physicalai/gyms/utils.py
```diff
@@ -0,0 +1,55 @@
...
+
+    mapped_keys = [mapping.get(key, key) for key in camera_keys]
+    if len(mapped_keys) != len(set(mapped_keys)):
+        msg = f"camera_name_mapping creates duplicate output keys: {mapped_keys}"
+        raise ValueError(msg)
+
+    return dict(mapping)
+
+
+def remap_camera_images(
```
kdijkstraintel: The camera_mapping requires a lot of changes. I'm wondering is this issue should not be solved in a more general way. For example, whenever there are other modalities that need to be remapped in the future. 

I was wondering why not wrap a model (Molmoact2) in a wrapper that remaps the camera images or changes whatever is needed fro it to fit a downstream component (e.g. the gym). This would not change the interfaces to the individial gyms/environments.  In a similar fashion, why is mapping needed for a first party implementation? We could change Molmoact so that it just uses the correct names in the dict?

library/src/physicalai/policies/molmoact2/README.md
```diff
@@ -0,0 +1,374 @@
...
+                gym.close()
+```
+
+### Reported Results
+
+The following measurements are retained from the original integration report.
+Performance depends on hardware, precision, runtime versions, and benchmark
+configuration.
+
+NVIDIA A100, BF16:
```
kdijkstraintel: Do we want to report (runtime) stats here, how are these maintained? Also many of the other policies don't seem to contain README.md files (only the generic LeRobot policy has one)

library/src/physicalai/policies/molmoact2/policy.py
```diff
@@ -0,0 +1,1478 @@
...
+        self._apply_model_modifications()
+
+    def set_features(
+        self,
+        input_features: list[Feature],
+        output_features: list[Feature],
+        *,
+        copy_state_normalization: bool = False,
+        copy_action_normalization: bool = False,
+    ) -> None:
```
eugene123tw, line 509: Could you copy the example you have in README and it as example in docstring? The usecase is a bit unique, so I think it's good users know this. 

```py
Example:
        Initializing, setting features, and exporting a policy model:

        >>> import torch
        >>> policy = MolmoAct2(
        ...     pretrained_name_or_path="allenai/MolmoAct2-SO100_101",
        ...     norm_tag="so100_so101_molmoact2",
        ...     adapt_to_so101=True,
        ...     convert_pretrained_so101_stats=True,
        ... )
        >>> policy.set_features(
        ...     input_features=input_features,
        ...     output_features=output_features,
        ...     copy_state_normalization=True,
        ...     copy_action_normalization=True,
        ... )
        >>> policy = policy.to(dtype=torch.bfloat16).eval()
        >>> policy.export("exports/molmoact2-so101-torch", backend="torch")
```suggestion
    def set_features(
        self,
        input_features: list[Feature],
        output_features: list[Feature],
        *,
        copy_state_normalization: bool = False,
        copy_action_normalization: bool = False,
    ) -> None:
```
```

**Review by eugene123tw — DISMISSED** [review]

File: library/src/physicalai/data/observation.py [hunk 45]
```diff
@@ -473,14 +473,27 @@ class Feature:
...
-    q01: list[float] | float | None = None
-    q99: list[float] | float | None = None
+    """Statistics and optional selection mask for feature normalization.
+
+    Each statistic may be a scalar or a sequence matching the feature's
+    normalized dimensions.
+    """
+
+    #: Arithmetic mean used by mean/std normalization.
+    mean: list[float] | list[list[float]] | list[list[list[float]]] | float | None = None
```
samet-akcay: would it be an idea to use a type alias here to avoid repetition? something like 

```python
   type NormalizationValue = (
       float
       | list[float]
       | list[list[float]]
       | list[list[list[float]]]
       | None
   )
 ```

 Then:

 ```python
   @dataclass(frozen=True)
   class NormalizationParameters:
       """Statistics used to normalize a feature.

       Each statistic may be a scalar or a nested list matching the feature's
       normalized dimensions. ``None`` means that statistic is unavailable.
       """

       mean: NormalizationValue = None
       std: NormalizationValue = None
       min: NormalizationValue = None
       max: NormalizationValue = None
       q01: NormalizationValue = None
       q99: NormalizationValue = None
 ```

File: library/src/physicalai/policies/molmoact2/policy.py [hunk 46]
```diff
@@ -0,0 +1,1504 @@
...
+        pretrained_eager = pretrained_name_or_path is not None and norm_tag is not None
+        if user_eager or pretrained_eager:
+            self.initialize_model()
+
+    @classmethod
+    def from_config(  # noqa: PLR0913
+        cls,
+        config: MolmoAct2Config,
+        *,
+        preserve_pretrained_normalization_in_training: bool = False,
```
samet-akcay, line 269: so we have these args here because they are not in config? and it is because you want to keep the config to model, and don't include policy args there ?

library/src/physicalai/policies/molmoact2/policy.py
```diff
@@ -0,0 +1,1504 @@
...
+    ) -> MolmoAct2:
+        """Load a trained policy without resolving pretrained model weights.
+
+        The checkpoint config rebuilds the model before Lightning restores its state dict.
+        Tokenizer assets referenced by the config must remain available locally.
+
+        Returns:
+            The restored policy in the checkpoint's saved training mode.
+        """
+        kwargs["pretrained_name_or_path"] = None
```
samet-akcay, line 373: if we didn't save `pretrained_name_or_path` in hparams? would we still need to overwrite this method? 
Wondering if the default behaviour is correct for all other policies, or if your approach here can be generalised within the parent class?

library/src/physicalai/policies/molmoact2/policy.py
```diff
@@ -0,0 +1,1504 @@
...
+            raise TypeError(msg)
+        return self.model
+
+    def _require_config(self) -> MolmoAct2Config:
+        if self.config is None:
+            msg = "Policy config is not initialized"
+            raise RuntimeError(msg)
+        return self.config
+
+    def initialize_model(self) -> None:
```
samet-akcay, line 423: not for this pr, but can we utilize configure_model from lightning in the future?

library/src/physicalai/policies/molmoact2/README.md
samet-akcay: I like the readme file, but we should ideally report our own numbers here

library/pyproject.toml
```diff
@@ -130,6 +138,7 @@ all = [
     "physicalai-train[pi05]",
     "physicalai-train[groot]",
     "physicalai-train[smolvla]",
+    "physicalai-train[molmoact2]",
```
samet-akcay, line 141: not related to your pr, but would it be possible to have `policies` dependency group. you could also use it in ci as well like 

```diff
- uv sync --frozen --extra tests --extra pi05 --extra libero --extra smolvla --extra molmoact2 --extra rldx1 --extra cu128
+ uv sync --frozen --extra tests --extra policies --extra libero --extra cu128
```

library/src/physicalai/gyms/utils.py
alfieroddan: https://github.com/open-edge-platform/physical-ai-studio/pull/809/commits/e56a546ba10edb9aa7e7f28d8b368af622b47c7a

library/src/physicalai/policies/molmoact2/policy.py
alfieroddan: https://github.com/open-edge-platform/physical-ai-studio/pull/809/commits/9a8541df26c78b761405f23281f88c26624e3b53

library/src/physicalai/data/observation.py
alfieroddan: https://github.com/open-edge-platform/physical-ai-studio/pull/809/commits/2ca0445dd7146b148d2c52b81d5c73454bfda01d

library/src/physicalai/policies/molmoact2/policy.py
alfieroddan: Yeh these are training params / optimizations. My proposal here is that these don't describe the processors or the model so therefore don't need to be in the config.

For example we saved LORA is used in training. When we re-load the model - it's an un-necessary process in re-init.

library/src/physicalai/policies/molmoact2/policy.py
alfieroddan: Average user would only ever use pre-trained. Also, issue with this is that our tokenizer is custom and bundled with the pre-trained weights. So we'd have to downlod the tokenizer individually without the weights as a default if this was the case.

library/pyproject.toml
alfieroddan: Yeh agreed, can do with the integration tests re-vamp I think.

samet-akcay (dismissed):

samet-akcay (dismissed):

AlexanderBarabanov (dismissed):

samet-akcay (dismissed):

library/pyproject.toml
```diff
@@ -184,6 +194,9 @@ override-dependencies = [
     # Safe to override because executorch
     # production code does not import pytest at runtime.
     "pytest>=9.0.3",
+    # The pinned physicalai runtime currently caps Transformers below the
```
AlexanderBarabanov: Could you please check if we can remove it as physicalai hash has been updated in https://github.com/open-edge-platform/physical-ai-studio/pull/1163/changes

alfieroddan: Yep absolutely, resolved in https://github.com/open-edge-platform/physical-ai-studio/pull/809/commits/a546a17ceff7b17f323d832a27abf1ec329b957e

AlexanderBarabanov (dismissed):

samet-akcay (approved):

PR #818, opened 2026-07-20
feat(robot): integrate SharedRobot

maxxgx:
# Pull Request

## Description

- Use `SharedRobot` introduced in https://github.com/openvinotoolkit/physicalai/pull/189.
- Make `TeleoperateWorker` a `BaseThreadWorker`

## Breaking Changes

---


maxxgx:
@copilot resolve the merge conflicts in this pull request

application/backend/src/robots/catalog/so101.py
```diff
@@ -41,7 +42,16 @@ async def _build_so101_driver(robot: CatalogRobot[SO101RobotPayload], factory: C
         raise ResourceNotFoundError(ResourceType.ROBOT_CALIBRATION, robot.payload.serial_number)
 
     role = "follower" if robot.type == RobotType.SO101_FOLLOWER else "leader"
-    return SO101(port=port, calibration=_to_so101_calibration(calibration), role=role, unit="normalized")
+    return SharedRobot(
```
MarkRedeman: Could we keep building the `SO101` robot here, but have the factory wrap it with a `SharedRobot`?
Perhaps each robot can have a `to_dict()` type of method that can then be used to serialize it?

The reason I'm asking is that for the robot plugin design we will have third party code that will be responsible for building their own `Robot` implementation. Those plugins shouldn't care if it's going to be used as `SharedRobot`.

application/backend/src/robots/catalog/types.py
```diff
@@ -44,6 +44,7 @@ class PayloadContainer(Protocol[_PayloadT]):
 
 
 class CatalogRobot(PayloadContainer[_PayloadT], Protocol[_PayloadT]):
+    name: str
```
MarkRedeman: Why not use `type` as the name?

I'd prefer to avoid making changes to `CatalogRobot` and other plugin related code outside of my upcoming PRs for now.

application/backend/src/robots/catalog/widowxai.py
```diff
@@ -34,18 +41,36 @@
...
+        name=robot.name + "_left_arm",
+        robot_class=WidowXAI,
+        robot_kwargs={"ip": robot.payload.connection_string_left, "role": mode},
+    )
+    right_driver = SharedRobot(
+        name=robot.name + "_right_arm",
+        robot_class=WidowXAI,
+        robot_kwargs={"ip": robot.payload.connection_string_right, "role": mode},
+    )
+    return BimanualWidowXAI(left=left_driver, right=right_driver)  # pyrefly: ignore[bad-argument-type]
```
AlbertvanHouten: Why not update BimanualWidowXAI to work with SharedRobot so the typing can be fixed?

application/backend/src/robots/catalog/types.py
maxxgx: remove name

application/backend/src/robots/shared_robot_errors.py
```diff
@@ -0,0 +1,56 @@
...
+    RobotNameConflictError,
+    RobotProtocolMismatchError,
+    SharedRobotTransportError,
+)
+
+
+def translate_robot_error(
+    exc: BaseException,
+    *,
+    robot_name: str | None = None,
```
AlbertvanHouten, line 26: Including the robot name in all exceptions might be helpful for the user

maxxgx: We already get the name from the robot object directly. This works for `SharedRobot` as they always have a name

```
robot_name = getattr(self._robot, "name", None)
raise translate_robot_error(e, robot_name=robot_name) from e
```

pr-835-review-report.html
AlbertvanHouten, line 1: oops :)

application/backend/src/robots/catalog/so101.py
maxxgx: Solved with new component config from `physicalai`

application/backend/src/robots/catalog/widowxai.py
maxxgx: Done with new component config from physicalai

application/ui/rsbuild.config.ts
```diff
@@ -54,6 +54,7 @@ export default defineConfig({
         },
     },
     server: {
+        host: process.env.DEV_SERVER_HOST ?? 'localhost',
```
MarkRedeman: What's this change for?

application/plugin/src/physicalai_studio_plugin/catalog.py
```diff
@@ -36,8 +38,16 @@ class PayloadContainer(Protocol[_PayloadT]):
...
+    """Typed robot descriptor passed to catalog robot builders.
+
+    Studio builds the concrete model dynamically per registered robot type, so
+    builders must not narrow on a hand-written model class — match on
+    ``payload`` instead. These attributes are the stable surface builders can
+    rely on.
+    """
 
+    id: UUID
+    name: str
```
MarkRedeman: Do we need `name` if `id` is used by `shared_robot_name`?

application/backend/src/robots/catalog/so101.py
```diff
@@ -82,23 +91,32 @@ class SO101Robot(BaseRobot):
...
         raise ValueError(f"Could not resolve a serial port for {resource_key}")
 
-    calibration: SO101Calibration | None = None
-    if robot.payload.calibration is not None:
-        calibration = SO101Calibration(joints=robot.payload.calibration)
-
     role = "follower" if robot.type == "SO101_Follower" else "leader"
-    return SO101(port=port, calibration=calibration, role=role, unit="normalized")
+    if payload.calibration is None:
+        raise ValueError("SO101 calibration is required to build a SharedRobot driver")
```
MarkRedeman: Why is calibration strictly required?

At the moment this shouldn't be an issue, but I've been thinking that we should replace our calibration setup logic that uses lerobot's motor bus by the internal implementation of physicalai.
Making calibration required would complicate this.

On another note: if we want the calibration to not be none then we can also choose to remove the `None` option from the payload's type?

application/backend/src/robots/catalog/so101.py
```diff
@@ -82,23 +91,32 @@ class SO101Robot(BaseRobot):
...
+    # Construct the driver so its own validation runs here (clear, early errors),
+    # then hand SharedRobot only the recipe: the owner process builds the real
+    # driver, this process would immediately discard it.
```
MarkRedeman: Not sure if this comment is needed, since we'd then need to repeat it in every builder method

application/backend/src/robots/catalog/widowxai.py
```diff
@@ -51,14 +51,26 @@ class TrossenBimanualPayload(BaseModel):
...
 class TrossenSingleArmRobot(BaseRobot):
-    """Trossen WidowX AI follower or leader robot using an IP connection."""
+    """Trossen WidowX AI follower or leader robot using an IP connection.
+
+    Note:
+        Local convenience model for probe/identify helpers. Robots that reach a
+        builder come from ``RobotCatalogRegistry``, which generates a separate
+        model per type deriving straight from ``BaseRobot``, so an
+        ``isinstance`` check against this class in a builder never matches.
+        Match on the payload type instead.
```
MarkRedeman: Here and the comments below feel a bit out of date after moving the instantiation of `SharedRobot` to the factory class.

application/backend/src/workers/teleoperate_worker.py
```diff
@@ -108,25 +105,43 @@ def _align_feature_values(
...
+            if not self.is_alive():
+                if self.setup_error is not None:
+                    raise self.setup_error
+                raise RuntimeError("Teleoperation worker stopped before loading")
+            await asyncio.sleep(poll_interval)
 
     async def run_loop(self) -> None:
         try:
+            try:
+                self.follower = await self.robot_client_factory.build(self._follower_robot)
```
MarkRedeman: The double try here feels a bit excessive. Perhaps we can extract this part to a `asyn def _connect()` method?

application/ui/rsbuild.config.ts
maxxgx: It allows, for example with `0.0.0.0`, listen on all interfaces so I can access from another devices in the same network. Should be fine since we keep default to localhost?

application/backend/src/workers/teleoperate_worker.py
maxxgx: I've removed the nested try-except

application/backend/src/robots/catalog/so101.py
maxxgx: Removed

application/backend/src/robots/catalog/widowxai.py
maxxgx: Removed

application/backend/src/robots/catalog/so101.py
maxxgx: Made calibration optional and added `allow_uncalibrated'. Now uncalibrated SO101 can be created

application/plugin/src/physicalai_studio_plugin/catalog.py
maxxgx: You're right, `id` and `name` are not needed anymore

application/ui/rsbuild.config.ts
maxxgx: Removed :)

MarkRedeman (approved):

PR #819, opened 2026-07-21
feat: add remote trainer management UI with health checks and per-job interrupt handling

AlbertvanHouten:
## Description
Adds a full remote trainer management UI and backend to Physical AI Studio, letting users register direct remote training endpoints (as opposed to auto-discovered/relay trainers), monitor their health/devices/storage, and target training jobs at a specific remote trainer. This allows completely removing the environment variable based setup. 

**Backend**
- New `remote_trainers` table, repository, mapper, and Pydantic schemas (`RemoteTrainer`, `RemoteTrainerCreate`, `RemoteTrainerUpdate`, `RemoteTrainerHealth`) with an Alembic migration (`e4b2f1c8a907_add_remote_trainers`).
- `RemoteTrainerService` for CRUD plus `check_remote_trainer`, which probes a configured endpoint's `/health`, `/devices`, and (best-effort) `/storage` endpoints and reports a `healthy` / `degraded` / `unreachable` status with latency, available XPU/CUDA devices, and storage info.
- New `api/remote_trainers.py` router and `get_remote_trainer_service` dependency exposing the CRUD + health-check endpoints.
- `TrainingTarget` enum and validation added to `TrainJobPayload` (`schemas/job.py`) so a train job can be routed to a specific configured remote trainer.
- `training_backends/__init__.py` and `remote.py` updated to resolve the backend from the job's training target.
- `training_worker.py` reworked to support **per-job interrupt handling** for concurrent training jobs: jobs on distinct targets now start without waiting on each other, while jobs on the same target are serialized, and stopping one job no longer affects others running concurrently.
- `system_service.py` simplified by removing the now-unused cached remote backend probing logic (superseded by the direct remote trainer model).
- Misc: `job_repo.get_pending_jobs_by_type`, `job_service.get_pending_train_jobs`, scheduler wiring, `.env.example` and `settings.py` cleanup, README updates.

**Trainer microservice**

- Adds a `/storage` endpoint (`StorageInfo` schema + `get_storage_info`) so Studio can report available disk space for a remote trainer during health checks.
- Update logging to match style of backend logs
- Update cleanup logic to ensure no dataset/model artifacts are left behind for failed/cancelled jobs. 

**UI**
- New `remote-trainers-page` (list/add/edit/delete configured trainers, live health/device/storage status) with its own route under `routes/remote-servers`.
- `train-model-dialog.tsx` updated to let users pick a specific remote trainer as the training target, with polling for device availability.
- Project navigation and layout updated to surface the new remote servers section.
- New page is hidden behind a feature flag `PUBLIC_ENABLE_REMOTE_TRAINERS`. Can be set in the browser console via  `setFeatureFlag('remoteTrainers', true)`

## Breaking Changes
None expected for existing local/auto-discovered remote training flows. Requires running the new Alembic migration (`e4b2f1c8a907_add_remote_trainers`) to add the `remote_trainers` table.

---

## Screenshots
**Remote trainers page** 
image

**Add a remote trainer, connection type will be switchable in the future for SSH setups**
image

**Train model selection for remote trainer. Top right displays CPU/CUDA/XPU for currently selected trainer**
image

**Remote trained models have a purple pill badge while training to indicate where they run**

image


application/backend/src/api/remote_trainers.py
```diff
@@ -0,0 +1,55 @@
...
+@router.post("", status_code=status.HTTP_201_CREATED)
+async def create_remote_trainer(
+    config: RemoteTrainerCreate,
+    remote_trainer_service: Annotated[RemoteTrainerService, Depends(get_remote_trainer_service)],
+) -> RemoteTrainer:
+    """Persist a direct trainer endpoint."""
+    return await remote_trainer_service.create_remote_trainer(config)
+
+
+@router.post("/{remote_trainer_id}/health")
```
MarkRedeman: There doesn't seem to be any payload here, could this be a GET request?

application/backend/src/cli/run.py
```diff
@@ -124,19 +125,16 @@ def maybe_sync(cwd: Path, device: str, *extras: str, sync: bool | None) -> None:
 @_device_option
 @_sync_option
 def remote(host: str | None, port: int | None, device: str | None, sync: bool | None) -> None:
```
MarkRedeman: Do we still need this? If the main difference is that it uses `--extra cpu` then we can tell people to use that directly?

application/backend/src/repositories/base.py
```diff
@@ -87,15 +95,15 @@ async def save(self, item: ModelType) -> ModelType:
     async def update(self, item: ModelType, partial_update: dict) -> ModelType:
         # note: model_copy does not validate the model, so we need to validate explicitly
         to_update = item.model_copy(update=partial_update, deep=True)
-        to_update = cast("ModelType", item.__class__.model_validate(to_update.model_dump()))
+        to_update = cast("ModelType", item.__class__.model_validate(to_update.__dict__))
```
MarkRedeman: What's the motivation for this change, would this not introduce a conflict with types that include json like our robots' payload?

application/backend/src/repositories/base.py
```diff
@@ -72,8 +72,16 @@ async def get_one(
             return self.from_schema(first_result)
         return None
 
-    async def get_all(self, extra_filters: dict | None = None, expressions: list[Any] | None = None) -> list[ModelType]:
+    async def get_all(
+        self,
+        extra_filters: dict | None = None,
+        expressions: list[Any] | None = None,
+        order_by: Any | None = None,
```
MarkRedeman, line 79: Perhaps we can add a default `order_by`, for instance `created_at`? (though might need to watch out to guarantee that this is available)

At the moment in most of our queries we ignore the order by and get a nondeterministic ordering.

application/backend/src/services/job_service.py
```diff
@@ -67,6 +83,10 @@ async def submit_train_job(self, payload: TrainJobPayload) -> Job:
     async def get_pending_train_job(self) -> Job | None:
         return await self.repo.get_pending_job_by_type(JobType.TRAINING)
 
+    async def get_pending_train_jobs(self) -> list[Job]:
+        """Return pending training jobs in submission order."""
+        return await self.repo.get_pending_jobs_by_type(JobType.TRAINING)
+
```
MarkRedeman, line 89: Duplicated / can we simplify this? Perhaps replace `get_pending_train_job`?

application/backend/src/services/remote_trainer_service.py
```diff
@@ -0,0 +1,127 @@
...
+        return remote_trainer
+
+    async def check_remote_trainer(self, remote_trainer_id: UUID) -> RemoteTrainerHealth:
+        """Check a configured trainer's liveness and available compute devices."""
+        remote_trainer = await self.get_remote_trainer(remote_trainer_id)
+        checked_at = datetime.now(UTC)
+        started = perf_counter()
+        base_url = str(remote_trainer.url).rstrip("/")
+        timeout = httpx.Timeout(_HEALTH_CHECK_TIMEOUT_S)
+        status: Literal["healthy", "degraded", "unreachable"] = "healthy"
```
MarkRedeman: Perhaps extract this literal to `HealthStatus = Literal["healthy", "degraded", "unreachable"]`

application/backend/src/repositories/base.py
AlbertvanHouten: There was a pydantic serialization warning that is fixed by changing this to a dict. As far as I could tell there is no issue with the robots payload.

application/backend/src/repositories/base.py
AlbertvanHouten: Yeah, lets do it as part of a separate PR.

application/backend/src/cli/run.py
AlbertvanHouten: No, can be removed indeed.

application/backend/src/api/remote_trainers.py
AlbertvanHouten: Good catch, will change to GET

application/backend/src/services/job_service.py
AlbertvanHouten: If we use the list method for single job fetching we will need to add additional None checks so either way I think we end up with about the same code.

application/backend/src/services/remote_trainer_service.py
AlbertvanHouten: done

application/ui/src/features/remote-trainers/remote-trainers-page.tsx
```diff
@@ -0,0 +1,577 @@
...
+    const storageState = isChecking
+        ? 'neutral'
+        : health === undefined
+          ? 'neutral'
+          : health.storage
+            ? 'positive'
+            : health.status === 'unreachable'
+              ? 'neutral'
+              : 'yellow';
+    const lastChecked = health ? new Date(health.checked_at).toLocaleString() : 'Not checked';
```
MarkRedeman, line 283: This part is a bit too complicated. I'd extract a `getCapabilityState` and `getStorageState` functions that return these.

application/ui/src/features/remote-trainers/remote-trainers-page.tsx
```diff
@@ -0,0 +1,577 @@
...
+
+    return (
+        <View UNSAFE_className={classes.detailPane}>
```
MarkRedeman, line 286: Perhaps extract to a `RemoteTrainerHeading` component, along with a `HealthAndCapabilityDetail` and `ConnectionDetail` component.
This could help reduce the amount of state kept in the current component and reduce some of the ternary logic you're using here.

application/ui/src/features/remote-trainers/remote-trainers-page.tsx
```diff
@@ -0,0 +1,577 @@
...
+    const [selectedRemoteTrainerId, setSelectedRemoteTrainerId] = useState<string>();
+    const [formRemoteTrainer, setFormRemoteTrainer] = useState<SchemaRemoteTrainer | null>();
+    const [deleteRemoteTrainer, setDeleteRemoteTrainer] = useState<SchemaRemoteTrainer>();
+    const [healthByTrainerId, setHealthByTrainerId] = useState<Record<string, SchemaRemoteTrainerHealth>>({});
+    const [checkingTrainerIds, setCheckingTrainerIds] = useState<Set<string>>(new Set());
+    const checkedTrainerUrls = useRef<Map<string, string>>(new Map());
+    const checkGenerationByTrainerId = useRef<Map<string, number>>(new Map());
+    const selectedRemoteTrainer = remoteTrainers.find((remoteTrainer) => remoteTrainer.id === selectedRemoteTrainerId);
+
+    const checkRemoteTrainer = useCallback(
```
MarkRedeman: Extract to a hook and leverare tanstack/query to handle all race conditions including error handling.
Use computed variables instead of extra `useState` if the variables are derivable from a query state.

application/ui/src/features/remote-trainers/remote-trainers-page.tsx
```diff
@@ -0,0 +1,577 @@
...
+        async (remoteTrainerId: string) => {
+            const generation = (checkGenerationByTrainerId.current.get(remoteTrainerId) ?? 0) + 1;
+            checkGenerationByTrainerId.current.set(remoteTrainerId, generation);
+            setCheckingTrainerIds((ids) => new Set(ids).add(remoteTrainerId));
+            try {
+                const health = await healthCheckMutation.mutateAsync({
```
MarkRedeman: Probably changing the health check to be a GET could simplify some of this logic as you could make this a `useSuspenseQuery`.

application/ui/src/features/remote-trainers/remote-trainers-page.tsx
```diff
@@ -0,0 +1,577 @@
...
+            }
+        },
+        [healthCheckMutation]
+    );
+
+    useEffect(() => {
+        if (selectedRemoteTrainer === undefined) {
+            setSelectedRemoteTrainerId(remoteTrainers[0]?.id);
+        }
+    }, [remoteTrainers, selectedRemoteTrainer]);
```
MarkRedeman: Avoid doing this: use `remoteTrainers[0]?.id` as the default variable for the `useState`. To make sure `remoteTrainers` is available when initializing this state we can use `$api.useSuspenseQuery` instead of `$api.useQuery`.
Add a `<Suspense>` boundary in the parent that renders this component.

application/ui/src/features/remote-trainers/remote-trainers-page.tsx
```diff
@@ -0,0 +1,577 @@
...
+        checkedTrainerUrls.current = new Map(
+            [...checkedTrainerUrls.current].filter(([remoteTrainerId]) => configuredTrainerIds.has(remoteTrainerId))
+        );
+        remoteTrainers.forEach((remoteTrainer) => {
+            if (checkedTrainerUrls.current.get(remoteTrainer.id) !== remoteTrainer.url) {
+                checkedTrainerUrls.current.set(remoteTrainer.id, remoteTrainer.url);
+                void checkRemoteTrainer(remoteTrainer.id);
+            }
+        });
+    }, [checkRemoteTrainer, remoteTrainers]);
```
MarkRedeman: This could probably be replaced by a `$api.useQuery` (or `$api.useSuspenseQuery`) and moving its state to the component that uses it.

application/ui/src/features/remote-trainers/remote-trainers-page.tsx
```diff
@@ -0,0 +1,577 @@
...
+                                        ))}
+                                        <Text UNSAFE_className={classes.cardMetaText}>
+                                            {healthByTrainerId[remoteTrainer.id]?.devices?.[0]?.name ??
+                                                (checkingTrainerIds.has(remoteTrainer.id)
+                                                    ? 'Checking capability…'
+                                                    : 'Capability not reported')}
+                                        </Text>
+                                    </Flex>
+                                </Flex>
+                            </ActionButton>
```
MarkRedeman: Extract to a `SelectTrainerButton`. Consider changing this to a `NavLink` instead of a button and using react router's routing.

application/ui/src/features/remote-trainers/remote-trainers-page.tsx
```diff
@@ -0,0 +1,577 @@
...
+            </div>
+            <DialogContainer onDismiss={() => setFormRemoteTrainer(undefined)}>
+                {formRemoteTrainer !== undefined && (
+                    <RemoteTrainerForm
+                        remoteTrainer={formRemoteTrainer ?? undefined}
+                        close={() => setFormRemoteTrainer(undefined)}
+                    />
+                )}
+            </DialogContainer>
+            <DialogContainer onDismiss={() => setDeleteRemoteTrainer(undefined)}>
```
MarkRedeman: These containers could be combined. Instead of having `setFormRemoteTrainer` and `setDeleteRemoteTrainer` use `[action, setAction] = useState<'delete' | 'create'>(undefined)`.
The selected trainer id can be determined by `useParams`.

application/ui/src/routes/models/train-model-dialog.tsx
```diff
@@ -152,63 +162,146 @@ const PolicySelection = ({ selectedPolicy, onSelectionChange, isDisabled, traini
...
+/**
+ * Checks and polls a selected remote trainer's health.
+ *
+ * The trainer is re-checked whenever `remoteTrainerId` changes, and polled
+ * every `REMOTE_UNAVAILABLE_POLL_MS` while it is unreachable so the UI
+ * recovers automatically once the trainer comes back. `checkHealth` is also
+ * exposed so callers can force a final check (e.g. right before submitting).
+ */
+const useRemoteTrainerHealth = (remoteTrainerId: string | null) => {
+    const healthCheckMutation = $api.useMutation('post', '/api/remote-trainers/{remote_trainer_id}/health');
```
MarkRedeman: Consider using a `$api.useQuery` instead, this enables polling and depends on the `remote_training_id`. I'd expect this whole hook can be replaced by a few lines of using `$api.useQuery`.

MarkRedeman (approved):

Issue #717, opened 2026-06-17
[Epic]: integrate studio with inference framework


sovrasov:
### Summary

Now, users have to manually create inference scripts/configs when they want to move inference from studio to deployment. Studio allows downloading individual exported policies, but with no extra information on how to actually organize inference.

This epic is set up to discuss, plan and implement required changes.

### Motivation

Once implemented, the feature would allow a smooth transfer from studio to phusicalai inference framework.

### Scope

- UI for downloading deployable model package
- package structure
- inference pipeline configuration

### Tasks

TBD

### Acceptance Criteria

- Deployable package is ready to be executed using one cli command

sovrasov:
Potential deployable package structure:
- `README.md`: model card + inference instructions
- `runtime_config.yaml`: a configuration for `PolicyRuntime`. See example [here](https://github.com/openvinotoolkit/physicalai/blob/main/examples/runtime/runtime.yaml)
- exported model folder
- `calibration.yaml`: robot calibration (if needed)

maxxgx:
This is addressed with the integration of the runtime into the application: https://github.com/open-edge-platform/physical-ai-studio/pull/992 and https://github.com/open-edge-platform/physical-ai-studio/pull/936

PR #820, opened 2026-07-21
refactor: move SO101 calibration into payload, remove active_calibration_id from schema

MarkRedeman:
Having dedicated SQL tables and services only for the SO101 calibration resulted in quite a bit of unnecessary complexity where we'd need to inject things like the `RobotCalibrationService` into the robot client factory and other components only so that we could build SO101 robots.

This became even more clear while working on the robot plugin design where the SO101 robot simply felt awkward vs other robots.
To fix this we simply move the full calibration json into the SO101's payload. This results in a bit of a bigger payload that could result into slower API performance however it's still manageable. Moreover since it's unlikely that people will add many (100+) robots it likely won't have any real effect.

Another nice advantage is that for tickets like #717 we can also simply load the calibration json from the robot's payload and add this to the model's readme / deployment examples.


application/backend/src/alembic/versions/c9d8e7f6a5b4_backfill_so101_calibration_into_payload.py
MarkRedeman, line 1: Note this migration only moves the calibration into the robot payload, actually removing the older calibration tables will be done in a follow up PR so that while testing this we won't immediately lose that data.

application/backend/src/schemas/robot.py
```diff
@@ -27,6 +27,14 @@ class RobotType(StrEnum):
 # ============================================================================
 
 
+class SO101CalibrationValue(BaseModel):
+    id: int
```
AlbertvanHouten: is this id used anywhere?

MarkRedeman: It references the motor id (so 1 to 6 for so101), I think `motor_id` might be a better name here, but it would be better to keep in sync with lerobot's format so that people can easily switch between lerobot and physicalai

application/backend/src/schemas/robot.py
```diff
@@ -35,6 +43,10 @@ class SO101RobotPayload(BaseModel):
         description="Serial port path; leave empty to auto-discover via serial_number",
     )
     serial_number: str = Field(default="", description="USB serial number of the robot (when available)")
+    calibration: dict[str, SO101CalibrationValue] | None = Field(
```
MarkRedeman: Would like to check if we can't use `SO101Calibration` from the physicalai library directly, saving us needing to define `SO101CalibrationValue`

MarkRedeman:
Few notes for reviewing:
1. On the backend side we mostly ended up removing code as we don't need all the database logic anymore
2. On the frontend side a bit of extra logic was needed to make sure that when editing a SO101 robot we don't overwrite the callibration values accidentally
3. The manual calibration setup is still working and writes the calibration file to the control board at the end of the setup (writing happens at `_handle_stop_recording()`)

AlbertvanHouten (dismissed):

AlbertvanHouten (dismissed):

AlbertvanHouten (approved):

PR #822, opened 2026-07-21
fix: torchvision version corresponds to torch version

sovrasov:
# Pull Request

## Description

torchvision upper bound was not bumped together with torch, which caused incompatibility when installing torch 2.11 or newer


AlbertvanHouten:
Please also update https://github.com/open-edge-platform/physical-ai-studio/blob/main/application/trainer/pyproject.toml

AlbertvanHouten (approved):

PR #823, opened 2026-07-21
feat: add trainer dockerfile and enhance health endpoint

AlbertvanHouten:
## Description
This PR adds a docker image to deploy the remote trainer. It includes both an image for CUDA and for XPU. It also adds a github action workflow to verify the image build and works correctly and publishes it to ghcr. Additionally, the health endpoint for the trainer is updated with information about the trainer image if available.


application/trainer/src/trainer/schemas.py
```diff
@@ -47,6 +47,17 @@ class DeviceInfo(BaseModel):
     index: int | None = Field(default=None, description="Device index among those of the same type (null for CPU)")
 
 
+class HealthInfo(BaseModel):
```
MarkRedeman: Could this extend the `BaseSettings` from `pydantic_settings`? Then you can construct it from environment variables so that you don't need to do the `os.environ.get` stuff you're doing in the health check endpoint?

application/trainer/README.md
```diff
@@ -77,6 +78,145 @@ To run the ASGI app module directly:
...
+  "status": "healthy",
+  "protocol_version": 1,
+  "device_type": "cuda",
+  "build_revision": "<git-sha>",
+  "build_date": "<RFC 3339 timestamp>",
+  "application_version": "<version>"
+}
+```
+
+### Run a container manually
```
MarkRedeman, line 114: Perhaps we can also update the docker compose file i n`application/docker` so that it references the trainer images, or add a new `docker-compose.trainer.yaml` configuration that can be used standalone?

application/trainer/README.md
```diff
@@ -77,6 +78,145 @@ To run the ASGI app module directly:
...
+```bash
+docker run --rm --gpus all \
+  nvidia/cuda:12.8.0-base-ubuntu24.04 \
+  nvidia-smi
+```
+
+Start the trainer with GPU access and a loopback-only port binding:
+
+```bash
+docker run --rm \
```
MarkRedeman, line 175: In our docker compose we also use `ipc: host` as an alternative to setting say `shm_size: 16g`, this should help in making sure the trainers get access to as much compute as possible.

@copilot can you make any suggestions on how to add this info to our readme? See https://github.com/open-edge-platform/physical-ai-studio/blob/main/application/docker/docker-compose.yaml

application/trainer/README.md
AlbertvanHouten: Most of the time users shouldn't bother to run these images manually as they should be provisioned automatically through the SSH config in the UI but I'll add a separate `docker-compose.trainer.yaml` for completeness

application/trainer/src/trainer/schemas.py
AlbertvanHouten: Good point, fixed

MarkRedeman (approved):

AlexanderBarabanov (approved):

PR #832, opened 2026-07-22
fix: Switch docker build workflow to self-hosted runner

AlbertvanHouten:
## Description

Switch the docker build workflow step to the self-hosted runner to prevent issues with storage.


ivanzati (approved):
Thank you, Albert!

PR #835, opened 2026-07-23
feat(robots): allow registering third party robots using a new plugin system

MarkRedeman:
This PR moves robot catalog integration toward the new plugin architecture and updates both backend and UI to consume the catalog in a more dynamic way.

## Overview

This PR moves robot catalog integration toward the new plugin architecture and updates both backend and UI to consume the catalog in a more dynamic way.

## What changed

1. **RobotProbe integration**
Each plugin defines its own `RobotProbe`, which is responsible for:
- discovering (serial) devices: used by the UI to let the user pick their device
- identify a robot: convenience endpoint that moves the gripper of a robot
- online check: used in the robot list to check if the backend can connect to the robot

2. **RobotAsset integration**
Each plugin is responsible for configuring its `RobotAsset`. This determines the URDF that the UI shows and how to do the feature mapping to a joint mapping in this URDF.

3. **`robot_type` handling**
Before `RobotType` was a static enum. Now it's been changed to be a dynamic enum based on the loaded plugins. I've tried to keep things type stable where possible.

4. **`physicalai-studio-plugin` package adoption**
The `RobotCatalogDefinition` is now exposed as a new package that third party plugins can use to implement their own robots.

5. **Backend/UI plugin architecture usage**
- UI robot typing now derives from OpenAPI operation types instead of hard-coded robot unions.
- Robot form discover/identify flows were refactored to use the active catalog type and shared catalog hooks.
- Robot type picker descriptions are no longer hard-coded in UI and now rely on catalog-provided display values.

## Validation

- Added/updated backend API tests for:
  - invalid probe payload behavior
  - unknown robot type behavior across catalog routes
- Removed legacy schema test tied to the old bimanual plugin model shape.

## TODO:

1. Determine a publishing workflow for the new studio plugin package.
2. Look into changing `register` to `register_robot` in the registry, so that later we can extend its capabilities with `register_camera`, `register_teleoperator` etc?
3. Improve UI integration. Consider using a architecture like [react-jsonschema-form](https://github.com/rjsf-team/react-jsonschema-form) to generate UI form fields for plugins, instead of our hardcoded forms.


application/backend/src/api/robot_catalog.py
```diff
@@ -42,23 +71,81 @@ async def list_robot_catalog(catalog_service: RobotCatalogServiceDep) -> list[Ro
...
+            resource_type=ResourceType.ROBOT,
+            resource_id=definition.type,
+            message=f"Robot type {definition.type} does not support identification.",
+        )
+    await definition.probe.identify(payload, robot_manager, joint)
+
+
+@router.post("/{robot_type}/is-online")
+async def check_robot_online(
+    robot_type: str,  # noqa: ARG001 - consumed by RobotDefinitionDep via _get_robot_definition
```
maxxgx: Are you sure `robot_type` is needed? I think `RobotDefinitionDep` should already include it.

application/backend/src/repositories/mappers/project_robot_mapper.py
```diff
@@ -1,6 +1,19 @@
...
+from schemas.robot import Robot, RobotAdapter
+
+_TYPE_NORMALIZATION: dict[str, str] = {
+    "SO101_FOLLOWER": "SO101_Follower",
+    "SO101_LEADER": "SO101_Leader",
+    "TROSSEN_WIDOWXAI_FOLLOWER": "Trossen_WidowXAI_Follower",
+    "TROSSEN_WIDOWXAI_LEADER": "Trossen_WidowXAI_Leader",
+    "TROSSEN_BIMANUAL_WIDOWXAI_FOLLOWER": "Trossen_Bimanual_WidowXAI_Follower",
+    "TROSSEN_BIMANUAL_WIDOWXAI_LEADER": "Trossen_Bimanual_WidowXAI_Leader",
+}
```
maxxgx, line 14: Is this conversion needed for backward compatibility? Why not a migration script?

application/plugin/README.md
```diff
@@ -0,0 +1,352 @@
...
+[project.entry-points."physicalai.studio.catalog_plugins"]
+my-robot = "physicalai_my_robot_plugin.studio_catalog:register_physicalai_studio_plugin"
+```
+
+The callable must accept a single argument — the registry — and call `registry.register(definition)` for each robot type:
+
+```python
+def register_physicalai_studio_plugin(registry: Any) -> None:
+    for definition in _definitions():
+        registry.register(definition)
```
maxxgx: It seems you renamed it to `register_robot` in the implementation

application/backend/src/robots/robot_service.py
```diff
@@ -20,13 +21,20 @@ async def get_robot_list(project_id: UUID) -> list[Robot]:
...
     async def find_online_robots(project_id: UUID) -> list[RobotWithConnectionState]:
         robots = await RobotService.get_robot_list(project_id)
-        discovery = DiscoveryManager()
-        await discovery.refresh_hardware_ports()
+
+        # Single serial port scan shared across all probes
+        manager = RobotConnectionManager()
+        await manager.find_robots()
+
+        registry = RobotCatalogRegistry()
```
maxxgx, line 29: Is it necessary to re-register all robots every time? Maybe keep only one lifecycle catalog for the entire app

application/plugin/pyproject.toml
```diff
@@ -0,0 +1,67 @@
+[project]
+name = "physicalai-studio-plugin"
+version = "0.1.0"
+description = "Plugin SDK for Physical AI Studio — types and protocols for building robot catalog plugins"
+requires-python = ">=3.12"
+dependencies = [
+    "pydantic>=2.12",
+    "physicalai",
```
maxxgx, line 8: Do we depend on `physicalai` only for the typing?
Maybe there a way to expose the interfaces without requiring to install all base dependencies, e.g. `physicalai[interfaces]`?

application/backend/src/robots/robot_client_factory.py
```diff
@@ -39,15 +41,23 @@ async def build(self, robot: Robot) -> RobotClient:
...
+        port = self._resolve_port(self.robot_manager.robots, port_info)
+        if port is not None:
+            return port
+
+        await self.robot_manager.find_robots()
+        return self._resolve_port(self.robot_manager.robots, port_info)
+
+    @staticmethod
+    def _resolve_port(discovered: list[SerialPortInfo], target: SerialPortInfo) -> str | None:
+        if target.serial_number is not None and target.serial_number != "":
```
maxxgx:
```suggestion
        if target.serial_number:
```

This is more pythonic :)

application/backend/src/robots/catalog/registry.py
```diff
@@ -1,24 +1,138 @@
...
+
+    def make_robot_type(self) -> Any:
+        models = self.get_robot_types()
+        return Annotated[_build_union(models), Field(discriminator="type")]
+
+    def _load_external_plugins(self) -> None:
+        discovered_entry_points = list(entry_points(group=CATALOG_PLUGIN_ENTRYPOINT_GROUP))
+
+        for discovered_entry_point in discovered_entry_points:
+            register_plugin = discovered_entry_point.load()
```
maxxgx, line 126: Does loading involve IO (e.g. over serial)? If so, maybe we can add retry/timeout to make it more robust

application/backend/src/robots/catalog/registry.py
```diff
@@ -1,24 +1,138 @@
...
+            asset = RobotAsset(
+                urdf_relative_path=relative_path,
+                packages={package: Path(path) for package, path in package_paths.items()},
+                joint_map=getattr(definition, "joint_map", {}),
+                root_resolver=getattr(definition, "asset_root_resolver", None),
+            )
+
+        robot_payload = getattr(definition, "robot_payload", None)
+        if robot_payload is None:
+            robot_payload = definition.robot_model.model_fields["payload"].annotation
```
maxxgx: Why both `robot_payload` and `payload` fields?

application/plugin/src/physicalai_studio_plugin/factory.py
```diff
@@ -0,0 +1,19 @@
...
+
+if TYPE_CHECKING:
+    from .schemas import SerialPortInfo
+
+
+class CatalogRobotFactory(Protocol):
+    """Factory protocol provided by Studio to robot builders.
+
+    Plugin ``robot_builder`` callables receive an instance of this protocol
+    so they can resolve serial ports and load calibrations at build time.
```
maxxgx: Calibration is not loaded anymore but part of the payload, right?

maxxgx (commented):
Thanks, looks good overall

application/backend/src/utils/serial_robot_tools.py
```diff
@@ -29,7 +18,6 @@ def from_port(port: ListPortInfo) -> SerialPortInfo | None:
     # Also accept virtual/PTY ports (pid is None) like socat-created devices
     if port.pid is not None and port.pid not in {21971, 29987}:
         logger.debug("Found usb port with unexpected PID, {device}: {pid}", device=port.device, pid=port.pid)
-        return None
 
     return SerialPortInfo(connection_string=port.device, serial_number=serial_number or None)
```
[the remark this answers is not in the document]
MarkRedeman: This is expected as we can't have a strong guarantee that all robots we will support use these same pids.

application/backend/src/api/robot_catalog.py
MarkRedeman: It's not needed, but I either had to add this `noqa: ARG001` comment to exclude the unused arg from our linter or add another noqa for the missing arg.
OpenCode decided to add the `noqa: ARG001`.. :smile:

application/backend/src/repositories/mappers/project_robot_mapper.py
MarkRedeman: Yep, I'm adding a small comment with an explainer.

I don't mind adding a migration script, but wanted to avoid it during the development of this branch as I was switching back and forth a lot between this branch and develop.

I'm thinking of creating a separate PR with a migration script and removal of this code after this gets merged, is that ok?

application/backend/src/robots/robot_service.py
MarkRedeman: Yes I'd like to apply proper dependency injection here. However some of our services are static and make it hard to do DI.
The registration itself is quite cheap (especially while we have a limited set of plugins) but I agree it feels wasteful to do.

Is it ok if I make this improvement in a follow up PR?

---

One other thing that I've got on mind is that perhaps later the registry can read the registered plugins from our database. This would be useful if later on we allow users to install plugins via the UI. Though that requires more work.

application/plugin/pyproject.toml
MarkRedeman: Yes no runtime dependencies. Wanted to mention `physicalai[interfaces]` to you as well.

application/backend/src/robots/catalog/registry.py
MarkRedeman: No IO, but we do do some reflection to generate the robot types etc.

In theory third party plugins could do IO work though, so might be good to add extra guards here.
Is it ok if I do that as a follow up?

application/backend/src/robots/catalog/registry.py
MarkRedeman: Leftover from an earlier experiment where I had the plugins expose a `robot_model`, instead of `robot_payload`. Will remove this.

application/plugin/pyproject.toml
maxxgx: Will keep `physicalai[interfaces]` in mind :)

maxxgx (approved):

AlexanderBarabanov (approved):

PR #837, opened 2026-07-23
fix(gyms): fix LiberoGym image rotation and init-state index

Daankrol:
## Summary

Two bugs remained in `LiberoGym` after #826 fixed the reset order.

### Bug 1 — `_init_state_id` never advanced between episodes

Every episode in a multi-episode run replayed the same init state (`0`). This eliminated diversity across episodes and made aggregate success rates unreliable.

**Fix:** increment `_init_state_id` after each reset with mod wrapping against `len(self._init_states)`.

### Bug 2 — 180° image flip applied to policy observations

`_format_raw_obs()` rotated every camera image 180° before passing it to the policy. Pretrained checkpoints (`HuggingFaceVLA/smolvla_libero`, `lerobot/pi05_libero_finetuned_v044`) are trained via LeRobot's `LiberoEnv`, which does **not** rotate images in its observation path — the flip exists only in `render()` for human-viewable display.

Feeding the policy upside-down images causes near-zero success rates even with the correct scene layout.

**Fix:** remove the flip from `_format_raw_obs()`; apply it only in `render()`, matching `lerobot/envs/libero.py`.

## Files changed

- `library/src/physicalai/gyms/libero.py`

### DIFF
--- a/library/src/physicalai/gyms/libero.py
+++ b/library/src/physicalai/gyms/libero.py
@@ -55,7 +55,9 @@ Example:
 
 Note:
     - Requires optional dependencies: `uv pip install hf-libero`
-    - Images are automatically rotated 180° to match LeRobot conventions
+    - Images fed to policies are in raw robosuite orientation (no rotation), matching
+      the LeRobot training convention. The 180° flip is applied only in render()
+      for human-viewable output.
     - State space: 8-dim (3 pos + 3 orientation + 2 gripper)
     - Action space: 7-dim
     - See TASK_SUITE_MAX_STEPS for episode limits per suite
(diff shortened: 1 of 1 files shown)


library/src/physicalai/gyms/libero.py
```diff
@@ -365,9 +367,11 @@ def reset(
 
         raw_obs = self.env.reset()
 
-        # Apply init state if available
+        # Apply init state if available, then advance the index so successive
+        # episodes use distinct starting configurations.
         if self.init_states and self._init_states is not None:
-            raw_obs = self.env.set_init_state(self._init_states[self._init_state_id])
+            raw_obs = self.env.set_init_state(self._init_states[self._init_state_id % len(self._init_states)])
```
daniil-lyakhov, line 373: Is it really possible that `self._init_state_id == len(self._init_states)`?
I think this guard could possibly hide some different logic bug then

Daankrol: _init_state_id is incremented every episode and is not bounded by len(self._init_states). If the benchmark runs more episodes than there are init states (which is the common case e.g., 20 episodes, 10 init states), the index would go out of range without the modulo. The % len wrap is intentional: it cycles through the available init statesa, ensuring all init states are reused evenly across a longer run. This mirrors how LeRobot's LiberoEnv handles multi-episode evaluation.

PR #848, opened 2026-07-24
fix: lerobot api usage in video repair function

sovrasov:
## Description

API usage is restored, the function can be called without an explicit failure


alfieroddan (approved):
Thanks!

PR #856, opened 2026-07-27
refactor: Enhance dependency injection

AlbertvanHouten:
## Description

This pull request introduces significant improvements to dependency injection and service management in the backend, with a focus on making services request-scoped and database session-aware. 

### Dependency Injection and Service Refactoring

* Refactored service dependencies in `dependencies.py` to use FastAPI's `Depends` with explicit, request-scoped database sessions (`AsyncSession`). This ensures that services such as `ProjectService`, `RobotService`, `ModelService`, and others are instantiated per request with the correct session, improving reliability and testability. 
* Introduced new dependency types (e.g., `SettingsDep`, `AsyncSessionDep`, `RobotClientFactoryDep`) and replaced direct instantiation of services with dependency-injected versions throughout the codebase. 

### Database Migration and Error Handling
* Added the `ALLOW_UNKNOWN_DB_REVISION` environment variable to `.env.example` and implemented logic in the migration manager to allow startup if the database references an unknown Alembic revision, providing a recovery path for certain migration issues (mainly useful for developers that switch branches which might result migration changes).

### DIFF
--- a/application/backend/src/api/dependencies.py
+++ b/application/backend/src/api/dependencies.py
@@ -1,5 +1,5 @@
 from functools import lru_cache
-from typing import Annotated
+from typing import Annotated, cast
 from uuid import UUID
 
 from fastapi import Depends, status
@@ -7,6 +7,7 @@ from fastapi.exceptions import HTTPException
 from fastapi.requests import HTTPConnection
 
 from core.scheduler import Scheduler
+from robots.robot_client_factory import RobotClientFactory
 from services import (
     DatasetDownloadService,
     DatasetService,
@@ -25,9 +26,11 @@ from services.job_service import JobService
 from services.log_service import LogService
 from services.robot_catalog_service import RobotCatalogService
 from services.system_service import SystemService
-from settings import get_settings
+from settings import Settings, get_settings
 from utils.serial_robot_tools import RobotConnectionManager
 from workers.model_worker_registry import ModelWorkerRegistry
+
+SettingsDep = Annotated[Settings, Depends(get_settings)]
 
 
 def is_valid_uuid(identifier: str) -> bool:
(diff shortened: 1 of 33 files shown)


application/backend/src/services/dataset_service.py
```diff
@@ -2,53 +2,43 @@
...
-            if dataset is None:
-                raise ResourceNotFoundError(ResourceType.DATASET, str(dataset_id))
-
-            await repo.delete_by_id(dataset_id)
-
-            if remove_files:
-                shutil.rmtree(Path(dataset.path).expanduser())
+    def __init__(self, session: AsyncSession) -> None:
+        self.session = session
+        self.repo = DatasetRepository(session)
```
MarkRedeman, line 15: Not directly needed now, but we could also choose to inject the actual `DatasetRepository` (so that for testing we could use an `InMemoryDatasetRepository`)

That would probably require a bunch more work though.

application/backend/src/api/dependencies.py
```diff
@@ -68,12 +82,24 @@ def get_robot_manager_service(request: HTTPConnection) -> RobotConnectionManager
     if robot_manager is None:
         raise RuntimeError("Robot manager not initialized")
 
-    return robot_manager
+    return cast("RobotConnectionManager", robot_manager)
```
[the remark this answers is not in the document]
AlbertvanHouten: This change conflicts with ruff rule TC006

application/backend/src/services/dataset_service.py
AlbertvanHouten: Yeah we could indeed go even further and use dependency injection for all repositories too but for now our API is quite testable by just swapping out a service. Maybe as a follow up in the future we could use DI for the repositories.

application/backend/src/workers/training_worker.py
```diff
@@ -45,12 +46,12 @@ def __init__(self, stop_event: EventClass, interrupt_event: EventClass, event_qu
         self.interrupt_event = interrupt_event
 
     async def run_loop(self) -> None:
-        job_service = JobService()
         logger.info("Training Worker is running")
         while not self.should_stop():
             settings = get_settings()
 
-            job = await job_service.get_pending_train_job()
+            async with get_async_db_session_ctx() as session:
```
MarkRedeman, line 53: I guess the reason we can't use DI here for the job service is that we don't want to lock the sql database?

MarkRedeman (approved):

PR #858, opened 2026-07-27
feat(pi05): auto-detect adaRMSNorm conditioning from checkpoint

pwolnows:
## Problem

`OpenVINO/pi05-libero-fp16-ov` on HuggingFace was exported from an **older codebase that predates bf16 computation and adaRMS layer support** in `PaliGemmaWithExpertModel`. Comparing it against the native `lerobot/pi05_libero_finetuned_v044` checkpoint (which runs in bfloat16 with adaRMS conditioning) produces random-looking output.

Structural evidence from the OpenVINO IR:

| Export | bf16 ops | fp32 ops | fp16 ops |
|--------|----------|----------|----------|
| `OpenVINO/pi05-libero-fp16-ov` (broken HF) | **0** | 15877 | 3084 |
| Current `main` branch export | **10422** | 7619 | 2059 |
| This PR's export | **10422** | 7619 | 2059 |

The zero bf16 count in the HF model confirms it was traced from a model without bf16 computation. The current `main` branch (hardcoded `use_adarms=[False, True]`) already produces the correct graph with 10422 bf16 tensors — identical to this PR's export.

## Root Cause

The HF model is stale. It was not re-exported after bf16/adaRMS support was added to `PaliGemmaWithExpertModel`. The model code on `main` was already correct.

What was missing was the **config/detection layer**: `Pi05Config` had no `use_adarms` or `adarms_cond_dim` fields, so:
1. Checkpoint introspection was impossible — there was no way to know programmatically whether a checkpoint was trained with adaRMS.
2. `Pi05Config.use_adarms` defaulted to `False`, meaning any code path that constructed the config without going through the hardcode in `Pi05Model` would build a model without adaRMS layers.

### `config.py`
Added `use_adarms: bool = True` and `adarms_cond_dim: int | None = None` to `Pi05Config`. Default is `True` to match the model-level invariant: adaRMS is always enabled for the action expert in `Pi05Model` (matching upstream LeRobot Pi05). Removed the over-constrained `__post_init__` guard that blocked `Pi05Config()` construction when `adarms_cond_dim` was `None` — the model derives the conditioning dim from `action_expert_config.width`, not from config.

### `pretrained_utils.py`
Added `detect_adarms_from_checkpoint(weights_file, hf_config)`:
- Scans safetensors keys for `*.input_layernorm.dense.weight` to detect adaRMS weights
- Injects `use_adarms=True` and `adarms_cond_dim=<dim>` into `hf_config` for checkpoints that predate the config fields

### `policy.py`
Calls `detect_adarms_from_checkpoint()` in `_from_hf()` before `Pi05Config.from_dict()`, enabling automatic detection for legacy checkpoints.

### `model.py`
`Pi05Model.__init__` now accepts `use_adarms: bool` (default `True`) and passes `[False, use_adarms]` to `PaliGemmaWithExpertModel`. `policy._initialize_model` wires `self.config.use_adarms` through.

## Validation

Validated with `test_golden_action` from [openvinotoolkit/physicalai#208](https://github.com/openvinotoolkit/physicalai/pull/208) — **30 samples** from `lerobot/libero_10_image`, `use_random_input_noise=False`, on **Intel Arc A770 GPU** (`openvino==2026.2.1`). Results after commit `76c90d02` on this branch:

| Export | Passing samples | Max L2 | Mean L2 | Threshold | Result |
|--------|----------------|--------|---------|-----------|--------|
| `OpenVINO/pi05-libero-fp16-ov` (stale HF) | 1 / 30 | **14.32** | **7.40** | 2.0 | ❌ FAIL |
| Local re-export from this PR | 30 / 30 | **1.87** | **0.17** | 2.0 | ✅ PASS |

The stale HF model needs to be replaced with a re-export from current `main` (or this PR).

## Testing

Unit tests in `library/tests/unit/policies/test_pi05.py` covering `detect_adarms_from_checkpoint()`.

### Related

- openvinotoolkit/physicalai#208 — golden-action integration test suite


library/src/physicalai/policies/pi05/pretrained_utils.py
```diff
@@ -240,6 +242,49 @@ def resolve_feature_shape(feat_name: str, hf_config: dict[str, Any], feat_stats:
     return (1,)
 
 
+def detect_adarms_from_checkpoint(weights_file: Path, hf_config: dict[str, Any]) -> None:
```
sovrasov, line 245: A unit test with mocked checkpoint/config would be nice to see

sovrasov:
Nice catch, thanks Piotr! Interesting thing is that the model still demonstrate good perf numbers on libero even with incorrect weights loading

library/src/physicalai/policies/pi05/pretrained_utils.py
pwolnows: test added

sovrasov:
@pwolnows could you share a script / snippet how pi05 was created and exported? There are some parameters that can affect exact outputs matching, but preserve results in envs, because the policy performs denoising internall

eugene123tw:
adaRMS is intentionally hardcoded in Pi05Model, not a configuration option:

In model.py, the adaRMS setting is hardcoded:

```python
self.paligemma_with_expert = PaliGemmaWithExpertModel(
    paligemma_config,
    action_expert_config,
    use_adarms=[False, True],  # VLM: disabled, Action expert: ENABLED
    precision=dtype,
    image_size=self._image_resolution[0],
    freeze_vision_encoder=freeze_vision_encoder,
    train_expert_only=train_expert_only,
)
```

This design is **intentional and aligned with upstream LeRobot Pi05**, adaRMS is always enabled for the action expert (`[False, True]`), never disabled.

Adding `use_adarms` and `adarms_cond_dim` to `Pi05Config` would **not change the model behavior** (it's hardcoded).

PR #866, opened 2026-07-29
chore: update PAI rev and rename config to training_config

maxxgx:
## Description

- Update to latest PAI which includes new features
- Temporarily renaming `physicalai.config` to `physicalai.training_config` as it conflicts with the same package name in `physicalai` repo. The naming will be restore once we have consolidated both configs in the `physicalai` repo


samet-akcay:
@maxxgx, do we want to manually update this? Can this be automated ?

maxxgx:
> do we want to manually update this? Can this be automated ?

Probably renovate could do it, though I think it's safer to update it manually as `physicalai` can introduce braking changes.

sovrasov:
I was also thinking of auto-updates, but for now it looks like literally every update contains unpredictable manual changes that renovate can not handle. A PR with uv locks updated can make sense as a starting point though. Perhaps, we can set that up in the future

sovrasov (approved):

AlexanderBarabanov (approved):

PR #870, opened 2026-07-29
docs: VLA Evaluation Harness

alfieroddan:
## Description

Design doc to integrate [vla-evaluation-harness](https://github.com/allenai/vla-evaluation-harness/tree/main).

### DIFF
--- a/library/docs/explanation/benchmark/README.md
+++ b/library/docs/explanation/benchmark/README.md
@@ -5,6 +5,7 @@ Policy evaluation and benchmarking infrastructure.
 ## Documents
 
 - **[Overview](overview.md)** - Architecture and components
+- **[VLA Evaluation Harness](vla-evaluation-harness.md)** - External benchmark model-server integration
 
 ## Components
 
--- a/library/docs/explanation/benchmark/vla-evaluation-harness.md
+++ b/library/docs/explanation/benchmark/vla-evaluation-harness.md
@@ -1,0 +1,235 @@
+# VLA Evaluation Harness Integration
+
+## Purpose
+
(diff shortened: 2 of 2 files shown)


samet-akcay (changes requested):
Thanks just a minor comment..

as we briefly discussed, perhaps we could name the files/objects consistent with the upstream vla-eval? Like you initially proposed?

For example, upstream would expect something like `PhysicalAIModelServer`, same role as `LeRobotModelServer` / `OpenVLAModelServer`?

As far as I understand, benchmark-specific servers use `{Benchmark}{Model}ModelServer`, e.g. `RoboDojoPi05ModelServer`, maybe we could have the same convention ? 

Regarding file naming, they tend to use the original package name such as `lerobot.py`, `openvla.py` etc. Not sure if we can follow their convention here like `physicalai.py` perhaps we could be more explicit here, and use their convention when we upstream ?

alfieroddan:
> as we briefly discussed, perhaps we could name the files/objects consistent with the upstream vla-eval? Like you initially proposed?
>
> For example, upstream would expect something like `PhysicalAIModelServer`, same role as `LeRobotModelServer` / `OpenVLAModelServer`?
>
> As far as I understand, benchmark-specific servers use `{Benchmark}{Model}ModelServer`, e.g. `RoboDojoPi05ModelServer`, maybe we could have the same convention ?
> 
> Regarding file naming, they tend to use the original package name such as `lerobot.py`, `openvla.py` etc. Not sure if we can follow their convention here like `physicalai.py` perhaps we could be more explicit here, and use their convention when we upstream ?

Have updated to follow this naming convention. I agree it's smarter to follow upstream.

PR #873, opened 2026-07-30
fix(smolvla): apply action padding mask in the training loss

Daankrol:
## Summary

`SmolVLAModel.compute_loss` looks up the end-of-episode action padding mask under the key `extra.actions_id_pad`. LeRobot emits it as `action_is_pad`. Because the lookup uses `.get()`, the mismatch silently returns `None`, the `if actions_is_pad is not None:` guard never fires, and **the masking branch has never executed in this repo**.

The consequence is that action-chunk steps falling past the end of an episode (which LeRobot fills by clamping to the final frame's action) are trained on as if they were genuine supervision.

## Why the padding exists

Given `chunk_size=N`, one training sample is (observation at frame `i`, actions for frames `i..i+N-1`). Near the end of an episode that window runs off the end, so LeRobot clamps the query indices and flags the clamped steps:

```python
# lerobot/datasets/dataset_reader.py:222-231
query_indices = {
    key: [max(ep_start, min(ep_end - 1, abs_idx + delta)) for delta in delta_idx]
    for key, delta_idx in self.delta_indices.items()
}
padding = {
    f"{key}_is_pad": torch.BoolTensor(
        [(abs_idx + delta < ep_start) | (abs_idx + delta >= ep_end) for delta in delta_idx]
    )
    for key, delta_idx in self.delta_indices.items()
}
```

The clamped steps are verbatim copies of the last real action. Training on them teaches "freeze in the final pose."

### Small worked example

Episode of 273 frames (indices 0-272), `chunk_size=50`, sample at frame 250. The chunk wants 250-299 but only 250-272 exist:

| chunk step | 0-22 | 23-49 |
|---|---|---|
| source frame | 250-272 | **272, 272, 272, ...** (clamped) |
| `action_is_pad` | `False` | `True` |

So 27 of 50 targets are the same repeated action. On a real pick-and-place dataset (49 episodes, 13,362 frames, chunk 50) this works out to:

- **9.0%** of all `(frame, chunk-step)` action targets are repeated-final-action padding
- **18.0%** of frames produce a chunk that touches padding
- per-episode share ranges **6.3% to 20.1%** (shorter episodes are hit hardest)

If episodes end at a meaningful terminal state, say "gripper open above the drop box", that 9% is explicitly supervising the policy to stop there.

## How this affects model training

Chunked-action policies (ACT, Pi0, Pi0.5, SmolVLA, ...) all request `chunk_size` future actions for every recorded frame, so all of them hit this same end-of-episode clamping. What differs is whether the policy masks the clamped steps out of its training loss:

| Policy | Masks padded actions out of the loss? |
|---|---|
| ACT | Yes |
| SmolVLA | Yes (upstream); **no, before this PR** (this repo) |
| Pi0 | No |
| Pi0.5 | No |

For Pi0/Pi0.5 this is expected (see "Why pi0/pi05 are deliberately left alone" below). For SmolVLA in this repo it was an unintended bug, since the masking code existed but never ran.

Either way, whenever the padded tail *is* trained on (Pi0/Pi0.5 always, and SmolVLA before this fix), the effect on the resulting policy depends entirely on how the demonstrations were recorded:

- **Non-cyclic episodes** (recording stops right at the target, e.g. right after a drop-off) mean the repeated final action reinforces "freeze at the target." That's rarely the behavior you want deployed.
- **Cyclic episodes** (recording continues back to a home/reset pose before stopping) mean the same padding mechanism instead reinforces "hold at home." That's safe, correct, and is exactly what makes chained, repeated execution possible (e.g. picking up several objects one after another without a scripted reset in between).

So for Pi0/Pi0.5, cyclic episodes aren't just good practice, they're a requirement: it's the only way to control what the unmasked padding actually teaches. For ACT and (post-fix) SmolVLA it's "only" good practice, since the padding is masked out of the loss either way, but it's still what enables chained execution at inference time. See the SO-101 dataset review in this repo for a worked example with real recorded-episode numbers (`library/docs/how-to/training/so101-pick-place-dataset-review.md`).

## Changes

1. **Read the correct key.** `extra.actions_id_pad` to `extra.action_is_pad`, matching LeRobot and the sibling ACT port (`policies/act/model.py:239,282,698`), plus the docstring.

2. **Fix the denominator.** Once masking actually applies, `losses.mean()` is wrong: it counts the zeroed-out padded elements in the denominator, shrinking the loss (and therefore the gradient) in proportion to the padding fraction. On the dataset above that is a silent ~9% reduction in effective learning rate. Now:

   ```python
   num_valid = ((~actions_is_pad).sum() * losses.shape[-1]).clamp_min(1)
   loss = losses.sum() / num_valid
   ```

   `clamp_min(1)` guards the fully-padded-chunk case.

Fixing only (1) without (2) would be a half-fix that silently rescales the gradient.

## Upstream comparison

| Policy | Masks action padding? | Key | Reference |
|---|---|---|---|
| **SmolVLA (upstream LeRobot)** | Yes, with `num_valid` denominator | `action_is_pad` | `policies/smolvla/modeling_smolvla.py:380, 387-389, 410-411` |
| **SmolVLA (this repo, before)** | No, silently disabled | `actions_id_pad` (wrong) | `policies/smolvla/model.py:213` |
| **ACT (this repo)** | Yes | `action_is_pad` | `policies/act/model.py:239,282` |
| pi0 / pi05 (upstream LeRobot) | No | n/a | `modeling_pi0.py:1319`, `modeling_pi05.py:1291` |
| openpi Pi0 / Pi0.5 (JAX) | No | n/a | `models/pi0.py`; pad flags dropped in `training/data_loader.py` |

`actions_id_pad` has **zero occurrences anywhere in upstream lerobot**. `actions_is_pad` exists upstream only as a *local variable name*, which is very likely how the typo crept into the dict key during the port. The canonical key is constructed as `f"{key}_is_pad"` with `key == "action"`.

Upstream SmolVLA also uses the valid-element denominator, so change (2) restores parity rather than inventing new behaviour.

## Why pi0 / pi05 are deliberately left alone

pi0 and pi05 have the *same* unmasked behaviour in this repo, but unlike SmolVLA that is **not** a port bug:

- Upstream LeRobot `pi0`/`pi05` both reduce with a plain `losses.mean()` and never read any pad key.
- openpi (the original π implementation these are ported from) drops the pad flags in the dataloader before the model ever sees them.
- Most importantly, **the released pretrained weights (`lerobot/pi05_base` etc.) were trained with this exact unmasked objective.** Changing the loss for fine-tuning would introduce a train/pretrain objective mismatch and break comparability with published π0.5 numbers.

Note that the `losses[:, :, :original_action_dim]` line in pi0/pi05 (and here) is unrelated: it strips **action-dimension** padding (6-DoF zero-padded to `max_action_dim=32`) on axis 2. The chunk-time axis is axis 1. Two different kinds of padding; the comment there can read as though it covers both.

If masking is ever wanted for π, it should land as an opt-in config flag defaulting to `False`, with the same `num_valid` denominator, not a silent behaviour change. Happy to split that into a follow-up issue.

## Tests

New `TestActionPaddingMask` in `library/tests/unit/policies/test_smolvla.py`, driving `SmolVLAModel.compute_loss` against a stubbed inner model (no downloads, ~5 s):

| Test | Guards |
|---|---|
| `test_mask_is_read_from_lerobot_key_only` | Exact key spelling; a typo'd key must not silently half-apply |
| `test_padded_steps_are_excluded_from_the_loss` | Padded steps contribute nothing regardless of magnitude |
| `test_denominator_counts_only_valid_steps` | Values chosen so correct (2.0), mask-with-plain-mean (1.0) and no-mask (50.5) are all distinguishable |
| `test_fully_padded_chunk_does_not_divide_by_zero` | `clamp_min(1)` edge case |
| `test_no_mask_falls_back_to_plain_mean` | Non-chunked datasets keep the old path |
| `test_masking_is_autograd_safe` | Gradients flow; padded steps get exactly zero gradient (this branch never executed before, so its autograd behaviour was previously unverified) |

**4 of the 6 fail on `main`.** Full file: 41 passed. `ruff check` / `ruff format` clean on the changed source.

## Impact

SmolVLA training loss values will change; they were previously diluted by unmasked padding. Any existing SmolVLA benchmark numbers are not directly comparable to post-merge runs. This is expected and is the point of the fix.

## Follow-up

- Consider an opt-in `mask_action_padding` flag for pi0/pi05 (default `False`).
- Worth grepping for other `EXTRA + "..."` string keys that could drift the same way; `.get()` makes this class of bug fully silent.


library/src/physicalai/policies/smolvla/model.py
```diff
@@ -225,7 +225,15 @@ def compute_loss(self, batch: dict[str, torch.Tensor]) -> tuple[torch.Tensor, di
...
 
-        loss = losses.mean()
+        if actions_is_pad is None:
+            loss = losses.mean()
+        else:
+            # Divide by the number of *valid* elements, not the full tensor.
+            # A plain .mean() would count the zeroed-out padded steps in the
+            # denominator and shrink the loss (and gradient) in proportion to
+            # the padding fraction.
+            num_valid = ((~actions_is_pad).sum() * losses.shape[-1]).clamp_min(1)
```
sovrasov, line 232: Won't `(~actions_is_pad).sum()` be already the number of valid elements?

Daankrol: You're right that (~actions_is_pad).sum() alone gives the number of valid timesteps, not the number of valid scalar elements in losses. Since losses has shape [batch, chunk, action_dim] (after truncation to original_action_dim), each valid timestep still contributes action_dim scalar loss values that get summed via losses.sum(). So the denominator needs to be (~actions_is_pad).sum() * losses.shape[-1] to match the numerator, which is what the code does. If we used (~actions_is_pad).sum() alone as the denominator, the loss would be inflated by a factor of action_dim.

sovrasov (approved):

PR #875, opened 2026-07-30
docs: add dataset recording guide

Daankrol:
## Summary

Adds a how-to guide covering **what to record and why**, so demonstration
datasets actually train well:

- The single most important rule: record cyclic episodes (return to the
  start state before stopping), with the technical rationale for
  chunked-action policies moved to an appendix at the end.
- Rig/camera setup checklist before recording.
- Staged recording: adding diversity one axis at a time instead of all at
  once.
- Consistency guidance, including picking one object-selection rule when
  multiple candidates are present.
- Task description guidance for language-conditioned policies.
- A per-policy "how long should I train" section (ACT/Diffusion vs.
  SmolVLA/Pi0/Pi0.5), with `physicalai fit` CLI examples.
- A troubleshooting table mapping symptoms to likely causes.

Points to the existing Physical AI Studio guides
(`application/docs/05-recording-datasets.md`,
`application/docs/06-training-policies.md`) for the actual recording and
review workflow, since that's how datasets are recorded in this product,
rather than documenting raw `lerobot-record` CLI usage.

Replaces the "Custom Datasets - TBA" placeholder row in the how-to index
with a link to the new guide.

## Notes

- Doc-only change.
- Checked with `prettier --check` and `markdownlint-cli2`; all internal
  links (including anchors) verified to resolve.

### DIFF
--- a/library/docs/how-to/data/recording-a-dataset.md
+++ b/library/docs/how-to/data/recording-a-dataset.md
@@ -1,0 +1,256 @@
+# How to record a dataset for imitation learning
+
+A goal-oriented guide for recording demonstration datasets that actually train
+well. It codifies the [LeRobot AGENT_GUIDE
+§5](https://github.com/huggingface/lerobot/blob/main/AGENT_GUIDE.md) data
+recommendations.
+
+> **Good data beats clever models.** Every hour spent on rig setup and recording
+> discipline saves several hours of training and debugging.
+
+This guide covers **what to record and why**. For the click-by-click mechanics
+of the recorder and the episode viewer, see the Physical AI Studio guides:
+
+- [Recording Datasets](../../../../application/docs/05-recording-datasets.md) —
+  creating a dataset, the record loop, reviewing and exporting episodes.
+- [Training Policies](../../../../application/docs/06-training-policies.md) —
+  training a model from the Models page.
+
+Physical AI Studio identifies datasets by their own dataset ID, not a Hugging
+Face repo — there is nothing to configure there before you start.
+
+---
+
+## The single most important rule: record cyclic episodes
+
+**An episode must end in the same state it started in.**
+
+```text
+BAD   home -> approach -> grasp -> move to target -> release              [stop recording]
+GOOD  home -> approach -> grasp -> move to target -> release -> return home -> settle ~1s   [stop recording]
+```
+
+If you stop recording at the target and reset the arm manually before the next
+episode, the dataset never shows the model what to do after a successful
+execution. At inference the policy finishes one repetition and is immediately in
+a state it has never seen — it cannot chain repetitions, and behaviour becomes
+undefined.
+
+Add a deliberate ~1 s settle at home before you click accept, so the terminal
+state is consistent across episodes. There's a second, less obvious reason this
+matters for some policies — see [Why cyclic episodes matter](#why-cyclic-episodes-matter-technical-detail)
+at the end of this guide.
+
+---
+
+## Before you record
+
+Fix these first. More episodes will not compensate for a broken setup.
+
+| Check                       | Why                                                                                           |
+| --------------------------- | --------------------------------------------------------------------------------------------- |
+| Rig and cameras bolted down | Any camera movement between sessions invalidates every earlier episode.                       |
+| Object/background contrast  | If you can barely see the object in the camera feed, neither can the policy.                  |
+| Lighting                    | Diffuse and consistent. No moving shadows. Lighting matters more than resolution.             |
+| Camera coverage             | Can _you_ do the task from the camera views alone? Check every target position, not just one. |
+| Two cameras                 | 1 fixed overhead/front + 1 wrist. Multi-view consistently outperforms single-view.            |
+| Practice runs               | Do 5-10 unrecorded runs first. Hesitant, inconsistent demos teach hesitation.                 |
+
+---
+
+## Staged recording: add diversity one axis at a time
+
+The most common mistake is jumping straight to the full task. If you vary object
+position **and** distractor count **and** target position across 50 episodes,
+every axis ends up with a handful of samples and the policy generalises to none
+of them.
+
+| Stage | Episodes | What changes                                                          | Purpose                                     |
+| ----- | -------- | --------------------------------------------------------------------- | ------------------------------------------- |
+| **1** | ~50      | **One** object, **one** fixed target, position varied in a small zone | Constrained baseline. Proves the rig works. |
+| **2** | ~30      | Add distractor objects, target still fixed                            | Teaches object selection.                   |
+| **3** | ~40      | 3-4 target positions, distractors present                             | Teaches spatial generalisation.             |
+
+**Train after Stage 1 before recording Stage 2.** A quick ACT run tells you
+whether the problem is your rig or your model, in minutes rather than after
+recording 70 more episodes on a broken setup. See
+[How long should I train?](#how-long-should-i-train) below.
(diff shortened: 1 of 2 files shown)


Daankrol:
Note on the "How long should I train?" section: it currently includes an
epoch/step conversion and a note on translating epochs into the Studio
Models page's **training steps** field, because that's the only unit the
GUI currently accepts.

If/when the Studio training settings switch to accepting **epochs**
directly (matching the `physicalai fit --trainer.max_epochs` CLI path),
that conversion section and the "Training from the Studio GUI" subsection
become unnecessary and should be removed — epochs would then be a single
consistent unit across the CLI and the GUI, with no manual translation
needed.

library/docs/how-to/data/recording-a-dataset.md
```diff
@@ -0,0 +1,256 @@
...
+
+## How long should I train?
+
+This differs by policy, mainly because of whether the policy starts from a
+pretrained backbone.
+
+| Policy      | Pretrained backbone?      | Typical epochs                                                     | Notes                                                                                                                 |
+| ----------- | ------------------------- | ------------------------------------------------------------------ | --------------------------------------------------------------------------------------------------------------------- |
+| ACT         | No — trained from scratch | No pretrained schedule; use validation loss to decide when to stop | Fast per step, good first baseline.                                                                                   |
+| Diffusion   | No — trained from scratch | Longer than ACT                                                    | Benefits from more training than ACT on the same data.                                                                |
```
maxxgx, line 159: Do we support diffusion policy in the library?

Daankrol: No we don't. Lerobot does. We should remove this here. @copilot

library/docs/how-to/data/recording-a-dataset.md
MarkRedeman, line 1: Perhaps move this into `application/docs` since it mainly covers the application workflow?

Can also do that later: I'm also fine with keeping this as is here for now while we rethink some of the structure of our application docs.

library/docs/how-to/data/recording-a-dataset.md
```diff
@@ -0,0 +1,255 @@
...
+  creating a dataset, the record loop, reviewing and exporting episodes.
+- [Training Policies](../../../../application/docs/06-training-policies.md) —
+  training a model from the Models page.
+
+Physical AI Studio identifies datasets by their own dataset ID, not a Hugging
+Face repo — there is nothing to configure there before you start.
+
+---
+
+## The single most important rule: record cyclic episodes
```
MarkRedeman, line 23: One thing we could do to improve the UX here: when the user saves or discards an episode we could (optionally) automatically move the leader and follower arms to their home positions.
We could allow setting this home position per dataset.

This way we would help enforce this rule as well as prevent issues where you might stop recording episodes while your arms are in a position where they fall down if disconnected..

library/docs/how-to/data/recording-a-dataset.md
```diff
@@ -0,0 +1,255 @@
...
+
+Fix these first. More episodes will not compensate for a broken setup.
+
+| Check                       | Why                                                                                           |
+| --------------------------- | --------------------------------------------------------------------------------------------- |
+| Rig and cameras bolted down | Any camera movement between sessions invalidates every earlier episode.                       |
+| Object/background contrast  | If you can barely see the object in the camera feed, neither can the policy.                  |
+| Lighting                    | Diffuse and consistent. No moving shadows. Lighting matters more than resolution.             |
+| Camera coverage             | Can _you_ do the task from the camera views alone? Check every target position, not just one. |
+| Two cameras                 | 1 fixed overhead/front + 1 wrist. Multi-view consistently outperforms single-view.            |
```
MarkRedeman, line 55: Perhaps rephrase in: "Two cameras per arm"? Since for bimanual we often use 3 cameras (1 overhead + 1 wrist camera per arm)

library/docs/how-to/data/recording-a-dataset.md
```diff
@@ -0,0 +1,255 @@
...
+every axis ends up with a handful of samples and the policy generalises to none
+of them.
+
+| Stage | Episodes | What changes                                                          | Purpose                                     |
+| ----- | -------- | --------------------------------------------------------------------- | ------------------------------------------- |
+| **1** | ~50      | **One** object, **one** fixed target, position varied in a small zone | Constrained baseline. Proves the rig works. |
+| **2** | ~30      | Add distractor objects, target still fixed                            | Teaches object selection.                   |
+| **3** | ~40      | 3-4 target positions, distractors present                             | Teaches spatial generalisation.             |
+
+**Train after Stage 1 before recording Stage 2.** A quick ACT run tells you
```
MarkRedeman, line 73: I think this part can feel ambiguous:
- Do we intent to have people reuse the model they used on earlier stages?
- Do we intent users to train on the dataset including each stage, or only 1 stage a time?

library/docs/how-to/data/recording-a-dataset.md
```diff
@@ -0,0 +1,255 @@
...
+whether the problem is your rig or your model, in minutes rather than after
+recording 70 more episodes on a broken setup. See
+[How long should I train?](#how-long-should-i-train) below.
+
+### Sizing
+
+Target roughly **45,000+ frames** total (episode count x average episode
+length x fps), which is a solid reference point for a first task. Episode
```
MarkRedeman, line 81: Perhaps rephrase to: "Target roughly 30 minutes recording." This would fit nicely with our default 30 fps.
On the application side we should show the total recording time of each dataset.

library/docs/how-to/data/recording-a-dataset.md
```diff
@@ -0,0 +1,255 @@
...
+
+---
+
+## Consistency
+
+Within and across episodes, keep the same:
+
+- grasp pose and approach vector
+- movement timing and speed
+- **object selection rule** when multiple candidates are present
```
MarkRedeman, line 101: With my dice collection dataset I mainly tried to grab the dice closest to the gripper, but didn't always use this. The Pi05 model I got is working pretty well.

I do expect having a consistent rule will help though, but figured sharing this anecdote may be useful.

library/docs/how-to/data/recording-a-dataset.md
```diff
@@ -0,0 +1,255 @@
...
+| Policy ignores the object       | Camera framing, contrast, or lighting — not the model                                     |
+| Policy flaps / oscillates       | Inconsistent demos, or needs more training                                                |
+| Fails at one specific stage     | Record 10-20 more episodes targeting that stage                                           |
+| Works at trained positions only | Under-sampled diversity — record more episodes per position                               |
+| Completes once then freezes     | Non-cyclic episodes (see [above](#the-single-most-important-rule-record-cyclic-episodes)) |
+| Picks a random object each time | No consistent selection rule in the demonstrations                                        |
+
+---
+
+## Why cyclic episodes matter (technical detail)
```
MarkRedeman, line 222: Will this still be important if we use `action_is_pad` on all our models (is that realistic)?

MarkRedeman (approved):
Thanks Daan! My comments are mainly food-for-thought, let's remove the line about diffusion models as Max suggested.

ArendJanKramer (approved):
Very well written, thank you. It's concise and helpful. What I still find unclear, and maybe that's something difficult to cover, is the relation between embodiments and amount of episodes. For (bi)trossen the model has to learn an embodiment it hasn't seen in the pretrained data for example

PR #883, opened 2026-07-30
feat: use max epochs instead of max steps #846

### LINKED ISSUE #846: [Feature]: Use number of epochs rather than steps for model training
### Problem Statement

I want to use number of epochs as parameter for training rather than steps. It's considered for VLAs to train them 5-10 epochs (https://huggingface.co/docs/lerobot/v0.6.0/en/hardware_guide#training-time).

### Proposed Solution

- Have epochs a primary / additional parameter for training. Determine the number of steps according to the formula (see HF page linked above) 

total_frames    = sum of frames over all episodes      # 50 ep × 30 fps × 30 s ≈ 45,000
steps_per_epoch = ceil(total_frames / (num_gpus × batch_size))
total_steps     = epochs × steps_per_epoch

### Component

Library (VLA framework)

### Alternatives Considered

_No response_

### Acceptance Criteria

_No response_

### Additional Context

_No response_

MarkRedeman:
Change the UI and backend so that we now use `max_epochs` instead of `max_steps`. The goal here is that we will get more consistent training runs as epochs is not depending on the batch size and makes sure the model is trained on all the users's data.

image

### DIFF
--- a/application/backend/src/services/training_backends/local.py
+++ b/application/backend/src/services/training_backends/local.py
@@ -29,6 +29,9 @@ if TYPE_CHECKING:
     from training import TrainingJobSpec
 
 
+_DEFAULT_MAX_EPOCHS = 10
+
+
 class LocalTrainingBackend:
     """Train in the worker process with Lightning."""
 
@@ -88,7 +91,7 @@ def build_spec(context: TrainingContext) -> TrainingJobSpec:
     return TrainingJobSpec(
         # A resumed run's architecture is dictated by the base model's checkpoint.
         policy=(context.base_model or context.model).policy,
-        max_steps=payload.max_steps,
+        max_epochs=payload.max_epochs,
         batch_size=payload.batch_size,
         num_workers=payload.num_workers,
         val_split=payload.val_split,
--- a/application/backend/src/services/training_backends/local.py
+++ b/application/backend/src/services/training_backends/local.py
@@ -29,7 +29,7 @@ if TYPE_CHECKING:
     from training import TrainingJobSpec
 
 
-_DEFAULT_MAX_EPOCHS = 10
+_DEFAULT_MAX_EPOCHS = 5
 
 
 class LocalTrainingBackend:
--- a/application/backend/src/services/training_backends/local.py
+++ b/application/backend/src/services/training_backends/local.py
@@ -20,6 +20,7 @@ from typing import TYPE_CHECKING
 
 from loguru import logger
 
+from schemas.job import _DEFAULT_MAX_EPOCHS
 from services.training_backends._log_format import render_progress_log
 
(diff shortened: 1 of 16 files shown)


application/backend/src/services/training_backends/local.py
```diff
@@ -75,14 +79,11 @@ async def train(self, context: TrainingContext) -> None:
...
-            callbacks=[
-                checkpoint_callback,
-                self._progress_callback(context),
-            ],
+            callbacks=[checkpoint_callback, self._progress_callback(context)],
             accelerator=accelerator,
             strategy=strategy,
             devices=devices,
-            max_steps=payload.max_steps,
+            **train_limit,  # pyrefly: ignore[bad-argument-type]
```
MarkRedeman: Ignoring `bad-argument-type` here is needed because pyrefly couldn't infer the contents of `train_limit`.
A simpler approach could be to simply remove using `max_steps` from studio's training flow. (otherwise we'd need to add a bunch of extra verbose code to resolve this without an ignore)

application/backend/src/services/training_backends/local.py
```diff
@@ -88,7 +91,7 @@ def build_spec(context: TrainingContext) -> TrainingJobSpec:
     return TrainingJobSpec(
         # A resumed run's architecture is dictated by the base model's checkpoint.
         policy=(context.base_model or context.model).policy,
-        max_steps=payload.max_steps,
+        max_epochs=payload.max_epochs if payload.max_epochs is not None else 5,
```
maxxgx:
```suggestion
        max_epochs=payload.max_epochs if payload.max_epochs is not None else _DEFAULT_MAX_EPOCHS,
```

maxxgx (commented):
Looks good, there is a minor inconsistency with the value of `_DEFAULT_MAX_EPOCHS`

application/ui/src/features/models/details/model-details.tsx
```diff
@@ -156,7 +156,11 @@ const TrainingParameters = ({ summary }: { summary: SchemaModelDetailResponse['t
         <View>
             <Heading>Training parameters</Heading>
             <Flex direction='column' gap='size-75'>
-                <DetailRow name='Max steps' value={summary.max_steps} />
+                {summary.max_epochs !== null && summary.max_epochs !== undefined ? (
```
AlbertvanHouten, line 159: When would this be null/undefined?

MarkRedeman: it's mainly for backwards compatibility: if anyone had training jobs before this change then `max_epochs` will be undefined.

AlbertvanHouten (dismissed):

AlbertvanHouten (approved):

PR #884, opened 2026-07-30
ci(trainer): move Trivy scan off the publish gate and stop chasing latest

AlbertvanHouten:
## Description

Fixes the `Trainer Images` publish job, which was failing Trivy scans on `main` due to a blocked egress endpoint (`raw.githubusercontent.com`), and reworks trainer image tagging since no release channel for trainer images exists yet.

### 1. Move the Trivy scan out of the publish gate

The immutable-digest scan in `trainer-images.yml`'s `publish` job kept failing because `harden-runner`'s allowlist was missing endpoints Trivy needs (`raw.githubusercontent.com`, `mirror.gcr.io`), blocking sign/tag on every push to `main`. Rather than just widening the allowlist, the scan step is removed from `publish` entirely and replaced with a new nightly/`workflow_dispatch` job, `trivy-trainer-image-scan`, in `security-scan.yml` that scans the published `:main` tag for both `cuda` and `xpu` with its own scoped allowlist and artifact name.

### 2. New tagging scheme (no `latest`, no release channel yet)

`trainer-images.yml` has no release trigger, so every published trainer image is a dev build — advancing `latest` from `main` was misleading. Publish now produces:

| Tag                          | Advanced by           | Points at                                           |
| ----------------------------- | --------------------- | ---------------------------------------------------- |
| `<version>-dev-<short-sha>`  | every push to `main`  | that exact commit (immutable)                       |
| `main`                       | every push to `main`  | newest `main` build                                 |
| `protocol-<N>`               | every push to `main`  | newest build for that trainer API protocol version  |

Moving tags (`main`, `protocol-<N>`) are only promoted after the immutable digest is signed, so a failed scan/sign leaves them pointing at the last good build.

Docs (`remote-trainer.md`, `.env.trainer.example`, `docker/README.md`) and the trainer compose defaults are updated to reference `main` instead of `latest`.


.github/workflows/trainer-images.yml
```diff
@@ -13,6 +13,8 @@ on:
       - application/trainer/**
       - library/**
       - .github/workflows/trainer-images.yml
+  release:
```
AlexanderBarabanov: Please remove this trigger and all logic related with release publication - it shall be added in a **separate** PR following existing strategy (RC creation -> promotion), for reference
- https://github.com/open-edge-platform/physical-ai-studio/blob/main/.github/workflows/publish-rc-app.yml and 
- https://github.com/open-edge-platform/physical-ai-studio/blob/main/.github/workflows/promote-rc-app-release.yml

.github/workflows/trainer-images.yml
```diff
@@ -129,9 +131,9 @@ jobs:
           python -c 'import json, os; health = json.load(open("health.json")); assert health == {"status": "healthy", "protocol_version": 1, "device_type": os.environ["DEVICE"], "build_revision": os.environ["GITHUB_SHA"], "build_date": os.environ["BUILD_DATE"], "application_version": os.environ["VERSION"]}'
 
   publish:
-    name: Publish, attest, scan, and sign (${{ matrix.device }})
+    name: Publish, attest, and sign (${{ matrix.device }})
     needs: build-and-smoke-test
-    if: github.ref == 'refs/heads/main'
+    if: github.ref == 'refs/heads/main' || github.event_name == 'release'
```
AlexanderBarabanov: Current behavior for application images is to publish it on schedule ([workflow](https://github.com/open-edge-platform/physical-ai-studio/blob/main/.github/workflows/docker-publish.yml)), here we publish trainer images for every `push` to `main` - shall we align publication process?

.github/workflows/trainer-images.yml
AlexanderBarabanov: 1. This will produce full 40-char commit SHA - but tag format shall be aligned with application images existing tags - [example ](https://github.com/open-edge-platform/physical-ai-studio/pkgs/container/physical-ai-studio-cpu/1095945176?tag=0.1.0-dev-d318690d)- `0.1.0-dev-d318690d`, https://github.com/open-edge-platform/physical-ai-studio/blob/main/.github/workflows/docker-publish.yml#L91-L92. Please use `{version}-dev-{shortsha}` format, `version` is already captured in `build-metadata` job from `application/VERSION` file.


2. As release logic will be implemented in a separate workflow and removed from this file - here we can add a second tag `${{ github.ref_name }}` -> so images will be tagged with `main` tag and step `Publish moving main tag after signing` is not required.

.github/workflows/trainer-images.yml
AlbertvanHouten: This is more for in between releases. If we make a change to the trainer and we don't make a release for a while, we can fallback to this main tagged image so developers/people that checkout the code will still have a working SSH deployed trainer.

AlexanderBarabanov (approved):

PR #885, opened 2026-07-31
refactor(library): consume Runtime physicalai.config

samet-akcay:
## Summary

- Remove the Studio-owned `library/src/physicalai/config/` package and rely on Runtime `physicalai.config` (`Config`, `FromConfig`, `instantiate_obj`, `loading`, etc.).
- Point library docs and agent guidance at Runtime config docs instead of duplicated explanation pages under `library/docs/explanation/config/`.
- Drop Studio-local config unit tests that duplicated Runtime coverage; refresh `library/uv.lock` and backend/plugin pins for Runtime `e19a2e52` (includes merged [openvinotoolkit/physicalai#212](https://github.com/openvinotoolkit/physicalai/pull/212)).
- Fix LeRobot checkpoint reload: strip Lightning `type` metadata and use non-strict `dict_to_dataclass` when rebuilding LeRobot configs.


library/docs/development/security.md
```diff
@@ -25,7 +25,7 @@ These rules apply when writing, editing, or reviewing code under `library/` (`ph
...
 
-7. No unbounded recursion in config walkers. Every recursive config walker (e.g., `_instantiate_recursive` in `config/instantiate.py`) must enforce an explicit depth limit via a `depth: int` parameter with a hard cap (`MAX_INSTANTIATE_DEPTH = 32`). Do not rely on `sys.getrecursionlimit()`. Required pattern: `if depth > MAX_INSTANTIATE_DEPTH: raise ValueError(f"Config nesting exceeds maximum depth {MAX_INSTANTIATE_DEPTH}")`.
+7. No unbounded recursion in config walkers. Every recursive config walker in `physicalai.config` (for example `_instantiate.py` and `_normalize.py`) must enforce an explicit depth limit via a `depth` parameter with a hard cap (`_MAX_CONFIG_DEPTH` in `physicalai.config._types`, currently `10`). Do not rely on `sys.getrecursionlimit()`. Required pattern: raise `ConfigError` when depth exceeds the cap.
```
MarkRedeman, line 28: The reference here can probably be removed since it's handled in the framework?

library/src/physicalai/policies/lerobot/policy.py
```diff
@@ -531,10 +531,13 @@ def load_from_checkpoint(
                 msg = f"Checkpoint missing '{POLICY_NAME_KEY}'. Cannot determine which LeRobot policy to reconstruct."
                 raise KeyError(msg)
 
+        config_dict = dict(config_dict)
+        config_dict.pop("type", None)
```
MarkRedeman, line 535: Why is type being removed here? (`dict_to_dataclass` from `physicalai.config` seems equivalent to the previous implementation)

MarkRedeman (approved):
Thanks Samet!

Added two minor suggestions, they can be handled in a separate PR if needed.

AlexanderBarabanov (approved):

PR #893, opened 2026-08-03
fix: inputs reorder for SmolVLA

sovrasov:
## Description

- Alfie's changes allowing reordering inputs from training datasets to map them into expected inputs of pre-trained base model properly + arbitrary empty camera insertion
- Fix for legacy snapshots loading after bumping hf transformers: strict method fails, since the latest hf modules have some parameters which are defined but not used by SmolVLA and therefore are missing in older snapshots
- Updated metadata for export

Example:

```python
    policy = SmolVLA(
        pretrained_name_or_path="lerobot/smolvla_base",
        train_expert_only=False,
        chunk_size=50,
        n_action_steps=50,
        image_key_reorder_map={
            "wrist": 1,
            "top": 0,
        },
        num_cameras=3,
    )
```
In that case, an empty camera image will be appended to the end of observation images, based on the fact reorder_map suggests indices 0 and 1 for existing reordered image keys.

```python
    policy = SmolVLA(
        pretrained_name_or_path="lerobot/smolvla_base",
        train_expert_only=False,
        chunk_size=50,
        n_action_steps=50,
        image_key_reorder_map={
            "side": 2,
            "top": 0,
        },
        num_cameras=3,
    )
```
In that case, an empty camera image will be inserted between top and size images


alfieroddan (approved):
Thanks, looks good!

Perhaps we need some docs on how to actually choose the number of cameras / the order?

For example for marks dataset - I'd first have to look at the smolvla-base config.json - https://huggingface.co/lerobot/smolvla_base/blob/main/config.json. Then i'd have to compare with the original paper -https://arxiv.org/pdf/2506.01844. 

Where I'd find: 
`We found this inconsistency detrimental during
pretraining, and that a consistent camera ordering is rather beneficial for training in this data regime. To address this
standardization challenge, we manually mapped each camera to a standardized view type—prioritizing top, wrist, and
side perspectives–and renamed them as OBS_IMAGE_1, OBS_IMAGE_2, and OBS_IMAGE_3, respectively.`

Then i'd look at the metadata from the dataset - https://huggingface.co/datasets/MarkRedeman/dice-cleanup-combined/blob/main/meta/info.json. 

So then I'd eventually get to this:
```python
policy = SmolVLA(
    pretrained_name_or_path="lerobot/smolvla_base",
    train_expert_only=False,
    image_key_reorder_map={
        "gripper": 1,
        "overview": 0,
    },
    num_cameras=3,
)
```

Perhaps a worked example in the SmolVLA `README.md` or docs would be a good idea?

sovrasov:
> Thanks, looks good!
> 
> Perhaps we need some docs on how to actually choose the number of cameras / the order?
> 
> For example for marks dataset - I'd first have to look at the smolvla-base config.json - https://huggingface.co/lerobot/smolvla_base/blob/main/config.json. Then i'd have to compare with the original paper -https://arxiv.org/pdf/2506.01844.
> 
> Where I'd find: `We found this inconsistency detrimental during pretraining, and that a consistent camera ordering is rather beneficial for training in this data regime. To address this standardization challenge, we manually mapped each camera to a standardized view type—prioritizing top, wrist, and side perspectives–and renamed them as OBS_IMAGE_1, OBS_IMAGE_2, and OBS_IMAGE_3, respectively.`
> 
> Then i'd look at the metadata from the dataset - https://huggingface.co/datasets/MarkRedeman/dice-cleanup-combined/blob/main/meta/info.json.
> 
> So then I'd eventually get to this:
> 
> ```python
> policy = SmolVLA(
>     pretrained_name_or_path="lerobot/smolvla_base",
>     train_expert_only=False,
>     image_key_reorder_map={
>         "gripper": 1,
>         "overview": 0,
>     },
>     num_cameras=3,
> )
> ```
> 
> Perhaps a worked example in the SmolVLA `README.md` or docs would be a good idea?

Here I'm agree, that's a lot of information to fetch from different sources. I'll make a follow-up PR with readme for SmolVLA, which would be kind of a model card at the same time

PR #894, opened 2026-08-03
feat(training): make local XPU training runs survive OOM and mid-run hangs

yogit2020:
Long Pi0.5 fine-tuning runs on Intel Arc XPUs were losing **all** work in two distinct ways. Several runs reached their full step target and still produced no model. This PR fixes the one that is a software problem and bounds the damage from the one that isn't.

Failure analysis with per-boot kernel evidence is in `reports/` — start with `reports/README.md`.

## Failure A — run finishes training, then dies during save/export

A completed 5000-step run (`9200784f`) and a 2000-step run (`ec4c4cef`) both reached their full step target and produced **no model**. The process held ~26 GB of trainer/dataloader state on a 30 GB host, and OpenVINO's conversion pass asked for several GB more. An OOM kill is `SIGKILL`, so no downstream `try/except` can rescue it — the memory has to be handed back first.

- drop `trainer` and the datamodule, then `_release_memory()`, **before** export
- make OpenVINO opt-in via `PHYSICALAI_EXPORT_BACKENDS` (default `"torch"`). Torch export alone is enough to load a model in the GUI, and OpenVINO is the backend that actually killed runs here.
- release cached memory *between* backends, since conversion peaks well above the model's resident size

## Failure B — run hangs mid-training, nothing saved

A CPU core takes a GPU interrupt and spins forever in the `xe` driver's GuC message-ring read (`g2h_read` / `memcpy_fromio`) with interrupts disabled, so it can never yield. Training freezes mid-step with **no error and no exception** — the log simply stops. One run hung for 42 hours before anyone noticed.

This is an Intel `xe`/GuC driver bug and **this codebase cannot fix it**. What it can do is bound how much work a hang destroys:

- checkpoint on a **step interval** instead of monitoring `val/loss`. One epoch is ~11k steps at batch 8 on this dataset, so the previous `val/loss`-monitored callback never fired on a sub-epoch run and saved **nothing at all**.
- `save_top_k=-1` to keep every interval checkpoint (Lightning only accepts `-1`, `0` or `1` when `monitor=None`)
- `val_check_interval` so sub-epoch runs report `val/loss` during the run rather than only at the very end

## Also: NaN guard

`gradient_clip_val=1.0`. A 4k-step bf16 run diverged to NaN at ~step 2775 and kept training to completion; the following run had zero NaN across 1750 steps. Details in `reports/SESSION-2026-07-28-nan-divergence.txt`.

## What was ruled out

Recorded in `reports/` because the wrong conclusions cost real time:

- **GuC firmware age — disproven.** Lockups occur on both 70.45.2 and 70.65.0, one boot was clean for two days on the *older* version, and the first-ever fault predates the upgrade. Upgrading made lockups 4.9× more frequent.
- **GT1 / media tile / `action 3003` — weakened.** Explained one run (4193 GT1 events) but the next had zero GT1 events and zero `action 3003`.
- ReBAR, system RAM, NaN divergence, thermal/PCIe — ruled out with evidence.
- Current leading candidate for Failure B is deep CPU C-states racing the GPU IRQ. Flagged as **untested and circumstantial**, not presented as a conclusion.

## Notes for reviewers

- `reports/` is hardware-specific field documentation (Arc Pro B70 + B50, kernel 6.17.0-41). Happy to move it under `docs/` or drop it from this PR if it doesn't belong in-tree — the code changes stand on their own.
- The two HTML reports are self-contained (inline CSS/JS, light + dark themes, no network calls).
- Datasets are **not** committed — 576 MB of robot video. `reports/DATASETS.md` describes both sets with checksums and which model each produced.
- `PHYSICALAI_EXPORT_BACKENDS` changes the default export set from all-supported to torch-only. That is the deliberate fix for Failure A, but it *is* a behaviour change worth a second opinion.
- Tested on this hardware only; no CI coverage for the XPU path.

### DIFF
--- a/application/backend/src/services/training_backends/local.py
+++ b/application/backend/src/services/training_backends/local.py
@@ -64,12 +64,25 @@ class LocalTrainingBackend:
         strategy = get_lightning_strategy(device_type)
         devices = [device_index] if device_index is not None else 1
 
+        # Step-based checkpoints. The previous val/loss-monitored callback never
+        # fired on short runs: one epoch is ~11k steps at batch 8 on the current
+        # dataset, so a sub-epoch run completed no validation and saved nothing
+        # until the explicit save below. Saving on a step interval instead means
+        # an unattended run always has a recent state on disk.
         checkpoint_callback = ModelCheckpoint(
             dirpath=cache_path,
-            filename="model",
-            save_top_k=1,
-            monitor="val/loss",
-            mode="min",
+            filename="step{step:06d}",
+            # Keep this a divisor of the max_steps values actually used (1000 /
+            # 3000 / 5000 / 12000), so the *final* step always lands on an
+            # interval and gets a checkpoint from this callback. A coarser
+            # interval like 3000 would miss step 5000 entirely and leave the
+            # run depending on the explicit save below, which is the step that
+            # OOMed a completed 5000-step run on 2026-07-26.
+            every_n_train_steps=1000,
+            # save_top_k=-1 keeps every checkpoint. With monitor=None Lightning
+            # only accepts -1, 0 or 1, since there is no metric to rank by.
+            save_top_k=-1,
+            monitor=None,
         )
         csv_logger = CSVLogger(cache_path.parent, name=cache_path.stem)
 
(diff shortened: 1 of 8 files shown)


AlbertvanHouten:
Thanks for this PR. We are centralizing the training logic for local and remote training targets in #904 which also touch the files you have updated. I have opened #909 on top of #904 which resolves the problems you are fixing so I will close this PR in favour of #909. Feel free to review the changes in #909

PR #898, opened 2026-08-04
feat: add SnapFlow distillation for 1-step action generation on Pi0.5 and SmolVLA

Daankrol:
## Summary

Adds SnapFlow self-distillation ([arXiv:2604.05656](https://arxiv.org/abs/2604.05656)) to Pi0.5 and SmolVLA, letting a trained flow-matching VLA be distilled down to **single-forward-pass (1-NFE) action generation**.

The K-step Euler denoising loop is roughly 80% of Pi0.5's end-to-end inference latency. Setting `K = 1` naively does not work — the velocity field is calibrated for small local steps. SnapFlow fixes this with a short self-distillation fine-tune that learns a corrected one-shot jump.

SnapFlow is inherently a **two-phase** recipe: standard flow-matching training first, then a short distillation phase with the VLM frozen. This PR makes both phases first-class.

## Validated on real hardware

I ran this end to end on a physical SO-101 arm using the single-run callback recipe: 5 epochs of standard flow-matching finetuning followed by 3 SnapFlow distillation epochs, on a simple pick-and-place dataset of only **50 episodes**.

| | Baseline (multi-step) | SnapFlow (1-NFE) |
| --- | --- | --- |
| Inference latency | 244 ms | **125 ms** |
| Task accuracy | — | no observed drop |

**~2x lower latency with no accuracy regression I could detect**, from a 3-epoch distillation phase on a small dataset.

## What's included

**Shared policy surface** (`policies/common/snapflow.py`)

- `SnapFlowConfigFields` — frozen dataclass mixin holding `snapflow_enabled/alpha/lambda/num_inference_steps` plus validation, mixed into `Pi05Config` and `SmolVLAConfig`.
- `SnapFlowPolicyMixin` — one shared `enable_snapflow(alpha, lambda_, num_inference_steps)`. Policies supply two small hooks (`inner_model`, `freeze_vlm`) instead of duplicating the logic.

**Phase orchestration** (`train/callbacks.py`)

- `SnapFlowPhaseCallback` flips the policy into SnapFlow mode at a `start_epoch` or `start_step` boundary, then rebuilds the optimizer over the now-trainable parameters (~10% of the model). One `physicalai fit`, no checkpoint handoff.
- `restore_best_teacher` (default `True`): before activating SnapFlow, restores the best-`val/loss` weights from a monitored `ModelCheckpoint` instead of distilling whatever the live in-memory model happens to be at the boundary. SnapFlow's shortcut target is bootstrapped from the model's own marginal-velocity predictions, so distilling an undertrained teacher distills noise. Warns and falls back to the live model if no monitored checkpoint is configured.
- `scope_best_to_phase` (default `True`): resets every monitored `ModelCheckpoint`'s best-tracking state at the boundary. `val/loss` isn't comparable across phases (different `num_inference_steps`), so without this, phase-2 checkpoints could never be recognized as "best" against phase-1 scores. `last.ckpt` and on-disk phase-1 checkpoint files are unaffected.
- Checkpoint filenames get a `snapflow-` prefix once phase 2 activates, so phase-1 and phase-2 checkpoints are distinguishable on disk.

**Configs**

- `pi05_finetune_and_snapflow_distillation.yaml` — single-run worked example (phase 1 + phase 2 via the callback). This is the config used for the SO-101 run above.
- `pi05_snapflow_distillation.yaml`, `smolvla_snapflow_distillation.yaml` — phase-2-only templates for the two-explicit-runs workflow.

**Docs**

- `docs/how-to/training/snapflow_distillation.md` — why two phases are required, all three entry points (callback / two runs / Python API), hyperparameter guidance, and how to verify the transition actually happened.
- `docs/how-to/training/cli.md` — corrects `--ckpt_path` to the namespaced `--fit.ckpt_path` / `--validate.ckpt_path` / etc.

**Fixes picked up along the way**

- `cli/_dispatch.py`: method-level arguments are namespaced under the subcommand.
- Improved logging around phase transitions and best-teacher restoration so it's visible in the console/W&B logs which checkpoint was distilled from.

## Testing

- `library/tests/unit/policies/test_snapflow.py` — config validation, `enable_snapflow` semantics, and assertions that Pi05/SmolVLA share the mixin rather than carrying divergent copies.
- `library/tests/unit/train/test_callbacks.py` — phase-boundary firing on both step and epoch, optimizer rebuild, argument validation, best-teacher restoration, and checkpoint best-tracking scoping.
- `library/tests/unit/cli/test_cli.py` — subcommand-namespaced `ckpt_path` dispatch.

`prek run --all-files library/` is green, and `tests/unit` passes locally


sovrasov (commented):
Thanks for the PR. I have a only high-level comment now:
In each policy you have forward which is conditioned by snap flow_enabled:
```
def forward(batch):
    if not snapflow_enabled:
        return self.forward_fm(batch)
    # snapflow step here
```           
And this structure can be extracted, looks quite regular. Perhaps, it'd require a mixin for internal model, because snapflow is implemented mostly there, but that's it. Ideally, implementing snapflow for yet another flow matcher should be a matter of inheriting from a couple of mixins and implementing just a few methods from there

library/scripts/benchmark_snapflow_libero.py
samet-akcay, line 1: is this an internal script to keep, or aimed to be made for generic purposes? I see that it only uses pi05?

library/src/physicalai/cli/_logging.py
```diff
@@ -0,0 +1,44 @@
...
+
+Library modules only ever call ``logging.getLogger(__name__)``; none of them
+configure handlers, which is correct for a library but means ``logger.info``
+output is dropped when the CLI runs, because the root logger defaults to
+``WARNING``. Lightning solves this for itself by attaching a handler to its own
+``lightning.pytorch`` logger; this module does the same for ``physicalai``.
+
+Deliberately not ``logging.basicConfig``: that mutates the *root* logger and
+would hijack logging for anything embedding the library.
+"""
```
samet-akcay, line 15: How about sth like this?
```suggestion
   """Configure package-scoped console logging for CLI commands.

   Adds a formatted stderr handler to the ``physicalai`` logger when no existing
   logging configuration is detected, without modifying the root logger.
   """
```

library/src/physicalai/policies/common/__init__.py
samet-akcay, line 1: I see that @sovrasov uses `policies/utils` in https://github.com/open-edge-platform/physical-ai-studio/pull/949 and introduces `RTCMixin`

How about we store these in `library/src/physicalai/policies/mixins`?

So we would have something like the following:

 ```text
   policies/
     mixins/
       __init__.py
       rtc.py
       snapflow.py
     utils/
       normalization.py
 ```

library/src/physicalai/policies/common/snapflow.py
```diff
@@ -0,0 +1,431 @@
...
+# Paper defaults (arXiv:2604.05656 §3.6 + Appendix J). Note that the config
+# default for lambda is 1.0 for backwards compatibility with checkpoints trained
+# before the two-phase recipe landed; the phase-2 entry point uses 0.1.
+SNAPFLOW_DEFAULT_ALPHA = 0.5
+SNAPFLOW_DEFAULT_LAMBDA = 0.1
+SNAPFLOW_DEFAULT_NUM_INFERENCE_STEPS = 1
+
+
+@dataclass(frozen=True)
+class SnapFlowConfigFields:
```
samet-akcay: How about
```suggestion
class SnapFlowConfigMixin:
```

to follow the same convention as `SnapFlowModelMixin` and `SnapFlowPolicyMixin`?

This would also be in line with `policies/mixins/snapflow.py` module naming suggestion?

library/src/physicalai/policies/pi05/config.py
```diff
@@ -21,9 +21,11 @@
 
 from physicalai.config import Config
 
+from physicalai.policies.common import SnapFlowConfigFields
+
 
 @dataclass(frozen=True)
-class Pi05Config(Config):
+class Pi05Config(SnapFlowConfigFields, Config):
```
samet-akcay: Same as above
```suggestion
class Pi05Config(SnapFlowConfigMixin, Config):
```

library/src/physicalai/policies/pi05/policy.py
```diff
@@ -632,6 +633,26 @@ def training_step(self, batch: Observation, batch_idx: int) -> torch.Tensor:
         self.log("train/loss", loss_dict["loss"], prog_bar=True)
         return loss
 
+    @property
+    def inner_model(self) -> Pi05Model:
```
samet-akcay, line 637: do we need this `inner_model`? 

in physicalai, we have the following boundary:

 ```text
   Policy = Lightning orchestration wrapper
   Policy.model = the underlying torch model
 ```

so a policy should always have `self.model` that has the underlying torch model. Would it not possible to use it directly with snapflow? Am I missing anything ?

library/src/physicalai/train/callbacks.py
```diff
@@ -258,3 +335,519 @@ def _interact_policy_dataset(trainer: L.Trainer, model: L.LightningModule) -> No
     def on_fit_start(self, trainer: L.Trainer, pl_module: L.LightningModule) -> None:
         """Called at the start of `trainer.fit()`."""
         self._interact_policy_dataset(trainer, pl_module)
+
+
+class SnapFlowPhaseCallback(Callback):
+    """Enable SnapFlow self-distillation at a phase boundary within a single training run.
```
samet-akcay, line 341: nitpick.. perhaps shorten the docstring a bit to make a bit more concise ?

library/pyproject.toml
```diff
@@ -39,6 +39,7 @@ dependencies = [
     "openvino>=2026.0",
     "openvino_tokenizers>=2026.0",
     "onnxscript",
+    "wandb>=0.28.1",
```
samet-akcay: would it be an idea to make this as an optional dependency, like `physicalai[loggers]`? because we might need to support/add other loggers?

samet-akcay (changes requested):
Great effort @Daankrol! Really nice and clean architecture, I like it. 

Added some minor feedback here and there.

library/src/physicalai/policies/pi05/policy.py
samet-akcay: if you always need some interface within the `self.model`, you could perhaps define some protocol, that might look like this ?  
```python
   class SnapFlowModel(Protocol):
       def enable_snapflow(
           self,
           *,
           alpha: float,
           lambda_: float,
           num_inference_steps: int,
       ) -> None: ...

       def freeze_vlm(self) -> None: ...
 ```

but overall we would get rid of `self.inner_model` but use `self.model`?

File: library/pyproject.toml (hunk 6 above)
Comment (sovrasov) [reply library/pyproject.toml]: `wandb` is already in `loggers`, users can pick it from there. Is it required by default for snapflow?

File: library/configs/physicalai/pi05_finetune_and_snapflow_distillation.yaml [hunk 7]
```diff
@@ -0,0 +1,165 @@
...
+    # SnapFlow activates, then gets reset to track phase-2's own best (see
+    # snapflow_distillation.md). save_top_k=3 keeps the 3 best per phase.
+    - class_path: lightning.pytorch.callbacks.ModelCheckpoint
+      init_args:
+        monitor: val/loss
+        mode: min
+        save_top_k: 1
+        filename: "epoch{epoch:03d}"
+        auto_insert_metric_name: false
+        dirpath: /mnt/data/experiments/snapflow-multi-object-augment-best-ckpt
```
sovrasov: Does this path always exist? Absolute paths don't look like something that users would expect by default

library/src/physicalai/policies/pi05/policy.py
Daankrol: Yes we do need it for now. Thats because of SmolVLA. In SmolVLA:  self.model is a SmolVLAModel wrapper that owns a nested VLAFlowMatching at self.model._model. That inner VLAFlowMatching is what actually owns vlm_with_expert, the target-time embedding, and the _snapflow_* flags SnapFlow needs to flip. So SmolVLA's inner_model returns self.model._model, not self.model


We need to restructure SmolVLA so it actually follows our Policy.model setup. I think its best to move that to a different PR.

samet-akcay: Created an issue here https://github.com/open-edge-platform/physical-ai-studio/issues/993

samet-akcay (approved):
lgtm, thanks for the great effort

PR #903, opened 2026-08-04
feat: add a LR monitor to a set of trainer's default callbacks

sovrasov:
## Description

Adding LR monitoring by default makes trainer's logs more complete, as now we monitor only a small subset of training metrics.
Changes only affect library


daniil-lyakhov (dismissed):

alfieroddan (approved):
Nice 😄

PR #908, opened 2026-08-05
feat: add loggers dep group to library

sovrasov:
## Description

Running a training via API can result in a failure, because lightning defaults to tensorboard logging.
New dependencies group allows to explicitly install popular loggers (wandb, tensorboard)


MarkRedeman (approved):

ivanzati (approved):
Thank you, Vladislav!

PR #904, opened 2026-08-04
refactor: Centralize training in one place for remote and local training

AlbertvanHouten:
## Description

This pull request reorganizes the remote training ("trainer") service, improving separation from the main Studio backend and updating documentation, configuration, and CI workflows accordingly. The trainer is now treated as a distinct entry point within the backend project, with its own Dockerfile, environment variables, and documentation. Related CI and code ownership references have been updated, and the trainer-specific workflow is removed in favor of a unified approach. The documentation is clarified to reflect these changes, and a new error is added to handle unsupported resume operations on remote trainers.

**Trainer Service Separation and Documentation:**
- The trainer documentation is moved from `application/trainer/README.md` to `application/backend/docs/remote-trainer.md`, with extensive updates to clarify that the trainer is a separate entry point within the backend, and to explain configuration, environment variables (now `TRAINER_`-prefixed), and Docker image structure. References throughout the repo are updated to point to the new documentation location. 
- A new script entry point `physicalai-trainer` is added to `pyproject.toml` for launching the trainer service.

**CI/CD and Configuration Updates:**
- The dedicated trainer workflow `.github/workflows/trainer.yml` is removed, and the `trainer-images.yml` workflow is updated to build from `application/docker/Dockerfile.trainer` and to trigger on backend, plugin, and Dockerfile changes, reflecting the new structure. 

**Testing and Development Workflow:**
- The `tests` extra in `pyproject.toml` now includes `pytest-mock` and `httpx` to support improved testing.

**Backend Code Improvements:**
- A new `RemoteResumeUnsupportedError` exception is added to clearly reject attempts to resume training from a base model on a remote trainer, which is not supported by the current protocol.

PR #909, opened 2026-08-05
feat(training): Release accelerator memory after training

AlbertvanHouten:
Copy of #894 changed to align with training centralization in #904.

## Failure A — run finishes training, then dies during save/export

A completed 5000-step run and a 2000-step run both reached their full step target and produced **no model**. The process held ~26 GB of trainer/dataloader state on a 30 GB host, and OpenVINO's conversion pass asked for several GB more. An OOM kill is `SIGKILL`, so no downstream `try/except` can rescue it — the memory has to be handed back first.

- drop `trainer` and the datamodule, then `_release_memory()`, **before** export
- release cached memory *between* backends, since conversion peaks well above the model's resident size

### DIFF
--- a/application/backend/src/training/job.py
+++ b/application/backend/src/training/job.py
@@ -153,6 +153,13 @@ def run_training_job(
     has no artifact worth keeping. Callers distinguish a canceled run from a
     completed one by consulting ``should_stop`` again after this returns.
 
+    Before export, the trainer, datamodule, and their references back into the
+    policy are dropped and accelerator memory is released (see
+    :func:`_detach_trainer` and :func:`_release_memory`): the trainer's
+    optimizer state and dataloaders would otherwise stay resident throughout
+    export, and export conversion (OpenVINO in particular) can need several GB
+    on top of that.
+
     Args:
         spec: What to train.
         dataset_root: Local root of the LeRobot dataset to train on.
@@ -210,9 +217,10 @@ def run_training_job(
     _publish(cache_dir, output_dir)
 
     export_policy = _export_policy(spec, policy, output_dir)
+    _detach_trainer(export_policy, trainer)
     del trainer, datamodule, policy
-    _release_memory(accelerator)
-    _export(export_policy, output_dir, report, accelerator=accelerator)
+    _release_memory()
+    _export(export_policy, output_dir, report)
 
 
 def _load_policy_from_checkpoint(spec: TrainingJobSpec, checkpoint: Path) -> Policy:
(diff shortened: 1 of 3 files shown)


application/backend/src/training/job.py
```diff
@@ -186,16 +197,27 @@ def run_training_job(
...
         logger=CSVLogger(cache_dir.parent, name=cache_dir.stem),
         callbacks=[
-            ModelCheckpoint(dirpath=cache_dir, filename="model", save_top_k=1, monitor="val/loss", mode="min"),
+            ModelCheckpoint(
+                dirpath=cache_dir,
+                filename="step{step:06d}",
+                every_n_train_steps=min(CHECKPOINT_EVERY_N_STEPS, spec.max_steps),
+                save_top_k=-1,
+                monitor=None,
+            ),
```
AlbertvanHouten: @Daankrol Could you please check if this change makes sense? For more details why this change was made see the description for failure B in https://github.com/open-edge-platform/physical-ai-studio/pull/894

application/backend/src/training/job.py
```diff
@@ -186,16 +197,27 @@ def run_training_job(
     trainer = Trainer(
         logger=CSVLogger(cache_dir.parent, name=cache_dir.stem),
         callbacks=[
-            ModelCheckpoint(dirpath=cache_dir, filename="model", save_top_k=1, monitor="val/loss", mode="min"),
+            ModelCheckpoint(
+                dirpath=cache_dir,
+                filename="step{step:06d}",
+                every_n_train_steps=min(CHECKPOINT_EVERY_N_STEPS, spec.max_steps),
+                save_top_k=-1,
```
MarkRedeman: Please check if the model that we store here can still be found by our model export. Since the filename won't be `model.ckpt` anymore we might need to tell the export where to find the appropriate model.

application/backend/src/training/job.py
```diff
@@ -206,7 +208,11 @@ def run_training_job(
 
     trainer.save_checkpoint(cache_dir / CHECKPOINT_NAME)
     _publish(cache_dir, output_dir)
-    _export(_export_policy(spec, policy, output_dir), output_dir, report)
+
+    export_policy = _export_policy(spec, policy, output_dir)
+    del trainer, datamodule, policy
+    _release_memory(accelerator)
+    _export(export_policy, output_dir, report, accelerator=accelerator)
```
MarkRedeman: In a future follow up: I think it would be nice if we would allow to do the actual exports in a separate process so that it's memory usage does not affect the training job.

We could then also reuse that mechanism to allow users to export models to different formats _after_ finishing training.

MarkRedeman (dismissed):

Daankrol:
gradient_clip_val=1.0 overrides each policy's own tuned clip norm.
job.py:201. Every physicalai policy implements configure_gradient_clipping as:
clip_val = gradient_clip_val if gradient_clip_val is not None else self.config.optimizer_grad_clip_norm
(act/policy.py:566, pi05/policy.py:688, pi0/policy.py:429)
Lightning calls that hook unconditionally, so the policies already clip today. Passing 1.0 from the Trainer replaces their defaults:
policy	current	after this PR
ACT (act/config.py)	10000	1.0
SmolVLA (smolvla/policy.py)	10	1.0
Pi0 / Pi0.5 (pi0/config.py)	1.0	1.0
GR00T (groot/config.py)	1.0	1.0

I don't think this is a good idea.

Daankrol:
Why this was needed: `del trainer, datamodule, policy` before export didn't free anything in the common case (no `torch.compile` reload) , the exported policy still held `policy._trainer`, which kept the whole trainer graph (optimizer state, dataloaders, strategy) reachable, so `gc.collect()` had no unreachable cycle to reclaim. Added `_detach_trainer` to explicitly break that cycle first, moved the release to run after every export backend (not just before the first), and dropped the accelerator-string matching in `_release_memory` in favor of just checking each backend's availability directly.

Added a regression test (`test_trainer_and_datamodule_are_released_before_export`) that wires the mocked Trainer/datamodule the way Lightning does and asserts they're actually garbage-collected, confirmed it fails without `_detach_trainer` and passes with it.

Commit: 2a97ed24

Daankrol (approved):
Nice job Daan!

PR #919, opened 2026-08-06
fix(pi05,smolvla): exclude padded action steps from training and validation loss

Daankrol:
## Problem

Pi0.5's flow-matching loss reduced with a plain `losses.mean()`, ignoring the `extra.action_is_pad` mask that ACT and SmolVLA already read. LeRobot clamps action-chunk queries at episode boundaries and repeats the terminal action to fill the chunk, flagging those steps as `action_is_pad`. With `chunk_size: 50` at 30fps, that trained the final 1.7 seconds of every episode toward a frozen pose, and it biased the policy to stall at the end of a trajectory.

The unmasked denominator was a second bug: padded steps were counted in `.mean()`, so a sample's loss and gradient shrank in proportion to how much of its chunk was padding.

`compute_val_loss` had the same issue on both Pi0.5 and SmolVLA, diluting the reported validation loss with the same repeated actions.

Flagged as a known gap in `library/docs/how-to/data/recording-a-dataset.md:236-243`.

## Change

Mask padded steps out of both the training and validation loss on Pi0.5 and SmolVLA, dividing by the count of valid elements instead of the full tensor. Moved the shared logic onto the base `Model` class (`in_episode_bound`, `reduce_losses`) instead of duplicating it, since that duplication is how this drifted in the first place.

Preserved: batches without the padding key keep the old plain-mean path, and Pi0.5's SnapFlow consistency-distillation samples stay fully weighted in the training loss (that branch regresses onto a self-generated teacher, not the dataset action, so padding there isn't bad supervision).

Also updated the padding table in `recording-a-dataset.md`: Pi0.5 now masks correctly, split into training/validation columns since those can disagree. Pi0 is untouched by this PR and still shows "No".

## Tests

`TestActionPaddingMask` in both `test_pi05.py` and `test_smolvla.py` covers key-spelling correctness, exclusion, denominator correctness, padding-fraction invariance, all-padded edge case, per-sample masking, autograd correctness, and the SnapFlow exemption, for both `compute_loss` and `compute_val_loss`. Confirmed these fail against the pre-fix code (only the intentional no-change guards still pass).


library/src/physicalai/policies/base/model.py
```diff
@@ -65,6 +67,55 @@ def compute_val_loss(self, batch: dict[str, Any]) -> tuple[torch.Tensor, dict[st
...
+            batch: Preprocessed batch dict, optionally containing
+                ``extra.action_is_pad`` as a ``(batch, chunk)`` bool tensor.
+            exempt_idx: Optional indices of samples that should stay fully
+                weighted despite the padding, for objectives that do not regress
+                onto the dataset action (e.g. self-distillation).
+
+        Returns:
+            A ``(batch, chunk)`` bool mask, ``True`` where the step should
+            contribute to the loss, or ``None`` when the batch carries no
+            padding information (e.g. non-chunked datasets).
```
sovrasov: From one hand it's the base model class, but from another it's a model interface, where we have only major methods that are required to be implemented. Let's keep it clean.
These two methods look rather as extra loss-releted utils and can be converted to utils functions. Static specifier already suggests it can be done

sovrasov (changes requested):

sovrasov:
@eugene123tw please comment if you have any concerns towards pi05 accuracy
It'd be an idea to check pi05 on a real-world dataset after this change

eugene123tw:
@Daankrol

Thanks for this fix, I think the masking math matches upstream lerobot SmolVLA exactly, and applying it to pi05 is a reasonable improvement (note: upstream pi05 doesn't mask padding at all).

My concern is less about correctness and more about the pattern being easy to forget and silently wrong. Every chunked-action policy now has to remember to reduce via:

```py
loss = reduce_losses(losses, in_episode_bound(batch))
```

Two suggestions (can be follow-ups, not blockers):

* Add a regression test safety net — parametrize over every registered chunked policy and assert that flipping `action_is_pad` actually changes the loss. Something like

```py
CHUNKED_POLICIES = ["act", "pi0", "pi05", "smolvla", "groot"]


@pytest.mark.parametrize("policy_name", CHUNKED_POLICIES)
def test_action_is_pad_actually_affects_loss(policy_name, make_tiny_model, make_chunked_batch):
    model = make_tiny_model(policy_name).train()
    batch = make_chunked_batch(policy_name)          # has ACTION shape (B, T, D)

    b, t = batch[ACTION].shape[:2]
    no_pad  = torch.zeros(b, t, dtype=torch.bool)     # nothing padded
    tail_pad = no_pad.clone()
    tail_pad[:, t // 2:] = True                       # second half is boundary padding

    def loss_with(pad_mask):
        torch.manual_seed(0)                          # same noise/time both runs
        b2 = {**batch, f"{EXTRA}.action_is_pad": pad_mask}
        return model.compute_loss(b2)[0]

    loss_no_pad = loss_with(no_pad)
    loss_tail_pad = loss_with(tail_pad)

    assert not torch.allclose(loss_no_pad, loss_tail_pad), (
        f"{policy_name}: loss ignores action_is_pad — padded steps are not masked"
```

* Optional structural fix — make `Model.compute_loss` concrete: it calls an abstract `elementwise_action_loss(batch) -> (B, T, D)` and applies `reduce_losses(..., in_episode_bound(batch))` itself. Policies then only produce per-element losses and can't skip masking; the ones that genuinely differ (ACT/GR00T) opt out by overriding compute_loss explicitly, so the exception is visible rather than accidental.

One more request: since this changes the training objective, could you run a quick before/after on a representative SO101 dataset (same seed/steps/config, only the masking differs) and report eval success rate to confirm no regression? Note the padded fraction on SO101 is small, so I'd expect the result to be "no regression / within noise" rather than a big jump — that's fine.

Daankrol:
Let's wait for real-life so101 benchmark before merging

Daankrol:
Real life testing done on so101. Model trains and performance is good

Daankrol:
@copilot resolve the merge conflicts in this pull request

library/configs/physicalai/pi05_finetune_and_snapflow_distillation.yaml
alfieroddan: are these changes your local stuff or correct? I assume ./experiments would be better?

library/configs/physicalai/pi05_finetune_and_snapflow_distillation.yaml
```diff
@@ -156,10 +156,10 @@ trainer:
         save_top_k: 1
         filename: "epoch{epoch:03d}"
         auto_insert_metric_name: false
-        dirpath: ./experiments/snapflow-multi-object-augment-best-ckpt
+        dirpath: /mnt/data/experiments/pi05_padded_loss
```
sovrasov, line 159: Is the hard-coded absolute path useful here? Why does this PR need to change anything in snapflow recipes?

sovrasov (dismissed):

eugene123tw (approved):

PR #920, opened 2026-08-06
feat: Refine remote trainers

### LINKED ISSUE #912: Refine remote trainers

dwesolow:
## Description

image
image
image

## Related Issues

Closes #912


application/ui/src/features/remote-trainers/remote-trainers-table/remote-trainer-form/remote-trainer-form.tsx
```diff
@@ -0,0 +1,93 @@
...
+                await updateMutation.mutateAsync({
+                    params: { path: { remote_trainer_id: remoteTrainer.id } },
+                    body: { name: normalizedName, url },
+                });
+            } else {
+                await createMutation.mutateAsync({ body: { name: normalizedName, url } });
+            }
+            close();
+        } catch (mutationError) {
+            setError(getApiErrorMessage(mutationError) ?? 'The remote trainer could not be saved. Try again.');
```
MarkRedeman: Could this not be a computed state?
```typescript
const mutationError = createMutation.error ?? updateMutation.error;
error = getApiErrorMessage(mutationError);
```

to simplify you could consider extracting the mutations,
```tsx
const useUpdateMutation = (isEditing: boolean) => {
    const createMutation = $api.useMutation('post', '/api/remote-trainers', {
        meta: {
            invalidates: [['get', '/api/remote-trainers']],
        },
    });
    const updateMutation = $api.useMutation('patch', '/api/remote-trainers/{remote_trainer_id}', {
        meta: {
            invalidates: [['get', '/api/remote-trainers']],
        },
    });

    return isEditing ? updateMutation: createMutation
}
```

application/ui/src/features/remote-trainers/remote-trainers-table/remote-trainers-table.tsx
```diff
@@ -0,0 +1,201 @@
...
+type RemoteTrainersTableProps = {
+    remoteTrainers: SchemaRemoteTrainer[];
+    onEdit: (remoteTrainer: SchemaRemoteTrainer) => void;
+    onDelete: (remoteTrainer: SchemaRemoteTrainer) => void;
+};
+
+export const RemoteTrainersTable = ({ remoteTrainers, onEdit, onDelete }: RemoteTrainersTableProps) => {
+    const [expandedRemoteTrainerId, setExpandedRemoteTrainerId] = useState<string | undefined>(remoteTrainers[0]?.id);
+
+    const health = useRemoteTrainersHealth(remoteTrainers.map((remoteTrainer) => remoteTrainer.id));
```
MarkRedeman, line 169: Do we need to keep this query? I'm wondering if instead of doing this batch query it might not be simpler to have a single query per row (where each row gets a Suspense wrapper?) making the health check a bit more declarative.

MarkRedeman (dismissed):
Thanks Dawid this looks really good!

Some minor notes/suggestions, which I don't mind getting picked up in a follow up PR. The comments from copilot look ok, but can also be picked up later.

application/ui/src/features/remote-trainers/remote-trainers-table/remote-trainers-table.tsx
```diff
@@ -0,0 +1,201 @@
...
+}: RemoteTrainerRowProps) => {
+    const types = deviceTypes(health);
+    const contentId = `remote-trainer-detail-${remoteTrainer.id}`;
+
+    return (
+        <div
+            data-testid={`remote-trainer-row-${remoteTrainer.id}`}
+            onClick={onToggleExpanded}
+            className={`${classes.trainerRow} ${isExpanded ? classes.rowExpanded : ''}`}
+        >
```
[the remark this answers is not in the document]
MarkRedeman: @dwesolow here it doesn't seem to matter much as you've also placed the edit/delete dialogs outside of the table.
In the models table we do have a similar issue where clicking "Run model" opens a modal that is inside of a model's row. Because that dialog is inside of the row's DOM clicking anything inside of this dialog then triggers opening/closing said row. That's something we should fix when updating the models table.

application/ui/src/features/remote-trainers/remote-trainers-table/remote-trainer-form/remote-trainer-form.tsx
dwesolow: Yeah, I found lots of things to improve but did not want to refactor lots of stuff in the same PR but I can update this.

application/ui/src/features/remote-trainers/remote-trainers-table/remote-trainers-table.tsx
dwesolow: then we would not be able to display health status column:
image

application/ui/src/features/remote-trainers/remote-trainers-table/remote-trainer-form/remote-trainer-form.tsx
dwesolow: Updated

MarkRedeman (approved):

application/ui/src/features/remote-trainers/remote-trainers-table/remote-trainers-table.tsx
MarkRedeman: I was thinking that in that case you can do `useRemoteTrainerHealthQuery(trainerId)` in the component itself, instead of getting the health of all at the same time and doing prop drilling.

Not super important though, and might also have other downsides like needing to wrap with suspense etc

PR #921, opened 2026-08-06
fix: SmolVLA became untraceable after upgrading transformers

sovrasov:
## Description

The policy got untraceable by both torch.export and OV.convert_model, the patch fixes that by enrolling vision backbone inference a bit and creating an explicit mask.

No changes on libero_10: avg SR 28.0% with 10 episodes per task.


alfieroddan (approved):
😄

PR #923, opened 2026-08-07
feat(backend): persist SSH-provisioned remote training servers

AlbertvanHouten:
Studio needs to remember which remote machines a user can train on. The safest way to do that is to store nothing secret at all — just a pointer into the SSH config the user already maintains.

## What changed

- `RemoteServerDB` + rewritten `d4f8a1c9b3e6_add_remote_servers.py` (edited in place — it never reached `main`, so there's no released schema and no legacy data).
- `JobProvisioningDB`, keyed by `job_id`: image ref/digest, container id/name, remote and local tunnel ports, alias, trainer build/protocol versions.
- Repository, mapper, schema, and `services/remote_server_service.py`.
- Read-only SSH config reader: lists selectable `Host` aliases (excluding wildcards) and resolves one to its effective hostname/port/user for display.

## Design notes

- A server is identified by one non-secret value, an `ssh_host_alias` naming a `Host` stanza in `~/.ssh/config`. Hostname, port, and user resolve at read time rather than being persisted, so a stored record can't silently disagree with the config.
- **No secret fields.** `username`, `auth_type`, `ssh_secret_encrypted`, `ssh_key_passphrase_encrypted`, `host_key`, `host`, `port`, and the `uq_remote_servers_host_port_username` constraint are deliberately absent.
- **No `to_public()` sanitizer or internal/public schema split** — nothing is secret, so the machinery would imply otherwise.
- The config reader never surfaces `IdentityFile`, `IdentityAgent`, `CertificateFile`, or password directives.

## Out of scope

Nothing here opens a network connection — the reader parses a file, it doesn't dial. The CRUD router lands in #926.

---

**Plan:** [PR 2 — Remote-server persistence by SSH host alias](https://github.com/open-edge-platform/physical-ai-studio/blob/main/application/docs/feature_plans/remote-ssh-trainer-pr-plan.md#pr-2--remote-server-persistence-by-ssh-host-alias) · [backend plan](https://github.com/open-edge-platform/physical-ai-studio/blob/main/application/docs/feature_plans/remote-ssh-trainer-plan.md) · [full PR plan](https://github.com/open-edge-platform/physical-ai-studio/blob/main/application/docs/feature_plans/remote-ssh-trainer-pr-plan.md)

### DIFF
--- a/application/backend/src/services/ssh_config_reader.py
+++ b/application/backend/src/services/ssh_config_reader.py
@@ -178,15 +178,35 @@ def resolve_alias(config_path: Path, alias: str) -> ResolvedSshHost:
     Matches only a literal ``Host`` pattern equal to ``alias``: a wildcard
     stanza that would match it via glob is not a usable target, since aliases
     are created by literal name, and resolution here does no glob matching.
+
+    If the alias is defined by more than one stanza (common with ``Include``),
+    every matching stanza is scanned in file order and a later one overrides an
+    earlier one field-by-field, mirroring real ssh's "first obtained value is
+    used, but Included files are read in-place" semantics as they apply to
+    later stanzas overriding earlier ones - only a field the later stanza
+    actually sets is overridden, not the whole result.
     """
+    found = False
+    hostname: str | None = None
+    port: int | None = None
+    user: str | None = None
     for block in _iter_blocks(config_path):
         for pattern in block.patterns:
             if _is_literal_pattern(pattern) and pattern == alias:
-                return ResolvedSshHost(
-                    alias=alias,
-                    hostname=block.hostname or alias,
-                    port=block.port,
-                    user=block.user,
-                    found=True,
-                )
-    return ResolvedSshHost(alias=alias, found=False)
+                found = True
+                if block.hostname is not None:
+                    hostname = block.hostname
+                if block.port is not None:
+                    port = block.port
+                if block.user is not None:
+                    user = block.user
+                break
+    if not found:
+        return ResolvedSshHost(alias=alias, found=False)
+    return ResolvedSshHost(
+        alias=alias,
+        hostname=hostname or alias,
+        port=port,
+        user=user,
+        found=True,
+    )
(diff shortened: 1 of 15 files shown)


application/backend/src/services/ssh_config_reader.py
```diff
@@ -0,0 +1,206 @@
...
+    Matches only a literal ``Host`` pattern equal to ``alias``: a wildcard
+    stanza that would match it via glob is not a usable target, since aliases
+    are created by literal name, and resolution here does no glob matching.
+
+    If the alias is defined by more than one stanza (common with ``Include``),
+    every matching stanza is scanned in file order and a later one overrides an
+    earlier one field-by-field, mirroring real ssh's "first obtained value is
+    used, but Included files are read in-place" semantics as they apply to
+    later stanzas overriding earlier ones - only a field the later stanza
+    actually sets is overridden, not the whole result.
```
maxxgx: Doesn't the docs contradict itself? 

> "mirroring real ssh's first obtained value is used..." 

then

> "a field the later stanza actually sets is overridden"

It seems that ssh keeps the first value and does not overwrite it.

How common is having multiple "stanza" on the same host?

application/backend/src/services/ssh_config_reader.py
```diff
@@ -0,0 +1,206 @@
...
+
+def list_host_aliases(config_path: Path) -> list[SshHostAliasOption]:
+    """Return every literal ``Host`` alias in the config as a selectable option.
+
+    Wildcard and negated patterns are excluded: an alias is created by a
+    literal ``Host`` name, not a pattern that happens to match one. When
+    ``HostName`` is unset, the alias itself is the effective hostname - real
+    ssh behavior, not an invented default.
+    """
+    options: list[SshHostAliasOption] = []
```
maxxgx: Should this function follow `resolve_alias()` logic? Here looks like you can have duplicate aliases while in `resolve_alias` you overwrite with last view (see other comment). If could group by alias if that would be the case

application/backend/src/services/ssh_config_reader.py
AlbertvanHouten: Yeah, indeed. Fixed the docs now. As for multiple stanza's, I am not familiar with it myself but it seems it is used in some cases through Include where you inherit a base ssh config and then customize only parts of the stanza such as only updating the User or Hostname and inheriting the rest from base.

maxxgx (approved):

PR #924, opened 2026-08-07
feat(backend): add asyncssh transport, output sanitization, and two-tier preflight

AlbertvanHouten:
## Purpose
Build the security-critical remote execution boundary before exposing it through an API or job flow.

## What changed

- **`SshTransport`** over `asyncssh`, using the user's `~/.ssh/config` and `known_hosts`. Every command goes through `shlex.join` before the exec channel, so no argument can break out of its position — commands are never interpolated shell strings.
- **Actionable error mapping** for unknown host key, changed host key, passphrase-protected key with no agent, auth failure, and unreachable host. No raw exception text, hostname, or key path reaches the mapped message.
- **`sanitize_output`** strips `Cc`/`Cf` control and bidi-override characters and ESC-introduced sequences, then caps per-line and total length. Remote stdout/stderr is environment-influenced content, not trusted text.

## Two-tier preflight

The split exists so saving a server stays cheap while still being verified:

- **Tier 1** (cheap, save-gating): alias resolution, reachability, auth, host key, Docker, disk, driver, registry reachability via `HEAD`. Never touches an image. GPU occupancy is reported as a **non-blocking** warning — a busy GPU is transient and must not block editing a target.
- **Tier 2** (expensive, explicit): image resolve/pull, signature check, in-container device probe, protocol compatibility. The only path that can pull; Tier 1 never invokes it.

## Design notes

- Host-key verification is `asyncssh` against `known_hosts` — no TOFU, no pinning, no host-key column.
- XPU host detection is two probes (`xpu-smi`, else Intel render nodes); Tier 2's in-container `torch.xpu.is_available()` is authoritative.
- Nothing uses the transport yet; the API that calls it is #926.

---

**Plan:** [PR 3 — SSH transport, validation, and preflight primitives](https://github.com/open-edge-platform/physical-ai-studio/blob/main/application/docs/feature_plans/remote-ssh-trainer-pr-plan.md#pr-3--ssh-transport-validation-and-preflight-primitives) · [backend plan](https://github.com/open-edge-platform/physical-ai-studio/blob/main/application/docs/feature_plans/remote-ssh-trainer-plan.md) · [full PR plan](https://github.com/open-edge-platform/physical-ai-studio/blob/main/application/docs/feature_plans/remote-ssh-trainer-pr-plan.md)

### DIFF
--- a/application/backend/src/services/ssh/preflight.py
+++ b/application/backend/src/services/ssh/preflight.py
@@ -139,6 +139,27 @@ def _failure_detail(result: CommandResult) -> str | None:
     return _detail(result.stderr or result.stdout)
 
 
+# `xpu-smi discovery` renders an ASCII box-drawing table, e.g.:
+#   +-----------+--------------------------------------------------+
+#   | Device ID | Device Information                                |
+#   +-----------+--------------------------------------------------+
+#   | 0         | Device Name: Intel(R) Data Center GPU Max 1100    |
+#   |           | Vendor Name: Intel(R) Corporation                  |
+#   ...
+# `first_line()` would return the top border, not a device name, so this pulls
+# the "Device Name: ..." cell out of the table instead.
+_XPU_DEVICE_NAME_PATTERN: Final = re.compile(r"Device Name:\s*(.+?)\s*\|?\s*$")
+
+
+def _xpu_discovery_detail(result: CommandResult) -> str:
+    """Return a short device-name summary from `xpu-smi discovery`'s table output."""
+    for line in result.stdout.splitlines():
+        match = _XPU_DEVICE_NAME_PATTERN.search(line)
+        if match:
+            return match.group(1).strip()
+    return result.first_line()
+
+
 class _CheckRecorder:
     """Accumulates one tier's checks, timing each one individually."""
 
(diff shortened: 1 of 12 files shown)


RHeckerIntel (commented):
This is a shitton of code. Could you explain why we need this and a lot of these seem very generic code. did you research vendor options before implementing this?

AlbertvanHouten:
> This is a shitton of code. Could you explain why we need this and a lot of these seem very generic code. did you research vendor options before implementing this?

It is a lot of code but most of it is contained in either the preflight checks which need to check the existence of software custom to our product such as checking if docker is available, drivers are setup correctly and access to our container registry, the sanitization layer which needs to ensure all these checks are safe to run, and the transport layer which runs all this logic and uses asyncssh library to do a lot of the work.

application/backend/src/services/ssh/preflight.py
```diff
@@ -0,0 +1,1176 @@
...
+    protocol_version: int,
+    image_ref_hint: str | None,
+) -> str | None:
+    """Resolve the trainer image, recording ``IMAGE_RESOLVED``.
+
+    Prefers the protocol-pinned tag and falls back to ``latest``. The fallback is
+    a ``WARNING``, not a failure: a protocol bump can land in Studio before CI has
+    published a matching trainer image, and blocking every verification until then
+    would be worse than telling the user which tag was actually used. Only failing
+    both tags is fatal.
```
maxxgx, line 787: In the design doc and plan, it explicitly said `no 'latest' fallback`. I guess the implementation supersede doc in this case

maxxgx (approved):

RHeckerIntel:
> It is a lot of code but most of it is contained in either the preflight checks which need to check the existence of software custom to our product such as checking if docker is available, drivers are setup correctly and access to our container registry, the sanitization layer which needs to ensure all these checks are safe to run, and the transport layer which runs all this logic and uses asyncssh library to do a lot of the work.

Yeah, verified it all, checked for vendor options and whilst sanitize could partly be solved by regex or a separate library. I dont care too much about that. Regex is least readable code ever, but sanitization is a perfect example of correct regex usage.

I just cant believe we're the first to do this all. ftfy and sanitext both preserve \r. I guess we could reduce, but thats all good.

application/backend/src/services/ssh/preflight.py
AlbertvanHouten: It doesn't work currently because if the library has a change which needs an updated trainer image we currently have no way to know about it. Once we manage proper versioning for library releases we can use dedicated tags and check if the library version matches in the trainer vs what is running in studio.

PR #927, opened 2026-08-07
fix: reset selected episodes after delete

### LINKED ISSUE #916: Selected episodes don't get reset / invalidated when deleting episodes from a dataset
When deleting episodes from a dataset it seems like we don't (always?) reset / invalidate the episode list of the dataset.
Specifically when deleting all episodes it seems like the UI still has some episodes selected and will show the "Clear selection" or "Delete" button.

I'm assuming this may be related to some race condition where it takes a very long time for the backend to delete episodes (because it has to re-encode a bunch of episodes due to how lerobot's saves multiple episodes as a single video file).

Please investigate this issue and try to first fix it in on the UI side. Doing any optimistic updates to our query state could here.
It's best to test this on a big dataset (my 200 episode dataset should be fine) since then deleting episodes becomes slower.

Additionally we may want to show some more notifications to the user informing them that the deletion of the episodes is in progress.

MarkRedeman:
Fixes #916 and improves general UX when deleting episodes from a dataset:
- clarify how many episodes will get deleted
- fix typo in empty episodes
- reset selected episodes after deleting them, and wait for deletion to finish

The last item is important as episodes are referenced by episode index, so not resetting this meant that the user might accidentally delete other episodes if they click delete twice.

### DIFF
--- a/application/ui/src/features/datasets/episodes/use-episodes.ts
+++ b/application/ui/src/features/datasets/episodes/use-episodes.ts
@@ -29,7 +29,7 @@ export const useDeleteEpisodeQuery = (dataset_id: string) => {
     });
 
     const deleteEpisodes = (episodeIndices: number[]) => {
-        mutation.mutate({
+        return mutation.mutateAsync({
             params: {
                 path: {
                     dataset_id,
--- a/application/ui/src/routes/datasets/dataset-viewer.tsx
+++ b/application/ui/src/routes/datasets/dataset-viewer.tsx
@@ -20,6 +20,7 @@ import { useQuery } from '@tanstack/react-query';
 import { SchemaEpisode } from '../../api/openapi-spec';
 import { useDeleteEpisodeQuery } from '../../features/datasets/episodes/use-episodes';
 import { paths } from '../../router';
+import { pluralize } from '../../utils';
 import { ReactComponent as EmptyIllustration } from './../../assets/illustration.svg';
 import { useDataset } from './dataset-provider';
 import { EpisodeList } from './episode-list';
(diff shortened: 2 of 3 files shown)


application/ui/src/routes/datasets/dataset-viewer.tsx
```diff
@@ -113,12 +113,19 @@ export const DatasetViewer = () => {
...
+                                    await deletePromise;
+                                }}
                                 title='Delete episodes'
                                 variant='warning'
                                 primaryActionLabel='Delete'
+                                isPrimaryActionDisabled={isPending}
                             >
-                                Are you sure you want to delete the selected episodes?
+                                Are you sure you want to delete {selectedEpisodes.length} selected{' '}
+                                {selectedEpisodes.length === 1 ? 'episode' : 'episodes'}?
```
dwesolow: In geti we usually use
```
const pluralRules = new Intl.PluralRules('en');

export const pluralize = (count: number, singular: string, plural: string): string =>
    pluralRules.select(count) === 'one' ? singular : plural;
```

application/ui/src/features/datasets/episodes/use-episodes.ts
```diff
@@ -29,7 +29,7 @@ export const useDeleteEpisodeQuery = (dataset_id: string) => {
     });
 
     const deleteEpisodes = (episodeIndices: number[]) => {
-        mutation.mutate({
+        return mutation.mutateAsync({
```
dwesolow, line 32: Why `mutateAsync`? Could we not set `setSelectedEpisodes([])` in the `onSuccess`? Do we want to set episodes to `[]` even when deletion fails?

MarkRedeman: This makes it so that the dialog remains open until deletion finishes. It's not a great UX but a decent workaround as sometimes deleting episodes from a dataset takes a long time.

dwesolow (approved):

PR #938, opened 2026-08-10
feat(runtime): run teleoperation on physicalai RobotRuntime

maxxgx:
## Summary

Stacked on #937 (`max/runtime-1-deps-bump`). **Retarget to `main` after the deps PR merges.**

Phase A of the runtime migration: replace `TeleoperateWorker` with a `RuntimeSession` backed by physicalai `RobotRuntime`, including action sources, stream callbacks, and runtime-config export for teleop environments.

- Add `application/backend/src/runtime/` (session, contract, config builder, thread host, stream callbacks)
- Rewire robot control websocket to `RuntimeSession`; add `GET /{environment_id}/runtime-config`
- Remove `TeleoperateWorker` and its tests
- Extend `RobotClientFactory` with shared-robot / port resolver helpers
- UI: recoverable `leader_connection_lost` shows a warning instead of killing the session

## Tests

- [x] Manual: teleop session over websocket (hold ↔ teleop toggle, joint streaming)
- [x] Manual: export runtime-config YAML for a teleop environment


application/backend/src/runtime/config_builder.py
```diff
@@ -0,0 +1,125 @@
...
+        self._resolver = resolver
+
+    async def find_port(self, port_info: SerialPortInfo) -> str | None:
+        port = await self._resolver.find_port(port_info)
+        if port is None:
+            port = port_info.connection_string
+        return None if port is None else self._resolver.resolve_serial_device(port)
+
+
+async def _shared_robot_config(robot: Robot, resolver: PortResolver) -> dict[str, Any]:
```
MarkRedeman: We're doing a very similar thing in the `RobotClientFactory`, are there any plans to move either this logic (perhaps we can get rid of our `PhysicalAIRobotAdapter`?) in there or use the factory here?

MarkRedeman:
Have done some initial testing with the bimanual trossen setup and found no issues. The UX seems to be exactly the same. :+1:

application/backend/src/runtime/config_builder.py
maxxgx: Updated to reuse `RobotClientFactory` here. Should be a bit cleaner now.

I think we can eventually remove `PhysicalAIRobotAdapter` after the runtime migration.

MarkRedeman: Yeah agree, probably the biggest blocker will be enabling / disabling velocity and/or forces?

maxxgx: I think it should be manageable once we have the full event based architecture

application/backend/src/api/environments.py
```diff
@@ -41,6 +45,41 @@ async def get_project_environment(
     return await environment_service.get_environment_by_id(project_id, environment_id)
 
 
+@router.get("/{environment_id}/runtime-config")
```
MarkRedeman, line 48: I don't fully get yet where this is being used, will this be used in a follow up PR?

maxxgx: At the moment it's not wired. Will be used to export runtime inference config

application/backend/src/api/robot_control.py
```diff
@@ -80,66 +81,67 @@ async def robot_websocket(
...
-        outgoing_task = asyncio.create_task(handle_outgoing(websocket, worker, features, fps))
+        definition = robot_client_factory.catalog_registry.get_definition(follower.type)
+        if definition is None:
+            raise ValueError(f"Robot type is not part of the catalog: {follower.type}")
+        event_sink = QueueEventSink()
+        session = RuntimeSession(
+            document,
+            event_sink=event_sink,
+            include_velocities=definition.adapter_options.include_velocities,
+            external_effort_gain=definition.adapter_options.external_effort_gain,
```
MarkRedeman: I'm not sure about adding these as options to the runtime session, to me this feels like a setting solely for the actual robot and might differ between leader and follower.

application/backend/src/runtime/action_source.py
```diff
@@ -0,0 +1,182 @@
...
+        self._hold_target = np.array(robot_state.joint_positions, dtype=np.float32, copy=True)
+        changed = self._follower_source != "hold"
+        self._follower_source = "hold"
+        logger.error("Leader failed for {} consecutive ticks; switching to hold", self._leader_failures)
+        if changed:
+            self._emit_state()
+        self._event_sink.emit(
+            ErrorEvent(
+                message="The leader robot stopped responding. The follower switched to hold.",
+                error_code="leader_connection_lost",
```
MarkRedeman, line 164: I like this feature. Maybe later we can also extend some options so that instead of holding we could let the user configure it to go to the home/rest position?

application/backend/src/runtime/config_builder.py
```diff
@@ -0,0 +1,113 @@
...
+
+if TYPE_CHECKING:
+    from physicalai_studio_plugin import CatalogRobotFactory, SerialPortInfo
+
+    from schemas.project_camera import Camera
+    from schemas.robot import Robot
+
+
+class _StablePortFinder:
+    """Port finder for export: falls back to the stored port and prefers by-id paths."""
```
MarkRedeman: Wdyt of instead of adding this extra class to change the logic of `CatalogRobotFactory` to do the same? (replacing its implementation with this one)
Would that have any disadvantages?

application/backend/src/runtime/contract.py
```diff
@@ -0,0 +1,174 @@
...
+
+
+class DisconnectCommand(CommandBase):
+    command: Literal["disconnect"] = "disconnect"
+
+
+class LoadModelCommand(CommandBase):
+    command: Literal["load_model"] = "load_model"
+    model: dict[str, Any]
+    inference_device: dict[str, Any]
```
MarkRedeman: Can we make these payloads more specific? It's not clear to me yet what values they should hold.
In this case I'd expect an array of strings and not a dict?

Same for the other commands/fields that use `dict[,str Any]`

maxxgx: Good point, changed to `model_id` and `dataset_id` as we don't need the full schema. We can use a pydantic model If a more complex payload is required

application/backend/src/runtime/action_source.py
maxxgx: Thanks, yeah I think can add that later

application/backend/src/api/robot_control.py
maxxgx: I agree, don't feel should be a runtime session concern. I removed them

application/backend/src/runtime/config_builder.py
maxxgx: I moved the `by-id` resolution to the robot factory but I kept a fallback to the DB stored port in the config builder. 
Because we have to distinguish 2 scenarios:
1. live session (e.g. for teleop): robot is plugged so we should return `None` if we can't find that port
2. config export: allow export when a robot is not plugged in (we forward the port saved from the DB)

Do you think it would be possible to always save the `by-id` path in the DB? I remember you mentioned that some robots don't have serial numbers so maybe we can't do that. Not sure if there's a clean way to handle this as `by-id` is also OS specific

MarkRedeman: > Do you think it would be possible to always save the by-id path in the DB? I remember you mentioned that some robots don't have serial numbers so maybe we can't do that. Not sure if there's a clean way to handle this as by-id is also OS specific

Yes the [stararm](https://github.com/servodevelop/Star-Arm-102) that we have doesn't have a serial number. We also had some reports from people that their SO101 didn't have a serial number, this also happens if you'd use the SO101 via socat as a virtualusb.
So I guess that's not something we can guarantee from the application's point of view.

application/backend/src/api/environments.py
```diff
@@ -41,6 +45,43 @@ async def get_project_environment(
...
+async def get_runtime_config(
+    project_id: Annotated[UUID, Depends(get_project_id)],
+    environment_id: Annotated[UUID, Depends(get_environment_id)],
+    environment_service: Annotated[EnvironmentService, Depends(get_environment_service)],
+    robot_client_factory: RobotClientFactoryDep,
+) -> Response:
+    """Download a runnable physicalai teleoperation configuration."""
+    environment = await environment_service.get_environment_by_id(project_id, environment_id)
+    if len(environment.robots) != 1:
+        raise HTTPException(status_code=400, detail="Runtime export requires exactly one follower robot")
```
MarkRedeman, line 58: One thing we could consider is automatically wrap these robots into a `CompositeRobot`, which when implemented would be able to take any `n` robots and prefix each robot's joints with `robot_{n}_`.
This should then be compatible with the current runtime, though I do still wonder as well if it wouldn't be useful to allow multiple robots in a single runtime (without the composite wrapper).

@RHeckerIntel fyi

application/backend/src/api/environments.py
```diff
@@ -41,6 +45,43 @@ async def get_project_environment(
...
+        raise HTTPException(status_code=400, detail="Runtime teleoperation export requires a leader robot")
+
+    try:
+        document = await build_runtime_config(
+            follower=relation.robot,
+            leader=relation.tele_operator.robot,
+            cameras=environment.cameras,
+            fps=30,
+            robot_factory=robot_client_factory,
+            # The rig this describes does not have to be attached to export it.
```
MarkRedeman, line 70: I'm not sure what this comment means.. Does rig reference the robot + camera setup?

application/backend/src/robots/robot_client_factory.py
```diff
@@ -52,23 +40,50 @@ async def build(self, robot: Robot) -> RobotClient:
...
-    def _resolve_port(discovered: list[SerialPortInfo], target: SerialPortInfo) -> str | None:
-        if target.serial_number:
-            for serial_port in discovered:
-                if serial_port.serial_number == target.serial_number:
-                    return serial_port.connection_string
+        A ``/dev/serial/by-id`` alias is only substituted when the robot was
+        matched by serial number. Without one the match is by path alone, so
+        the device was located rather than identified, and a stable-looking
+        alias would dress that guess up as a verified one.
+        """
```
MarkRedeman, line 86: One thing to note about the original intent of this function (I'm not entirely sure if that's still being addressed here): this is used to resolve the device path (say `/dev/ttyACM0`) based on the `SerialPortInfo`, which can have a `serial_number` and/or a `connection_string`. 
If it contains a `serial_number` then we must find a port whose device matches this serial number.
If `port_info.serial_number` is `None` then we return the `connection_string`, if a device is connected to that port.

This logic allows us to store both the serial number and connection string when adding robots, so that the UI can show both.

One other caveat is that `robot_manager` caches the robots it finds, and thus if the user plugs in their robot after `find_robots()` is called then `robot_manager.robots` won't include it.

application/backend/src/runtime/action_source.py
```diff
@@ -0,0 +1,169 @@
...
+if TYPE_CHECKING:
+    from collections.abc import Mapping
+
+    from physicalai.capture.frame import Frame
+    from physicalai.robot.interface import Robot, RobotObservation
+
+    from runtime.contract import CommandMailbox, EventSink, FollowerSource
+
+
+class StudioActionSource:
```
MarkRedeman, line 19: Should this implement/extend the `physicalai.runtime.action_sources.ActionSource`?

application/backend/src/runtime/action_source.py
```diff
@@ -0,0 +1,169 @@
...
+                continue
+            if requested == "teleop" and not self._leader_reads_enabled:
+                self._event_sink.emit(
+                    ErrorEvent(
+                        message="The leader robot is not responding. Reconnect before resuming teleoperation.",
+                        error_code="leader_connection_lost",
+                    )
+                )
+                continue
+            if requested == "policy":
```
MarkRedeman, line 110: We don't specifically check if there is a policy, will that get added later?

application/backend/src/runtime/action_source.py
```diff
@@ -0,0 +1,169 @@
...
+        self._bus = None
+        self._session_id = None
+        self._hold_target = None
+        self._last_leader_action = None
+        self._last_leader_timestamp = None
+        self._leader_failures = 0
+        self._leader_reads_enabled = self._leader is not None
+        self._failure_logged = False
+
+    def _drain_commands(self, robot_state: RobotObservation) -> None:
```
MarkRedeman, line 89: np: I feel like "handle_commands" is a nicer name because drain to me sounds like we won't be doing anything with the commands

AlbertvanHouten (dismissed):

application/backend/src/api/environments.py
maxxgx: Yeah basically rig == environment

application/backend/src/runtime/action_source.py
maxxgx: Shouldn't be necessary as it's a protocol and we only use `StudioActionSource` inside studio for orchestration

application/backend/src/runtime/action_source.py
maxxgx: `self._arm_policy(task=self._task)` does that check already and will emit a "policy not loaded" event if not present

RHeckerIntel (approved):
lgtm; tested locally

Issue #930, opened 2026-08-07
Improve CI check times for PR triggers


AlexanderBarabanov:
Within this task, please review the existing CI workflows that trigger on `pull_request` and improve path-based skipping so that unnecessary jobs are not started for changes that cannot affect the code being tested (e.g. documentation-only edits, README updates, AGENTS.md changes).

Potential improvements

#### `backend.yml`

Current filter:
```yaml
test:
  - library/**
  - physicalai/** << - shall be removed
  - application/backend/**
  - .github/**
```

1. `.github/**` is a catch-all. Editing `copilot-instructions.md`, `renovate.json5` or any skill file currently triggers the full backend pipeline. Consider replacing `.github/**` with:
```
- .github/workflows/backend.yml
- .github/actions/**
```

2. No doc exclusions - files like `application/backend/README.md` and `library/docs/**` match the existing patterns and trigger the pipeline. Consider adding negation patterns:
```
- '!**/*.md'
- '!**/docs/**'
```

#### `library.yml`

Current filter:
```yaml
test:
  - library/**
  - physicalai/** << - shall be removed
  - .github/**
```

1. `.github/**` is a catch-all - same issue as above, consider to narrow filter.
2. No doc exclusions

#### `ui.yml`

Current filter:
```yaml
test:
  - application/ui/**
  - application/backend/**
  - .github/**
```

1. `.github/**` is a catch-all - same issue as above, consider to narrow filter.
2. `application/backend/**` is perhaps too broad for a UI trigger. The full UI pipeline fires for any backend change, including backend test files, docs, or config. Consider replacing with:
```
- application/backend/app/**
- application/backend/pyproject.toml
- application/backend/uv.lock
```
3. No doc exclusions for `application/ui/README.md`, `application/ui/AGENTS.md` etc.

#### `docker-build.yml`

Current filter:
```yaml
test:
  - library/**
  - physicalai/** << - shall be removed
  - application/**
  - .github/actions/docker-build/**
  - .github/workflows/docker-build.yml
```

1. `library/**` is too broad for a Docker build trigger. Library test files, docs, and scripts do not affect the Docker image. Consider replacing with only source and packaging files. e.g.:
```
- library/src/**
- library/pyproject.toml
- library/uv.lock
- library/MANIFEST.in
```

2. No doc exclusions for `application/**` Consider adding:

```
- '!**/*.md'
- '!**/docs/**'
```

[...]

PR #940, opened 2026-08-10
ci: Narrow path filters in PR triggered workflows

ivanzati:
## Description

Narrowing the path filters in PR triggered workflows to stop building images on updated docs for example

## Related Issues

Relates to #930

### DIFF
--- a/.github/workflows/backend.yml
+++ b/.github/workflows/backend.yml
@@ -53,14 +53,16 @@ jobs:
 
       - name: Check paths
         id: check_paths
-        uses: open-edge-platform/geti-ci/actions/check-paths@7e686c1248b3939f8ee8e04e2612da727e379d91
+        uses: open-edge-platform/geti-ci/actions/check-paths@2ed3ce382d1dbd6db52c7c6b93ffdae280a1c06f # check-paths/v0.1.0
         with:
           files_yaml: |
             test:
-              - library/**
+              - library/src/**
+              - library/pyproject.toml
+              - library/uv.lock
+              - library/MANIFEST.in
               - application/backend/**
               - .github/workflows/backend.yml
-              - .github/actions/**
               - '!**/*.md'
               - '!**/docs/**'
 
--- a/.github/workflows/backend.yml
+++ b/.github/workflows/backend.yml
@@ -58,9 +58,11 @@ jobs:
           files_yaml: |
             test:
               - library/**
-              - physicalai/**
               - application/backend/**
-              - .github/**
+              - .github/workflows/backend.yml
+              - .github/actions/**
+              - '!**/*.md'
+              - '!**/docs/**'
 
   code-quality:
     runs-on: ubuntu-latest
--- a/.github/workflows/docker-build.yml
+++ b/.github/workflows/docker-build.yml
@@ -59,7 +59,7 @@ jobs:
 
       - name: Check paths
         id: check_paths
-        uses: open-edge-platform/geti-ci/actions/check-paths@7e686c1248b3939f8ee8e04e2612da727e379d91
+        uses: open-edge-platform/geti-ci/actions/check-paths@2ed3ce382d1dbd6db52c7c6b93ffdae280a1c06f # check-paths/v0.1.0
         with:
           files_yaml: |
             test:
@@ -67,6 +67,7 @@ jobs:
               - library/pyproject.toml
               - library/uv.lock
               - library/MANIFEST.in
+              - library/.dockerignore
               - application/**
               - .github/actions/docker-build/**
               - .github/workflows/docker-build.yml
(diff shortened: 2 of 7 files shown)


.github/workflows/trainer-images.yml
```diff
@@ -9,6 +9,8 @@ on:
       - application/plugin/**
       - library/**
       - .github/workflows/trainer-images.yml
+      - "!**/*.md"
+      - "!**/docs/**"
```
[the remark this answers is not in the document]
ivanzati: Support is here according to https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#onpushpull_requestpull_request_targetpathspaths-ignore

.github/workflows/trainer-images.yml
```diff
@@ -9,6 +9,8 @@ on:
       - application/plugin/**
       - library/**
```
AlexanderBarabanov: Can we narrow `library/**` here as well to something like:
```
- library/src/**
- library/pyproject.toml
- library/uv.lock
- library/MANIFEST.in
- library/.dockerignore
```

Please note, that we need to include `library/.dockerignore` as this file is used to determine which files from the library directory are included in the Docker build context.

.github/workflows/trainer-images.yml
```diff
@@ -17,6 +19,8 @@ on:
...
```
AlexanderBarabanov: Can we narrow `library/**` here as well to something like:
```
- library/src/**
- library/pyproject.toml
- library/uv.lock
- library/MANIFEST.in
- library/.dockerignore
```

.github/workflows/trainer-images.yml
AlexanderBarabanov: For consistency - can we apply approach with `Check paths` job and `required-check` - the same as we have in other workflows?

.github/workflows/docker-build.yml
```diff
@@ -63,11 +63,15 @@ jobs:
         with:
           files_yaml: |
             test:
-              - library/**
-              - physicalai/**
+              - library/src/**
```
AlexanderBarabanov, line 66: Please include `library/.dockerignore` as this file is used to determine which files from the library directory are included in the Docker build context.

.github/workflows/backend.yml
```diff
@@ -58,9 +58,11 @@ jobs:
           files_yaml: |
             test:
               - library/**
```
AlexanderBarabanov: Can we make it more narrow? e.g.

```
- library/src/**
- library/pyproject.toml
- library/uv.lock
- library/MANIFEST.in
```

.github/workflows/backend.yml
```diff
@@ -58,9 +58,11 @@ jobs:
...
-              - physicalai/**
               - application/backend/**
-              - .github/**
+              - .github/workflows/backend.yml
+              - .github/actions/**
```
AlexanderBarabanov: Currently, there is only one local action that is not used by backend - i think we can remove it.

.github/workflows/library.yml
```diff
@@ -72,8 +72,10 @@ jobs:
           files_yaml: |
             test:
               - library/**
-              - physicalai/**
-              - .github/**
+              - .github/workflows/library.yml
+              - .github/actions/**
```
AlexanderBarabanov: Currently, there is only one local action that is not used by library - i think we can remove it.

.github/workflows/ui.yml
```diff
@@ -66,8 +66,13 @@ jobs:
           files_yaml: |
             test:
               - application/ui/**
-              - application/backend/**
-              - .github/**
+              - application/backend/app/**
+              - application/backend/pyproject.toml
+              - application/backend/uv.lock
+              - .github/workflows/ui.yml
+              - .github/actions/**
```
AlexanderBarabanov: Currently, there is only one local action that is not used by ui - i think we can remove it.

.github/workflows/backend.yml
AlexanderBarabanov: Here and in other workflows - please use a new tagged version of this action

`open-edge-platform/geti-ci/actions/check-paths@2ed3ce382d1dbd6db52c7c6b93ffdae280a1c06f # check-paths/v0.1.0`

.github/workflows/trainer-images.yml
ivanzati: Thanks, will do!

.github/workflows/backend.yml
ivanzati: Should not we start backend CI when one of action changes?

.github/workflows/backend.yml
ivanzati: Thank you, done

AlexanderBarabanov (dismissed):
Thank you!

.github/workflows/docker-build.yml
ivanzati: Yep, done

AlexanderBarabanov (approved):

PR #949, opened 2026-08-12
refactor: extract RTC to a mixin

sovrasov:
## Description

The PR extracts RTC functionality into a mixin and extends RTC to SmolVLA.
Other small improvements:
- RTC-related batch keys are added to core observation
- RTC now does not suppose the leftover of the previous chunk is padded and normalized: that's done inside the model.
- A cleaner RTC export
 
```python

policy = Pi05(pretrained_name_or_path="lerobot/pi05_base")
policy.rtc_enabled = True
policy.to_openvino("export_dir")
```


library/docs/explanation/export/README.md
```diff
@@ -265,6 +265,40 @@ input_sample = {
...
+denoising with the unconsumed tail of the previous action chunk. RTC is off by default
+and must be enabled _before_ export, because it changes the traced input schema:
+
+```python test="skip" reason="requires physicalai install and trained model"
+from physicalai.policies.smolvla import SmolVLA
+
+policy = SmolVLA.load_from_checkpoint("checkpoints/last.ckpt")
+
+# Must be set before tracing — it adds the RTC control inputs to the export schema
+policy.rtc_enabled = True
```
alfieroddan, line 280: I'm concerned that our lightning checkpoint would not re-load with `rtc_enabled`.

AFAIK from the implementation `to_torch` checks for `prev_chunk_left_over`, but lightning doesn't?

```python
policy.rtc_enabled = True
trainer.save_checkpoint("model.ckpt")

loaded = SmolVLA.load_from_checkpoint("model.ckpt")
# result
loaded.rtc_enabled  # False
``` 

but:

```python
policy.rtc_enabled = True
policy.to_torch("export/model.pt")

loaded = InferenceModel("export")

# result
TorchAdapter sees prev_chunk_left_over
        ↓
loaded policy.rtc_enabled # True
```

alfieroddan: Perhaps we can add a test for the saving methods to check RTC is still enabled?

library/src/physicalai/policies/pi05/model.py
```diff
@@ -1185,10 +1190,10 @@ def predict_action_chunk(self, batch: dict[str, Any]) -> Tensor:
...
         if self.enable_rtc:
             rtc_kwargs = {
-                "rtc_max_guidance": batch.get("max_guidance_weight", 0.0),
-                "rtc_execution_horizon": batch.get("execution_horizon", 0),
-                "rtc_latency": batch.get("inference_delay", 0.0),
-                "rtc_prev_action_chunk": batch.get("prev_chunk_left_over"),
+                "rtc_max_guidance": batch.get(RTC_MAX_GUIDANCE_WEIGHT, 0.0),
+                "rtc_execution_horizon": batch.get(RTC_EXECUTION_HORIZON, 0),
+                "rtc_latency": batch.get(RTC_INFERENCE_DELAY, 0.0),
+                "rtc_prev_action_chunk": self._pad_prev_chunk(batch.get(PREV_CHUNK_LEFT_OVER)),
```
alfieroddan, line 1196: Should we validate these?

mixin:
```python
class RTCModelMixin:
    def _validate_rtc_inputs(
        self,
        inference_delay: float,
        execution_horizon: int,
        max_guidance_weight: float,
    ) -> None:
        if not 0 <= inference_delay <= execution_horizon <= self._chunk_size:
            raise ValueError(
                "RTC timing values must satisfy "
                "0 <= inference_delay <= execution_horizon <= chunk_size"
            )

        if max_guidance_weight < 0:
            raise ValueError("RTC guidance weight must be non-negative")
```

in model:
```python
if self.enable_rtc:
    inference_delay = batch[RTC_INFERENCE_DELAY]
    execution_horizon = batch[RTC_EXECUTION_HORIZON]
    max_guidance_weight = batch[RTC_MAX_GUIDANCE_WEIGHT]

    self._validate_rtc_inputs(
        inference_delay=inference_delay,
        execution_horizon=execution_horizon,
        max_guidance_weight=max_guidance_weight,
    )

    rtc_kwargs = {
        "rtc_latency": inference_delay,
        "rtc_execution_horizon": execution_horizon,
        "rtc_max_guidance": max_guidance_weight,
        "rtc_prev_action_chunk": self._pad_prev_chunk(batch[PREV_CHUNK_LEFT_OVER]),
    }
```

library/src/physicalai/policies/pi05/model.py
```diff
@@ -1185,10 +1190,10 @@ def predict_action_chunk(self, batch: dict[str, Any]) -> Tensor:
         rtc_kwargs: dict[str, Any] = {}
         if self.enable_rtc:
             rtc_kwargs = {
-                "rtc_max_guidance": batch.get("max_guidance_weight", 0.0),
-                "rtc_execution_horizon": batch.get("execution_horizon", 0),
-                "rtc_latency": batch.get("inference_delay", 0.0),
-                "rtc_prev_action_chunk": batch.get("prev_chunk_left_over"),
+                "rtc_max_guidance": batch.get(RTC_MAX_GUIDANCE_WEIGHT, 0.0),
+                "rtc_execution_horizon": batch.get(RTC_EXECUTION_HORIZON, 0),
```
alfieroddan, line 1194: I also think we should fail early on something like execution horizon being 0.

A clear error may be better when RTC is enabled but an input is missing.

library/src/physicalai/policies/smolvla/model.py
```diff
@@ -1257,6 +1285,21 @@ def sample_actions(  # noqa: PLR0914
                 timestep=time_tensor,
                 target_time=target_time,
             )
+
+            if rtc_prev_action_chunk is not None:
+                prefix_weights = self._compute_prefix_weights(
+                    inference_delay=torch.tensor(rtc_latency, device=device),
```
alfieroddan, line 1291: `torch.as_tensor` better? An un-necessary copy. 

Goes for these in smolvla and pi05:
```python
inference_delay=torch.as_tensor(rtc_latency, device=device),
execution_horizon=torch.as_tensor(rtc_execution_horizon, device=device),
max_guidance_weight=torch.as_tensor(rtc_max_guidance, device=device),
```

library/src/physicalai/policies/mixins/rtc.py
alfieroddan, line 1: I don't expect the below in this PR but I am curious on your thoughts.

## 1. Where should action chunks be trimmed?

What about policies that trim chunks with a postprocessor compared with policies that cut them off internally in `forward()`?

Wouldn't this affect boundary implementations if we did it differently per model.

RTC needs the unexecuted end of the previous chunk. If a model removes that end internally, RTC cannot use it later.

For example:

```text
chunk_size = 12
n_action_steps = 8

Full prediction: [0 1 2 3 4 5 6 7 | 8 9 10 11]
Executed:        [0 1 2 3 4 5 6 7]
RTC keeps:                           [8 9 10 11]
```

### Example A: trim with a postprocessor

The model returns the full chunk:

```python
actions = model.predict_action_chunk(batch)  # 12 actions
actions_to_execute = chunk_trimmer(actions)  # first 8 actions
```

RTC can retain the full model output before the postprocessor trims what is returned for execution:

```text
RTC stores: [8 9 10 11]
Runtime executes: [0 1 2 3 4 5 6 7]
```

### Example B: slice inside the model

The model discards the end before returning:

```python
actions = model.sample_actions(batch)  # 12 actions
return actions[:, :n_action_steps]      # only 8 actions remain
```

Now the caller receives only:

```text
[0 1 2 3 4 5 6 7]
```

The RTC tail `[8 9 10 11]` is already lost and cannot be used by the next prediction.

Should we should return the full 12-action chunk? 

The policy or runtime can execute the first 8 while retaining the final 4 for RTC. All RTC-capable models should follow this same boundary.

## 2. Should we factor out a flow-matching class?

Should we factor out a flow-matching class to align all the implementations?

```python
velocity = self.denoise_step(...)

if previous_chunk is not None:
        velocity = self._rtc_correct(...)
```

library/docs/explanation/export/README.md
alfieroddan, line 1: Document how overlap is calculated? Or better yet RTC has it's own docs page - even if it's small.

```text
chunk_size = 12
execution_horizon = 8
overlap = 12 - 8 = 4
```

For normal synchronous inference, `n_action_steps` can be the execution horizon:

```python
overlap = chunk_size - n_action_steps
```

For asynchronous inference, the runtime should also track how many actions were executed while the next chunk was being generated.

library/src/physicalai/policies/mixins/rtc.py
```diff
@@ -0,0 +1,210 @@
...
+            the policy; see :class:`RTCPolicyMixin`.
+    """
+
+    enable_rtc: bool = False
+
+    # Declared for type checkers only; provided by the host model.
+    _chunk_size: int
+    _max_action_dim: int
+
+    def _pad_prev_chunk(self, prev_chunk: Tensor | None) -> Tensor | None:
```
alfieroddan, line 121: What about temporal alignment during asynchronous RTC inference with lightning models?

I assume the runtime handles this fine and ensures the previous tail is padded to a full chunk size? But should we check this in lightning?

`_pad_prev_chunk()` only pads the action dimension.

For example, with `chunk_size = 12` and `execution_horizon = 8`, the previous tail may contain only four valid actions. Runtime produces a tensor shaped (1, 12, action_dim) containing those four actions followed by eight zero-padded steps. However, `_compute_prefix_weights()` uses execution_horizon = 8, so positions 4 through 7 still receive nonzero weights. 

Wouldn't we want to ignore them? `effective_horizon = min(execution_horizon, remaining)`

Alternatively, should it provide an explicit validity mask?

Should _pad_prev_chunk() pad the temporal dimension for direct PyTorch inference?

library/docs/index.md
```diff
@@ -58,13 +58,15 @@ graph LR
...
-| **Pi0.5**    | Vision-Language-Action Model with Open-World Generalization | [Black et al. 2025](https://arxiv.org/pdf/2504.16054)                          |
+| Policy      | Description                                                     | Reference                                                |
+| ----------- | --------------------------------------------------------------- | -------------------------------------------------------- |
+| **ACT**     | Action Chunking with Transformers                               | [Zhao et al. 2023](https://arxiv.org/abs/2304.13705)     |
+| **SmolVLA** | Lightweight vision-language-action model                        | [Cadene et al. 2025](https://arxiv.org/abs/2506.01844)   |
+| **Pi0.5**   | Vision-Language-Action model with open-world generalization     | [Black et al. 2025](https://arxiv.org/pdf/2504.16054)    |
+| **RLDX-1**  | Flow-matching VLA with a Qwen3-VL backbone and MSAT action head | [RLWRLD RLDX-1](https://huggingface.co/RLWRLD/RLDX-1-PT) |
+
+Additional policies (ACT, Diffusion, SmolVLA, Pi0, Pi0.5, Pi0-FAST, GR00T, X-VLA) are also available
```
alfieroddan, line 68:
```suggestion
Additional policies (Diffusion, Pi0, Pi0-FAST, GR00T, X-VLA) are also available
```

Perhaps confusing with them as 1st party

alfieroddan (commented):
Thanks Vlad - looks really good. Left my comments below

Also for fun I ran a little simulation to viz the effect of RTC.

```python
# SPDX-License-Identifier: Apache-2.0

"""Plot how RTC smooths boundaries across several predicted action chunks.

Run from the ``library`` directory:

    uv run --with matplotlib python -m tmp.t_rtc
"""

from pathlib import Path

from matplotlib import pyplot as plt
import torch

from physicalai.policies.mixins import RTCModelMixin


torch.manual_seed(7)

chunk_size, action_steps, num_chunks = 24, 16, 6
overlap = chunk_size - action_steps
time = 0.5

total_steps = num_chunks * action_steps + overlap
reference = torch.nn.functional.avg_pool1d(
    torch.randn(1, 1, total_steps).cumsum(-1), 7, stride=1, padding=3,
).flatten()
predictions = (reference.unfold(0, chunk_size, action_steps) + 1.8 * torch.randn(num_chunks, 1))[:, None, :, None]
weights = torch.nn.functional.pad(torch.linspace(1.0, 0.2, overlap), (0, action_steps)).view(1, chunk_size, 1)


def rtc_rollout() -> torch.Tensor:
    """Execute action chunks while guiding each from the previous tail."""
    executed, previous = [], None

    for predicted in predictions:
        chunk = predicted
        if previous is not None:
            previous_tail = torch.nn.functional.pad(previous[:, action_steps:], (0, 0, 0, action_steps))
            velocity = RTCModelMixin._rtc_correct(  # noqa: SLF001
                x_t=torch.zeros_like(predicted),
                v_t=-predicted / time,
                prev_chunk_left_over=previous_tail,
                prefix_weights=weights,
                time=time,
                max_guidance_weight=torch.tensor(1.0),
            )
            chunk = -time * velocity

        executed.append(chunk[0, :action_steps, 0])
        previous = chunk

    return torch.cat(executed)


without_rtc = predictions[:, 0, :action_steps, 0].flatten()
with_rtc = rtc_rollout()
boundaries = torch.arange(action_steps, num_chunks * action_steps, action_steps)
without_rtc_jumps = (without_rtc[boundaries] - without_rtc[boundaries - 1]).abs()
with_rtc_jumps = (with_rtc[boundaries] - with_rtc[boundaries - 1]).abs()

if not with_rtc_jumps.mean() < without_rtc_jumps.mean():
    msg = "RTC did not reduce mean boundary discontinuity"
    raise RuntimeError(msg)

plt.figure(figsize=(12, 5))
plt.plot(reference[: num_chunks * action_steps], label="Smooth reference", color="0.65", linewidth=2)
plt.plot(without_rtc, label="Without RTC", color="tab:orange", linestyle="--")
plt.plot(with_rtc, label="With RTC", color="tab:green", linewidth=2)
plt.vlines(boundaries - 0.5, *plt.ylim(), color="black", alpha=0.3, linestyle=":", label="Execution boundary")
plt.xlabel("Action step")
plt.ylabel("Action")
plt.title("RTC smooths asynchronous action-chunk boundaries")
plt.legend()
plt.tight_layout()

output_path = Path(__file__).with_name("rtc.png")
plt.savefig(output_path, dpi=150)
plt.close()
print(  # noqa: T201
    f"Mean boundary jump: {without_rtc_jumps.mean().item():.3f} without RTC, "
    f"{with_rtc_jumps.mean().item():.3f} with RTC"
)
print(f"Plot: {output_path}")  # noqa: T201
```

Image

Mean boundary jump: 2.001 without RTC, 1.016 with RTC

library/docs/index.md
Daankrol: Also we don't have "Diffusion" so we can just drop that.

sovrasov: updated in [6867623](https://github.com/open-edge-platform/physical-ai-studio/pull/949/commits/6867623fa617ac75f098380de77109a8cc3c0b8b)

library/src/physicalai/policies/smolvla/model.py
sovrasov: Addressed in [560535b](https://github.com/open-edge-platform/physical-ai-studio/pull/949/commits/560535b00a089f551227c21177cbd7a0f2a7f8d6)

library/src/physicalai/policies/pi05/model.py
sovrasov: Addressed in [d3aa165](https://github.com/open-edge-platform/physical-ai-studio/pull/949/commits/d3aa1653df084bb10200a4ed4cbd29921cdf00b7)

library/src/physicalai/policies/pi05/model.py
sovrasov: Addressed in [10ff149](https://github.com/open-edge-platform/physical-ai-studio/pull/949/commits/10ff149445c66669cba3b5ba330fb3b8c0c16063)

library/docs/explanation/export/README.md
sovrasov: Added in [2d53fff](https://github.com/open-edge-platform/physical-ai-studio/pull/949/commits/2d53fffb5976e52f672dc29485b9223830f41566)

library/src/physicalai/policies/mixins/rtc.py
sovrasov: Temporal padding is a dynamic shape operation once we expose the horizon parameter as a variable. The initial implementation keeps it outside of model because of this reason I guess.

I think the whole idea or RTC is that execution horizon is significantly smaller than the whole chunk, so there's time for all RTC guidance phases. If the margin is that thin as in the example, inference delay would simply eat it leading to a situation where only few timestamps were corrected by RTC, which could have been done by linear interpolation with roughly the same outcome. 
In the mentioned example, if we imagine inference takes two timestamps, we have only 2 actions left to blend. I'd restrict the horizon to be <= chunk_size // 2, which would eliminate the need of an extra mask input.

library/src/physicalai/policies/mixins/rtc.py
sovrasov: 1. I think we should prohibit RTC when trimming + perform trimming outside of the model, as it's done now. That'd help debugging and allow more flexibility even after a model is exported.
2. What'd be nice to have a unified loop, like lerobot recently implemented.

library/docs/explanation/export/README.md
sovrasov: Extra docs are added in [709105e54f821213a9e2d6a3a03a1636a4457452](https://github.com/open-edge-platform/physical-ai-studio/pull/949/commits/709105e54f821213a9e2d6a3a03a1636a4457452)

library/src/physicalai/policies/mixins/rtc.py
alfieroddan: Makes sense, we'll have to assume runtime loop or whatever the user is using is capable of padding temporally I guess.

alfieroddan (approved):
Thanks Vlad!

PR #952, opened 2026-08-12
feat(runtime): add runtime session zenoh transport

maxxgx:
## Why

#938 runs teleop on `RobotRuntime`, but the session still lives in an `mp.Process` tied to the API. The next PR needs a child that can outlive the API and be found by other clients. This PR is that mechanism change: a detached process talking over Zenoh. Teleop behaviour stays the same.

## What this adds over #938

- A detached `Popen` child instead of `mp.Process`.
- Commands and telemetry over secure, loopback-only Zenoh.
- A one-shot stdin/stdout handshake (`READY` / `ERROR`), then everything else on Zenoh.
- Deterministic Studio-only ports so `SharedRobot` listeners do not collide.
- Closing the browser still stops the child. Owner semantics (attach, idle hold, survive refresh) are the next PR.

## What's next

- **B2 (#967)** — discoverable owner: name lock, attach-or-spawn, idle hold then exit, detach on close.
- **B3** — policy inference and model loading on the same owner.
- **B4** — recording on the runtime session (delete `RobotControlWorker`, close the reverse exclusivity gap, camera claim registry).

Stacked on #938. Do not merge to `main` until that PR lands.

## Tested scenarios

- [x] Teleop on hardware: hold ↔ teleop, joint streaming, close browser → session stops
- [x] Kill the API (`kill -9`) → child exits (B1 keeps this; B2 reverses it)
- [x] Two sessions on the same robot: second spawn fails, first stays alive


application/backend/src/runtime/hosts/session_worker.py
```diff
@@ -0,0 +1,193 @@
...
+from runtime.transport.server import RuntimeZenohServer
+
+if TYPE_CHECKING:
+    from types import FrameType
+
+_WATCHDOG_INTERVAL_S = 0.1
+
+
+def _watch_parent(parent_pid: int, stop: threading.Event) -> None:
+    """Stop when B1's spawning API exits; B2 replaces this with subscriber presence."""
```
AlbertvanHouten, line 31: I guess this B1/B2 is referring to the doc and will be replaced later?

application/backend/src/runtime/transport/server.py
```diff
@@ -0,0 +1,201 @@
...
+if TYPE_CHECKING:
+    from collections.abc import Callable
+
+    from runtime.contract import Command, RuntimeEvent
+
+
+class RuntimeZenohServer:
+    """Bind a process-local RuntimeSession to Studio's Zenoh contract."""
+
+    def __init__(self, name: str, *, instance_id: str | None = None) -> None:
```
AlbertvanHouten, line 24: Shouldn't instance_id be required here?

application/backend/src/runtime/hosts/session_worker.py
maxxgx: this is overwritten in subsequent prs

application/backend/src/runtime/transport/server.py
maxxgx: Applied on top of the stack: https://github.com/open-edge-platform/physical-ai-studio/pull/996/changes/53c7e6af4d79ae5ab5bc39a2636418cde93c8fa6

RHeckerIntel (approved):

PR #963, opened 2026-08-13
chore: Enable remote trainers by default

dwesolow:


AlbertvanHouten:
We can remove the feature flag alltogether. I don't think there will be anyone that disables it as users can just ignore the setting.

MarkRedeman:
> We can remove the feature flag alltogether. I don't think there will be anyone that disables it as users can just ignore the setting.

~~I was thinking of keeping it for some time. Max wanted to do some experiments with IP cameras https://github.com/openvinotoolkit/physicalai/pull/230 that we could conditionally enable using a feature flag~~

nvm, just realized you probably meant the specific flag, not the infrastructure.. :smile:

dwesolow:
> We can remove the feature flag alltogether. I don't think there will be anyone that disables it as users can just ignore the setting.

Sure, updated!

MarkRedeman (approved):

AlbertvanHouten (approved):

PR #967, opened 2026-08-14
feat(runtime): make the session a discoverable runtime owner

maxxgx:
## Why

#952 moved teleop into a detached child over Zenoh. Closing the browser still killed that child, and a second client (refresh, new tab, API restart) could not find it. This PR makes the child a discoverable owner so the session can outlive the UI and re-attach afterwards.

## What this adds over #952

- A host-local name lock on `rt-<follower uuid>`.
- Attach-or-spawn: a second client joins the running session instead of starting another.
- Losing the last subscriber switches to hold (torque on). The child exits after `RUNTIME_IDLE_TIMEOUT_S` (default 45s).
- Closing the browser detaches. An explicit disconnect still stops the child immediately.
- Deleting a held robot returns HTTP 423.
- Session restart from the UI.

Record still works. If a runtime session already holds the follower, starting a recording on that robot is refused so the two paths do not fight for the serial port. The reverse is still open: a runtime session can start while a record session is already driving the arm (to be addressed in B4).

## What's next

- **B3** — policy inference and model loading on the same owner.
- **B4** — recording on the runtime session (delete `RobotControlWorker`, close the reverse exclusivity gap, camera claim registry).

Stacked on #952. Do not merge to `main` until that PR lands.

## Tested scenarios

- [ ] Refresh reattaches to the same child (hold)
- [ ] API restart reattaches
- [x] Idle exit after `RUNTIME_IDLE_TIMEOUT_S`
- [x] Two tabs, one child; closing one does not kill the other
- [ ] Delete a held robot → 423
- [ ] Record while a runtime session holds the follower → error frame

### DIFF
--- a/application/backend/src/api/robot_control.py
+++ b/application/backend/src/api/robot_control.py
@@ -86,7 +86,9 @@ async def handle_incoming(websocket: WebSocket, client: RuntimeSessionClient) ->
                 continue
             client.apply(command)
     except WebSocketDisconnect:
-        pass
+        # Browser close / refresh is a detach. The child stays up until an
+        # explicit disconnect event or idle timeout.
+        logger.debug("Robot control websocket closed; detaching from the runtime session")
 
 
 async def start_runtime_session(
--- a/application/backend/src/api/robot_control.py
+++ b/application/backend/src/api/robot_control.py
@@ -10,12 +10,12 @@ from fastapi.websockets import WebSocketDisconnect
 from loguru import logger
 from pydantic import ValidationError
 
-from api.dependencies import RobotClientFactoryDep, get_project_id, get_robot_id, get_robot_service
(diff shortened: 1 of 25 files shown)


application/backend/src/api/robot_control.py
```diff
@@ -82,38 +79,50 @@ async def handle_incoming(websocket: WebSocket, client: RuntimeSessionClient) ->
                 continue
             client.apply(command)
     except WebSocketDisconnect:
-        client.apply(DisconnectCommand())
+        pass
```
AlbertvanHouten: Shouldn't this still disconnect or log some information about the disconnecT?

maxxgx: Added log. Only explicit disconnect command should shut it down

RHeckerIntel (approved):

PR #978, opened 2026-08-17
fix(deps): remove redundant pytorch-triton-xpu from xpu extras

AlbertvanHouten:
## Problem

Both `pytorch-triton-xpu==3.5.0` and `triton-xpu==3.7.0` were installed simultaneously, writing into the same `triton/` site-packages directory and corrupting it. Torch's XPU wheel now declares `triton-xpu` as its own transitive dependency, making the explicit `pytorch-triton-xpu` pin (added before torch had this transitive dep) redundant and actively harmful.


MarkRedeman (approved):
One thing to note: I kept getting a `AttributeError: module 'triton' has no attribute 'language'` error after running `uv sync` on this branch.
I was able to get it working after removing the current venv environment (`rm -rf .venv`) and reinstalling the dependencies.

AlbertvanHouten:
> One thing to note: I kept getting a `AttributeError: module 'triton' has no attribute 'language'` error after running `uv sync` on this branch. I was able to get it working after removing the current venv environment (`rm -rf .venv`) and reinstalling the dependencies.

Both pytorch-triton-xpu and triton-xpu share the same destination directory in the venv so any old venv will face this issue. Freshly created venvs should no longer run into this.

PR #976, opened 2026-08-17
feat(robots): stream joints without a runtime session

maxxgx:
## Why

The robot detail page spawned a full `rt-` session to display numbers it never writes. It collided with the environment form. Opening both on the same arm produced HTTP 423 for two pages that are not in conflict.

## What this adds over #971

- A read-only websocket at `/api/projects/{project_id}/robots/{robot_id}/observations/ws`.
- It attaches a `SharedRobot` subscriber (the same owner a live session already uses) and streams named `<joint>.pos` features. No `RobotRuntime`, no `SessionNameLock`, no `session_worker`.
- The detail page Connect switch opens that socket. Restart is gone from this page; it stays on the environment form, which still uses `useJointState`.

A read-only attach to avoid `423` error: `SharedRobot` is multi-subscriber by design.

## What's next

- **B3a** — cameras on the session, identity split so only environment-derived surfaces mint `rt-` sessions.
- **B3b** — policy inference; delete the model-worker stack.
- **B4** — recording on the runtime session (delete `RobotControlWorker`).

Stacked on #971. Do not merge to `main` until that PR lands.

## Stack

```
main
 └── #938 teleop
      └── #952 zenoh transport
           └── #967 runtime owner
                └── #971 physicalai bump
                     └── this PR (max/runtime-6-robot-observations)
```

## Tested scenarios

- [x] Connect on the robot detail page
- [x] Teleop still works
- [x] Detail page + environment form on the same robot at once: both work

<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>

PR #984, opened 2026-08-17
feat(runtime): hold cameras on the session and split identity

maxxgx:
## Why

The runtime session held no cameras, and any camera-set change looked like a different session (HTTP 423). The environment form and a later record client could not share a child: one needed frames, the other did not. Next phase PRs (inference and recoding) need those frames on the session before they can use them.

## What this adds over #976

- One control websocket: `/api/projects/{project_id}/runtime/ws`. The old `/robots/ws` route is gone.
- Handshake carries `follower_id`, optional `leader_id`, and optional `camera_ids`, all resolved against the project.
- Session identity is robot + leader + fps. Camera keys live in `/metadata` separately. A client that needs more cameras restarts the session; a subset attaches.
- The child instantiates, connects, and tears down the document's cameras (`validate_on_connect=True`, `overwrite_settings=False`).
- The environment form sends those camera ids on the handshake. Adding a camera does not restart the live preview.
- The socket URL includes `follower_id` so two follower panels do not share one handshake. Restart clients ignore events until they attach to the new generation.

## What's next

- **B3b** — policy inference; delete the model-worker stack.
- **B4** — recording on the runtime session (delete `RobotControlWorker`, camera claim registry).

Stacked on #976. Do not merge to `main` until that PR lands.

## Stack

```
main
 └── #938 teleop
      └── #952 zenoh transport
           └── #967 runtime owner
                └── #971 physicalai bump
                     └── #976 robot observations
                          └── this PR (max/runtime-7-session-cameras)
```

## Limits

- Editing a camera's resolution under a live session is ignored until the next connect. Will be address at a later stage
- A camera restart displaces whoever was attached; they rejoin the superset.

## Tested scenarios

- [x] Teleoperate from the environment form with cameras in the environment
- [x] Add a camera while the preview is live: robot panel keeps streaming
- [x] Refresh the environment form: reattaches to the same child
- [x] Client needing extra cameras restarts; subset attaches
- [x] Camera publisher at the wrong resolution fails connect (does not silently read)
- [ ] Change the leader → 423 and Restart works

### DIFF
--- a/application/backend/src/api/environments.py
+++ b/application/backend/src/api/environments.py
@@ -7,7 +7,7 @@ from fastapi.responses import Response
 from physicalai.config import to_yaml
 
 from api.dependencies import RobotClientFactoryDep, get_environment_id, get_environment_service, get_project_id
-from runtime.config_builder import build_runtime_config, runtime_config_change_me
+from runtime.config_builder import RUNTIME_FPS, build_runtime_config, runtime_config_change_me
 from schemas.environment import Environment, EnvironmentWithRelations, TeleoperatorRobotWithRobot
 from services.environment_service import EnvironmentService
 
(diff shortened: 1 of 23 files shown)


RHeckerIntel (dismissed):

maxxgx:
Folded into #992 so remaining runtime work ships as one PR against main.

PR #987, opened 2026-08-18
feat(runtime): run policy inference on the session

maxxgx:
## Why

The inference page still used the old model-worker path, not the runtime session from the last PRs.

That old path had few bugs: the task text never reached the model (VLAs need it), image colors were swapped, and the joint stream froze for seconds while a model loaded. The session already holds the robot and cameras (#984). This PR moves Play onto that session and deletes the old stack.

## What this adds over #984

- Loading a model happens on a background thread. The arm keeps ticking. If the model’s cameras do not match the environment, you get a clear error at load.
- Pressing Play resets leftover actions first, so the arm moves from where it is, not from an old chunk. The first Play inference also runs in the background, so the joint stream does not block.
- The inference page uses `/runtime/ws` and the modes `hold`, `teleop`, and `policy`. Failed Play/load shows an error. Camera panels reuse the existing live preview widget.
- The old `ModelWorker` / poller / mixer stack is gone.

## What's next

- **B4** — recording on the same session (then `RobotControlWorker` can go).

Stacked on #984. Do not merge to `main` until that PR lands.

## Stack

```
main
 └── #938 teleop
      └── #952 zenoh transport
           └── #967 runtime owner
                └── #971 physicalai bump
                     └── #976 robot observations
                          └── #984 session cameras
                               └── this PR (max/runtime-8-policy-inference)
```

## Limits

- Record cannot drive the arm from a model until B4. The record UI never did that anyway.
- Record still uses the old provider; inference uses the new one. B4 deletes the old one.
- First tick after opening a session still pays process startup. We did not add a warm process pool. Revisit if load-to-first-action is slower than ~10 seconds.
- Hardware checks and that timing median were **not run**.

## Tested scenarios

- [ ] Run a known ACT checkpoint on the inference page
- [x] Run a VLA checkpoint (Pi0, Pi0.5, or SmolVLA)
- [x] Load a model and watch the joint stream: no stutter during load
- [ ] Rename a camera, then load a model trained before the rename: error at load naming both sets
- [ ] Load-to-first-action median over five real runs (threshold 10s)


RHeckerIntel (dismissed):

maxxgx:
Folded into #992 so remaining runtime work ships as one PR against main.

PR #991, opened 2026-08-18
feat(peft): add LoRA/DoRA fine-tuning support for Pi0 and Pi0.5

Daankrol:
## Summary

Adds LoRA/DoRA (PEFT) fine-tuning support, shared across policies via a new
`physicalai.policies.mixins.peft` package, and wires it into both Pi0.5 and Pi0.

- **`physicalai.policies.mixins.peft`** (new, policy-agnostic): `PeftConfigMixin`,
  `PeftModelMixin`, `PeftPolicyMixin`, and shared helpers
  (`build_lora_config`/`inject_lora`/`merge_lora_`/`is_lora_injected`) built on
  `peft.inject_adapter_in_model` (not `get_peft_model`) so the module tree and
  state-dict keys are preserved for checkpointing/export/tracing. Lives under
  `policies/mixins/` alongside `SnapFlowConfigMixin`/`SnapFlowModelMixin`/
  `SnapFlowPolicyMixin`, the existing precedent for a shared Config/Model/Policy
  mixin triple.
- **`PeftPolicyMixin`** owns the full LoRA lifecycle for a host policy:
  - `_inject_lora()`: called from the policy's `_initialize_model()` once the model
    is built and any pretrained weights are loaded.
  - `export()`: merges LoRA into a disposable deep copy of the model and delegates
    to `super().export()` via cooperative `super()`, so exported artifacts stay
    `peft`-free. Host policies no longer need their own `export()` override; simply
    placing `PeftPolicyMixin` ahead of `ExportablePolicyMixin` in the class's base
    list is sufficient.
  - `on_fit_start()`: raises if `lora_enabled=True` but no adapters were ever
    injected, so a policy that forgets to call `_inject_lora()` fails loudly at the
    start of training instead of silently doing full fine-tuning.
- **Config safety**: `PeftConfigMixin.__post_init__` now rejects `lora_enabled=True`
  combined with a non-default `tune_*`/`freeze_*` value, since LoRA injection
  freezes all base parameters and those flags would otherwise be silently
  overridden. Each policy declares its own `_PEFT_EXCLUSIVE_FLAGS` mapping (Pi0:
  `tune_paligemma`/`tune_action_expert`/`tune_vision_encoder`; Pi0.5:
  `freeze_vision_encoder`/`train_expert_only`) to opt into this check.
- **Pi0.5**: `Pi05Config`/`Pi05Model`/`Pi05` integrate the mixins. Adds a working
  `pi05_lora.yaml` example config (rank 32, alpha 32, ~10x LR per LeRobot's PEFT
  docs) using a public dataset.
- **Pi0 / Pi0.5 (in `pi0/policy.py`)**: same integration pattern applied to
  `Pi0Config`/`Pi0Model`/`Pi0`, replacing the old, unused `pi0/components/lora.py`
  stub.
- LoRA is **disabled by default** (`lora_enabled=False`) in both; this only adds the
  capability, not a behavior change for existing configs.

## Scope / follow-ups

- Library + CLI only (`physicalai fit --config ...`). No backend/UI changes in this
  PR; a follow-up would be needed to expose LoRA options in the GUI training flow.
- ACT, SmolVLA, and GR00T are not covered here.
- A broader `Policy` lifecycle refactor (giving `_initialize_model`/`export` real,
  enforced extension points shared with `SnapFlowPolicyMixin`) was discussed in
  review and is left for a separate PR; see the review thread for details.

## Testing

- `test_peft.py` (new): model-agnostic tests for the shared config fields,
  injection/merge helpers, the `export()` cooperative-`super()` hoist, and the
  `on_fit_start()` guard, using synthetic `nn.Module`s and host classes.
- `test_pi05.py` / `test_pi0.py`: LoRA config validation/serialization, the new
  `lora_enabled` + `tune_*`/`freeze_*` conflict validation, plus
  `@pytest.mark.slow` integration tests against real (small-variant) models:
  trainable-param reduction, dtype handling (fp32 adapters under bf16 base, `auto`
  mode), DoRA injection, forward/backward gradient flow, merge-before-export
  prediction equivalence, and Lightning checkpoint round-trip.
- 197 non-slow tests pass across the three test files; all new `@pytest.mark.slow`
  Pi0/Pi0.5 integration tests verified passing; the full `tests/unit/policies/`
  suite (442 tests) passes.

## Other cleanup in this PR

- Regenerated the stale `pyrefly-baseline.json` entry for the removed
  `pi0/components/lora.py`.
- Added `wandb/` to `.gitignore`.
- Fixed a stale `pi05.yaml` comment claiming `peft` is installed via the
  `pi0`/`pi05` extras; it is a core dependency (`pyproject.toml`).


library/src/physicalai/policies/pi05/policy.py
```diff
@@ -149,6 +152,14 @@ def __init__(  # noqa: PLR0913
         # Finetuning
         freeze_vision_encoder: bool = False,
         train_expert_only: bool = False,
+        # LoRA
+        lora_enabled: bool = False,
```
eugene123tw, line 154: When `lora_enabled=True`, `inject_lora` freezes all existing parameters before injecting adapters, so `tune_action_expert` / `tune_paligemma` / `tune_vision_encoder` become no-ops. Is that intentional? Specifically, setting `tune_action_expert=False + lora_enabled=True` won't actually freeze the action expert, its q_proj/v_proj (and the action/time heads) are still LoRA-adapted by default, so it still receives gradient updates via the adapters even we disable gradient for those modules. Should this interaction be called out in the lora_enabled docstring, or should tune_*=False also exclude those submodules from the default LoRA target regex?

library/src/physicalai/policies/pi0/policy.py
```diff
@@ -129,6 +137,14 @@ def __init__(  # noqa: PLR0913, PLR0917
...
         decay_lr: float = 2.5e-6,
         grad_clip_norm: float = 1.0,
+        # LoRA / DoRA fine-tuning
+        lora_enabled: bool = False,  # noqa: FBT001, FBT002
+        lora_rank: int = 32,
+        lora_alpha: int | None = None,
+        lora_dropout: float = 0.05,
+        lora_target_modules: str | tuple[str, ...] | None = None,
+        lora_adapter_dtype: Literal["float32", "auto"] = "float32",
+        lora_use_dora: bool = False,  # noqa: FBT001, FBT002
```
eugene123tw, line 144: Doing would work for pi05, but adding LoRA to a policy today requires more manual wiring than a "mixin" implies and beyond mixing in PeftConfigMixin/PeftModelMixin/PeftPolicyMixin, you also have to remember to:

* call self._inject_lora() yourself inside _initialize_model() at the right spot
* manually override `export()` to swap in `_merged_lora_model_for_export()` with a try/finally
* thread all `lora_*` kwargs through every `__init__/_from_hf`

The first two are the risky ones — if a future policy (ACT/SmolVLA/GR00T/etc) forgets to call _inject_lora(), lora_enabled=True silently trains nothing, with no error.

Would it be worth giving `physicalai.policies.base.Policy` real extension points (`_on_model_built()`, `_prepare_model_for_export()`) that `PeftPolicyMixin` overrides via cooperative super(), instead of requiring each policy's `_initialize_model/export` to remember to call into PEFT explicitly? That way inheriting the mixin is sufficient and no manual invocation needed. Happy to sketch this out further if useful.

library/src/physicalai/policies/mixins/peft/__init__.py
samet-akcay, line 1: Is `peft` a policy? 

In https://github.com/open-edge-platform/physical-ai-studio/pull/768, many similar components are placed into `library/src/physicalai/policies/components` directory. Perhaps, `peft` can also be placed there?

samet-akcay: another pr... but this would also mean that we might need to refactor snapflow stuff and place it into `components` directory. 

or `peft` into `mixin`/ for now :)

library/src/physicalai/policies/pi0/config.py
```diff
@@ -48,13 +59,6 @@ class Pi0Config(Config):
     tune_action_expert: bool = True
     tune_vision_encoder: bool = False
```
samet-akcay, line 72: when `inject_lora` is called, it freezes params like so;

```python
for param in module.parameters():
    param.requires_grad = False
```

say, user sets `tune_<some-module>: true/false`, then this will be overwritten as well, right? 

Is this expected/intended, do we need to communicate with users?

samet-akcay: ah I see Eugene already provided a similar feedback down below

library/src/physicalai/policies/peft/policy.py
```diff
@@ -0,0 +1,99 @@
...
+
+        Returns:
+            A merged deep copy of ``self.model``, or ``None`` if there is nothing to merge.
+        """
+        self_ = cast("_PeftPolicyHost", self)
+        if not (self_.config.use_lora and self_.model is not None and is_lora_injected(self_.model)):
+            return None
+        merged_model = copy.deepcopy(self_.model)
+        merge_lora_(merged_model)
+        return merged_model
```
samet-akcay: Have you tried exporting a policy? Is this approach efficient in terms of memory consumption? as it might temporarily need memory for two copies of the model?

library/pyproject.toml
```diff
@@ -39,6 +39,7 @@ dependencies = [
     "openvino>=2026.3",
     "openvino_tokenizers>=2026.3",
     "onnxscript",
+    "peft>=0.20.0,<1.0.0",
```
samet-akcay, line 42: any thoughts on peft being part of the core, or optional dependencies?

library/src/physicalai/policies/pi05/policy.py
```diff
@@ -726,6 +784,41 @@ def get_supported_export_backends() -> list[str | ExportBackend]:
         """
         return [ExportBackend.TORCH, ExportBackend.OPENVINO]
 
+    def export(
```
samet-akcay: so would this mean each policy need to implement `export` method? 

`ExportablePolicyMixin` normally handles any export-related stuff, I guess the new implementation does now follow this?

samet-akcay: Maybe `PeftPolicyMixin` could have `export` method implementation. 

If your policy implementation be like this:

 ```python
   class Pi05(PeftPolicyMixin, SnapFlowPolicyMixin, ExportablePolicyMixin, Policy):
       ...
 ```

then the export in peft policy mixin would take precendence, and would be like

 ```text
   Pi05.export()
   -> PeftPolicyMixin.export()
   -> ExportablePolicyMixin.export()
 ```

library/src/physicalai/policies/pi05/policy.py
samet-akcay: Ok, I've checked the codebase, our policy implementations. we could potentially utilize Lightning's `configure_model` hook in our `Policy`, which we don't do much. 

We could have something like this.. --rough sketch

 ```python
   class Policy(L.LightningModule):
       def configure_model(self) -> None:
           if self.model is None:
               self._initialize_model_from_datamodule()

           self._on_model_initialized()
 ```

 Then the PEFT mixin could provide:

 ```python
   class PeftPolicyMixin:
       def _on_model_initialized(self) -> None:
           super()._on_model_initialized()

           if self.config.use_lora and not is_lora_injected(self.model):
               self._inject_lora()
 ```

 This would give us:

 ```text
   Lightning configure_model()
       -> policy builds model
       -> pretrained weights are loaded
       -> PEFT mixin injects adapters
       -> Lightning configures optimizers
       -> checkpoint state is restored
 ```

but this requires some discussion. it is somewhat related to @alfieroddan current molmo stuff as well. 
Perhaps we could arrange a call to discuss how to tackle this together with `Config` resolution?

library/pyproject.toml
Daankrol: It's small pure python and transitive dep because transformers/accelarate. It will probably be used for future models too. I recommend to use as core.

library/src/physicalai/policies/mixins/peft/__init__.py
Daankrol: Went with your mixins/ suggestion rather than components/. The components/ dirs in this repo (groot/components/, pi0/components/) only hold leaf nn.Modules, no mixins today. mixins/snapflow.py is the closer match (a Config/Model/Policy mixin triple), so I moved the package to policies/mixins/peft/ and re-exported the three mixins from policies/mixins/__init__.py alongside SnapFlow's.

library/src/physicalai/policies/pi0/policy.py
Daankrol: Agreed on the risk, and I moved export() up into PeftPolicyMixin (cooperative super(), no manual override needed per policy anymore). That fully covers your export side concern.

For _inject_lora(): rather than "trains nothing," a missed call actually falls back to full fine-tuning silently. So I added PeftPolicyMixin.on_fit_start() that raises if use_lora is set but no adapters were injected by the time training starts. Cheap, no base-class changes, and it catches this for any future policy that mixes in the class.

Your _on_model_built()/template-method idea is good but bigger: there's no single _initialize_model hook today, it has 3 different signatures across ACT/Groot/Pi0/Pi05/SmolVLA and two call sites each (eager __init__, lazy setup()). Worth doing, but as its own PR so SnapFlow can share it too rather than bundling it here.

library/src/physicalai/policies/pi05/policy.py
Daankrol: Confirmed both catches, it's actually worse than described: Pi0's default LoRA targets never touch PaliGemma at all, so tune_paligemma=True plus lora_enabled=True doesn't just get overridden, it's silently dropped. Went with a hard fail rather than silent override or auto-narrowing the target regex (that trades one silent surprise for another). Added _PEFT_EXCLUSIVE_FLAGS on PeftConfigMixin, and __post_init__ now raises a ValueError if lora_enabled=True is combined with a non-default tune_*/freeze_* value, pointing at lora_target_modules as the right knob instead. Documented the interaction on both configs too.

library/src/physicalai/policies/pi0/config.py
Daankrol: yeah, see my comment above. Let me know what you think

library/src/physicalai/policies/peft/policy.py
Daankrol: Yeah but I remember now that I tested on an H100, which is not a realistic setting. I'll check to see if we can do it without copy. Will get back on this.

library/src/physicalai/policies/pi05/policy.py
Daankrol: Done, exactly as sketched. PeftPolicyMixin.export() now does the merge/swap and calls super().export(). Removed the two duplicated overrides from pi0/policy.py and pi05/policy.py. MRO for both stays PeftPolicyMixin -> ... -> ExportablePolicyMixin -> Policy, so Pi0.export/Pi05.export resolve straight to the mixin now.

MarkRedeman (approved):

PR #971, opened 2026-08-15
chore(deps): bump physicalai to 9be95d88

maxxgx:
## Summary

Stacked on #967 (`max/runtime-4-runtime-owner`). 

Bump `physicalai` git rev to `9be95d88d5b9f8342f6952d8ec288fb108fa416e` and refresh `uv.lock` in library, backend, and plugin.

## Stack

```
main
 └── #938 teleop
      └── #952 zenoh transport
           └── #967 runtime owner
                └── this PR (max/runtime-5-physicalai-bump)
```

PR #992, opened 2026-08-18
feat(runtime): run joints, inference, recording, and export on the session

maxxgx:
The remaining runtime work after #938 / #952 / #967, previously split as #971, #976, #984, #987, #992, and #996. Folded here so it can land as one PR against `main`.

---

### #971 — bump physicalai to 9be95d88

Bump `physicalai` git rev to `9be95d88d5b9f8342f6952d8ec288fb108fa416e` and refresh `uv.lock` in library, backend, and plugin.

---

### #976 — stream joints without a runtime session

The robot detail page spawned a full `rt-` session to display numbers it never writes. It collided with the environment form. Opening both on the same arm produced HTTP 423 for two pages that are not in conflict.

- A read-only websocket at `/api/projects/{project_id}/robots/{robot_id}/observations/ws`.
- It attaches a `SharedRobot` subscriber (the same owner a live session already uses) and streams named `<joint>.pos` features. No `RobotRuntime`, no `SessionNameLock`, no `session_worker`.
- The detail page Connect switch opens that socket. Restart is gone from this page; it stays on the environment form, which still uses `useJointState`.

A read-only attach to avoid `423`: `SharedRobot` is multi-subscriber by design.

Tested:

- [x] Connect on the robot detail page
- [x] Teleop still works
- [x] Detail page + environment form on the same robot at once: both work

---

### #984 — hold cameras on the session and split identity

The runtime session held no cameras, and any camera-set change looked like a different session (HTTP 423). The environment form and a later record client could not share a child: one needed frames, the other did not. Inference and recording need those frames on the session before they can use them.

- One control websocket: `/api/projects/{project_id}/runtime/ws`. The old `/robots/ws` route is gone.
- Handshake carries `follower_id`, optional `leader_id`, and optional `camera_ids`, all resolved against the project.
- Session identity is robot + leader + fps. Camera keys live in `/metadata` separately. A client that needs more cameras restarts the session; a subset attaches.
- The child instantiates, connects, and tears down the document's cameras (`validate_on_connect=True`, `overwrite_settings=False`).
- The environment form sends those camera ids on the handshake. Adding a camera does not restart the live preview.
- The socket URL includes `follower_id` so two follower panels do not share one handshake. Restart clients ignore events until they attach to the new generation.

Limits:

- Editing a camera's resolution under a live session is ignored until the next connect.
- A camera restart displaces whoever was attached; they rejoin the superset.

Tested:

- [x] Teleoperate from the environment form with cameras in the environment
- [x] Add a camera while the preview is live: robot panel keeps streaming
- [x] Refresh the environment form: reattaches to the same child
- [x] Client needing extra cameras restarts; subset attaches
- [x] Camera publisher at the wrong resolution fails connect (does not silently read)
- [ ] Change the leader → 423 and Restart works

---

### #987 — run policy inference on the session

The inference page still used the old model-worker path. That path had a few bugs: the task text never reached the model (VLAs need it), image colors were swapped, and the joint stream froze for seconds while a model loaded. The session already holds the robot and cameras. This moves Play onto that session and deletes the old stack.

- Loading a model happens on a background thread. The arm keeps ticking. If the model’s cameras do not match the environment, you get a clear error at load.
- Pressing Play resets leftover actions first, so the arm moves from where it is, not from an old chunk. The first Play inference also runs in the background, so the joint stream does not block.
- The inference page uses `/runtime/ws` and the modes `hold`, `teleop`, and `policy`. Failed Play/load shows an error. Camera panels reuse the existing live preview widget.
- The old `ModelWorker` / poller / mixer stack is gone.

Limits:

- First tick after opening a session still pays process startup. We did not add a warm process pool. Revisit if load-to-first-action is slower than ~10 seconds.
- Hardware checks and that timing median were **not run**.

Tested:

- [ ] Run a known ACT checkpoint on the inference page
- [x] Run a VLA checkpoint (Pi0, Pi0.5, or SmolVLA)
- [x] Load a model and watch the joint stream: no stutter during load
- [ ] Rename a camera, then load a model trained before the rename: error at load naming both sets
- [ ] Load-to-first-action median over five real runs (threshold 10s)

---

### #992 (this PR, original) — record datasets on the session

The record page still used its own worker and websocket. That old path saved an action computed at a different moment than the robot actually moved, and copying a large dataset froze the arm for seconds.

- Loading a dataset, saving an episode, and discarding one run in the background. The arm keeps moving while a big copy or video encode happens. Starting an episode is still just flipping a flag.
- Frames are written only while you are recording, and not while the arm is holding still. What gets saved is the action the robot actually sent.
- Closing the browser mid-episode does not shut the session down. Stopping the session, or timing out after a save, still writes the dataset back to disk.
- Two projects cannot open the same camera at different resolutions. The error names the project that has it. Watching a preview during recording is fine; changing the resolution is not.
- The record page talks to `/runtime/ws` and uses the same modes as inference: `hold`, `teleop`, `policy`. The episode list updates when a save finishes, not when the tab closes.
- The old record worker, environment glue, and record websocket are gone.

Limits:

- Recording still copies the dataset twice — once into a cache when you start, once back when you stop. Same as today. A slow copy no longer freezes the arm; it just waits up to 60 seconds.
- Camera locks live only in API memory. They only need to stop two sessions in the same process fighting over resolution. If the API restarts mid-recording, the locks are gone even though the robot session is still running. Robot locks live on disk, because a session can keep moving the arm after the API dies.
- Stopping a session does not wait for a save that is still encoding video. We wait up to 60 seconds for queued work; we do not wait for the dataset library itself.
- If you walk away while recording, the session stays up until someone comes back or stops it, so the idle timer cannot throw away an open episode. After you save, the idle timer works as usual.

Hardware checks were **not run**.

Tested:

- [x] Record with two cameras and teleop. Save three episodes, discard one, save one more. Four episodes on disk and in the UI
- [x] Train on that dataset
- [x] Run the resulting model on the inference page in the same environment; the frames it sees match what was recorded
- [ ] Close the browser mid-episode, wait past the idle timeout: session still alive, arm holding still. Reopen and save
- [ ] Stop the session mid-recording: episodes you already saved are still there
- [ ] Open the camera preview at a different resolution while recording: you can watch, the recording resolution does not change
- [ ] Start a second session in another project on the same camera at a different resolution: rejected, naming the first project

---

### #996 — export a runnable inference bundle

Studio can run a trained model, but you could not take that same recipe to another machine. Teleop already has a YAML export. This packages inference: the PolicySource the session uses, plus the export weights, as a zip you run with `physicalai run`.

- Studio and the zip build the same PolicySource: `AsyncExecution(request_threshold=0.5)` and `ChunkedActionQueue` with `LerpSmoother`. PolicyLoader uses that fragment so the two cannot drift.
- `GET /api/models/{id}/exports/{backend}/download` stays weights-only. Pass `environment_id` and `device` and the zip also has `runtime.yaml` and a README, with weights under `exports/<backend>/`.
- Inference page: one download for the current run.
- Formats page: runtime export dialog on the OpenVINO card (environment, device, task). Weights Download is unchanged.
- Empty task is omitted. A non-empty task is baked in. `StudioActionSource` is not in the zip. Teleop `runtime-config` is unchanged.

Limits:

- Execution is `AsyncExecution` only. No picker, and no three-config zip.
- Device paths that are not by-id are flagged `CHANGE_ME` in the yaml and README.
- Hardware `physicalai run` from the extracted zip was **not run**.
- Store fps on the environment, not only in the session, is still not done.

Tested:

- [x] Inference page export: extract, fix `CHANGE_ME`, `physicalai run --config runtime.yaml --run.duration_s=60`, arm moves
- [ ] Formats OpenVINO runtime export dialog with environment, device, and task
- [x] OpenVINO Download without those params is still weights-only
- [ ] Empty task omitted from yaml; non-empty task present


maxxgx:
will address copilot comments in followup

RHeckerIntel (dismissed):

MarkRedeman (approved):

Issue #961, opened 2026-08-13
Refine models table page


MarkRedeman:
Apply the same styles to the models' table as we've set to the remote training / compute page.

Also consider looking into extending the model information with:
- Amount of epochs / steps used for training
- Dataset used for training (showing the amount of episodes, and total video length would be useful)
- Environment / robots used by the dataset

Issue #895, opened 2026-08-04
Extend model training options


MarkRedeman:
We'd like to extend the current model training dialog with additional options:
1. Train for a set number of epochs
2. Configure feature mapping (camera mapping for SmolVLA, Pi05)
3. Store a snapshot of the models' weights (#894 
4. Inform the user about pros & cons for setting precision, batch size, compile model. Guide the user on configuring these based on the selected model and training device
5. Enable or disable exported formats (PyTorch, OpenVINO, ONNX, ExecuTorch)

We can enable this UX by splitting up the train model dialog into multiple steps as sketched out below,

| **Policy selection (step 1)** | **Feature mapping (step 2, optoinal)** |
|:------------:|:-----------:|
| <img width="1043" height="703" alt="Image" src="https://github.com/user-attachments/assets/680e163b-fa87-432f-a947-4916e6197706" />           |    <img width="1009" height="803" alt="Image" src="https://github.com/user-attachments/assets/e40b7c03-585d-4d68-9a85-3ab7c22ea1e7" />       |
| **Training options (step 3)**   | **Export and optimization (step 4)**  |
|   <img width="1009" height="764" alt="Image" src="https://github.com/user-attachments/assets/1ae6ed40-062f-4c9c-ac78-c14bb5cb540b" />         |  <img width="1009" height="803" alt="Image" src="https://github.com/user-attachments/assets/e624fa09-bc2d-4f31-b8e7-15e4a01a850c" />         |

We need extra backend changes for the following:
- Feature mapping step
- Saving snapshots
- Export formats & optimization options

---

### Feature mapping

We need to detect what the input features are for the selected model. Since we only support 3 policies at the moment it would be fine to hard code this, though coming up with a generic approach would be nice as well.
If we detect that the environment that's selected in the first step is fully compatible with the model (e.g. the environment has a camera with the name "Overview" and and another with "Wrist", where "Overview" matches the input feature of the model), then we can skip this step and apply the mapping automatically.

### Saving snapshots

Originally we were planning to use this feature so that the user can export multiple versions of their optimized models while training. For instance if you'd train an Pi05 model for 10 epochs, then you may want to save OpenVINO exports of it on epoch 5, 6, 7, 8, 9 and 10, and then compare each of the models either during training of after training is done.
[...]

PR #1010, opened 2026-08-19
feat(ui, models): Display more model metrics

dwesolow:
## Description

image
## Related Issues

Part 3 of #961


MarkRedeman:
Few quick notes (most already applicable to what we have on main, but would be nice if we can improve this here):
1. Let's make all graphs x-axis start at `step = 0`, even if the data doesn't contain this point
2. Let's change the x scale so that the values are rounded, so instead of 49, 999, 1949 etc we would have 0, 1000, 2000
3. Change the graph colors: training loss in green, validation loss in red
4. Add a tooltip on hover showing the exact x and y values
5. (follow up?) when hovering inside of "training loss" show a vertical line in validation loss and learning rate and a tooltip of the graphs values at that point.
6. (follow up?) allow users to select an area in the chart and zoom into that, double click to reset zoom and/or add a button to reset zoom
7. (later, need to experiment with this) I think it could be nice if we show a vertical line or some symbol at every "epoch" tat's finished
8. (later, need to experiment with this) with Albert's fix from #1002 we no longer save the latest checkpoint, instead we store the checkpoint associated to the epoch where validation loss was lowest. In #895 we also want to allow users to store `n` checkpoints, in this case we can add symbols/anontations in our graphs to highlight those points

dwesolow:
> Few quick notes (most already applicable to what we have on main, but would be nice if we can improve this here):
> 
> 1. Let's make all graphs x-axis start at `step = 0`, even if the data doesn't contain this point
> 2. Let's change the x scale so that the values are rounded, so instead of 49, 999, 1949 etc we would have 0, 1000, 2000
> 3. Change the graph colors: training loss in green, validation loss in red
> 4. Add a tooltip on hover showing the exact x and y values
> 5. (follow up?) when hovering inside of "training loss" show a vertical line in validation loss and learning rate and a tooltip of the graphs values at that point.
> 6. (follow up?) allow users to select an area in the chart and zoom into that, double click to reset zoom and/or add a button to reset zoom
> 7. (later, need to experiment with this) I think it could be nice if we show a vertical line or some symbol at every "epoch" tat's finished
> 8. (later, need to experiment with this) with Albert's fix from [fix(training): Keep best checkpoint instead of last #1002](https://github.com/open-edge-platform/physical-ai-studio/pull/1002) we no longer save the latest checkpoint, instead we store the checkpoint associated to the epoch where validation loss was lowest. In [Extend model training options #895](https://github.com/open-edge-platform/physical-ai-studio/issues/895) we also want to allow users to store `n` checkpoints, in this case we can add symbols/anontations in our graphs to highlight those points

Thanks for the feedback! In this PR points 1-4 are resolved.

MarkRedeman (dismissed):

MarkRedeman (approved):

PR #1012, opened 2026-08-19
feat(backend): thread remote_server_id through training jobs

AlbertvanHouten:
The job model only understood "local" or "remote URL". SSH training needs a target of its own so a job can name a server, and that has to land without disturbing the two existing paths.

## What changed

- `TrainingTarget.SSH` and `TrainJobPayload.remote_server_id`, with the validator keeping local, direct-URL, and SSH mutually exclusive.
- `TrainingWorker._target_key` returns `ssh:<remote_server_id>`; it previously collapsed every SSH job on every server onto the single key `remote:None`.
- Reattach: an SSH job with a persisted `remote_job_id` requeues to `PENDING` on restart, like a direct-URL job.
- Submission validation in `JobService.submit_train_job`: server exists, last preflight healthy, alias still resolves.
- Versioned per-target phase table (`training_backends/phase.py`) plus a `report_phase` helper.
- `RemoteTrainingBackend` accepts an injected device; `get_training_backend` guards SSH payloads so they fail loudly instead of silently training locally.

## Design notes

- **Overloading `REMOTE` was rejected**: `get_training_backend` raises when `remote_trainer_url is None`, and the worker's `reattaching` check would misfire.
- **No new `JobStatus` member** — a GPU-busy wait is a `waiting` phase state in `extra_info["phase"]` and the job stays `pending`. A new member would break every generated UI consumer.
- Local and direct-URL keep `SNAPSHOT_UPLOAD_PROGRESS = 10` / `TRAINING_PROGRESS_END = 95` and their existing assertions.

## Out of scope

No SSH transport, tunnels, or Docker operations. `get_training_backend` still doesn't select an SSH backend — that's #1014.

---

**Plan:** [PR 6 — Training job contract and progress framework](https://github.com/open-edge-platform/physical-ai-studio/blob/main/application/docs/feature_plans/remote-ssh-trainer-pr-plan.md#pr-6--training-job-contract-and-progress-framework) · [backend plan](https://github.com/open-edge-platform/physical-ai-studio/blob/main/application/docs/feature_plans/remote-ssh-trainer-plan.md) · [full PR plan](https://github.com/open-edge-platform/physical-ai-studio/blob/main/application/docs/feature_plans/remote-ssh-trainer-pr-plan.md)

### DIFF
--- a/application/backend/src/api/dependencies.py
+++ b/application/backend/src/api/dependencies.py
@@ -30,7 +30,6 @@ from services.environment_service import EnvironmentService
 from services.event_processor import EventProcessor
 from services.job_service import JobService
 from services.log_service import LogService
-from services.remote_server_service import RemoteServerService
 from services.robot_catalog_service import RobotCatalogService
 from services.snapshot_service import SnapshotService
 from services.system_service import SystemService
--- a/application/backend/src/api/dependencies.py
+++ b/application/backend/src/api/dependencies.py
@@ -20,6 +20,7 @@ from services import (
     ProjectCameraService,
     ProjectService,
     ProjectThumbnailService,
+    RemoteServerService,
     RemoteTrainerService,
     RobotService,
 )
@@ -214,7 +215,7 @@ ModelDownloadServiceDep = Annotated[ModelDownloadService, Depends(get_model_down
 
 def get_job_service(session: AsyncSessionDep) -> JobService:
     """Provides a JobService instance for managing jobs."""
-    return JobService(session, RemoteTrainerService(session))
+    return JobService(session, RemoteTrainerService(session), RemoteServerService(session))
 
 
 JobServiceDep = Annotated[JobService, Depends(get_job_service)]
--- a/application/backend/src/services/job_service.py
+++ b/application/backend/src/services/job_service.py
@@ -5,25 +5,14 @@ from sqlalchemy.exc import IntegrityError
 from sqlalchemy.ext.asyncio import AsyncSession
 
 from db.schema import JobDB
-from exceptions import (
-    DuplicateJobException,
-    RemoteResumeUnsupportedError,
-    RemoteServerAliasNotFoundError,
-    RemoteServerNotReadyError,
-    ResourceInUseError,
-    ResourceNotFoundError,
-    ResourceType,
-    UnsupportedDeviceError,
-)
+from exceptions import DuplicateJobException, ResourceInUseError, ResourceNotFoundError, ResourceType
 from repositories import JobRepository
 from schemas import Job
 from schemas.base_job import JobStatus, JobType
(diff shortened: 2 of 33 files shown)


application/backend/src/services/job_service.py
```diff
@@ -60,6 +71,27 @@ async def submit_train_job(self, payload: TrainJobPayload) -> Job:
                 | {"remote_trainer_url": str(remote_trainer.url), "remote_trainer_name": remote_trainer.name}
             )
 
+        if payload.training_target is TrainingTarget.SSH:
```
MarkRedeman: This check, and the `payload.training_target is TrainingTarget.LOCAL` check after this feel like something that might become simpler if we apply a bit of polymorphism to the design, e.g. have the job service dispatch to a another service that's specific for local, remote or ssh trainers.

application/backend/src/workers/training_worker.py
```diff
@@ -87,17 +87,29 @@ def _on_done(completed_task: asyncio.Task[None]) -> None:
 
     @staticmethod
     def _target_key(payload: TrainJobPayload) -> str:
-        """Return the exclusive execution target for a training job."""
+        """Return the exclusive execution target for a training job.
```
MarkRedeman, line 91: Same polymorphism related comment here, given that we may also add other dedicated runners like "AWS", perhaps after these prs get merged we generalize this?

application/backend/src/api/dependencies.py
```diff
@@ -214,7 +214,7 @@ def get_model_download_service() -> ModelDownloadService:
 
 def get_job_service(session: AsyncSessionDep) -> JobService:
     """Provides a JobService instance for managing jobs."""
-    return JobService(session, RemoteTrainerService(session))
+    return JobService(session, RemoteTrainerService(session), RemoteServerService(session))
```
MarkRedeman, line 217: Perhaps here instead of giving access to both `RemoteTrainerService` and `RemoteServerService` we can find an generalization that also adds a `LocalTrainerService`, and we let the job service use the dedicated service depending on the job type?

MarkRedeman (dismissed):
It's not entirely clear to me why we need to add the `remote_server_id` now and not when we were adding the `REMOTE` trainers.
In a follow up I would be interested to know if some of this could be generalized a bit?

application/backend/src/services/job_service.py
AlbertvanHouten: fixed

application/backend/src/api/dependencies.py
AlbertvanHouten: I am going to leave this open for now as we don't have a LocalTrainerService yet. Lets address this at a later time/when more training targets land.

MarkRedeman (dismissed):

MarkRedeman (approved):

PR #1013, opened 2026-08-19
feat(backend): SSH Docker provisioning service

AlbertvanHouten:
Training on a remote host means starting a container there, which is effectively root-level access to someone's machine. Image identity, container privileges, and teardown therefore have to be guaranteed rather than best-effort — this is the largest security surface in the stack.

## What changed

- **`services/ssh/docker_ops.py`** — resolves the device-specific `protocol-<N>` image to an immutable digest (no fallback tag); `cosign verify`, failing closed; pre-pull `library-version` range check off the manifest; GPU-busy exponential backoff with a give-up timeout; disk re-check against the job's real snapshot size; least-privilege `docker run` (loopback-only port, non-root, dropped capabilities, no `--privileged`, read-only rootfs); list/stop/remove for the orphan sweep.
- **`services/ssh/tunnel.py`** — local-forward tunnel with a reconnect watchdog and bounded retry budget, so a dropped tunnel resumes against the still-running container rather than failing the job.
- **`services/ssh/provisioning.py`** — `SshProvisioningService` orchestrating provision / reattach / teardown / `sweep_orphans`, persisting state at each step and tearing down whatever already started on any failure.
- **`core/backend_instance.py`** — stable per-installation id used as an ownership label, so the sweep never touches a container owned by another Studio installation on the same server.
- **`trainer/schemas.py`, `Dockerfile.trainer`** — `HealthInfo` reports `library_version`; both targets carry the corresponding OCI label.
- **`exceptions.py`, `settings.py`** — actionable error types and SSH settings.

## Design notes

- **No fallback tag.** `trainer-images.yml` publishes no `latest`, and `main` tracks the newest build regardless of protocol — exactly the stale, wire-incompatible image to avoid. An unresolvable tag fails the job.
- **library-version**: older than Studio's → warning; equal/newer → proceed; a `policy` minimum unmet → fail *before* the pull. `/health` re-check is defense-in-depth; a disagreement fails the job.
- Launch by digest only; trainer bound to remote `127.0.0.1`; the sweep requires all management labels **and** a matching `backend_instance_id` **and** no reattach claim.

## Out of scope

Worker wiring is #1014 — this module is usable standalone against a `JobProvisioningRepository` and a `RemoteServer`.

---

**Plan:** [PR 7 — SSH Docker provisioning service](https://github.com/open-edge-platform/physical-ai-studio/blob/main/application/docs/feature_plans/remote-ssh-trainer-pr-plan.md#pr-7--ssh-docker-provisioning-service) · [backend plan](https://github.com/open-edge-platform/physical-ai-studio/blob/main/application/docs/feature_plans/remote-ssh-trainer-plan.md) · [full PR plan](https://github.com/open-edge-platform/physical-ai-studio/blob/main/application/docs/feature_plans/remote-ssh-trainer-pr-plan.md)


application/backend/src/services/ssh/provisioning.py
```diff
@@ -0,0 +1,438 @@
...
+    from settings import Settings
+
+# Port the trainer image listens on inside the container. This launch path
+# never sets `TRAINER_PORT`, so this must match `TrainerSettings.port`'s
+# default and the Dockerfile's `EXPOSE`.
+TRAINER_CONTAINER_PORT = 8001
+
+
+@dataclass
+class ProvisionedTrainer:
```
maxxgx: Should it be a frozen dataclass? What if `base_url` changes, eg. the tunnel port part of the url?

application/backend/src/core/backend_instance.py
```diff
@@ -0,0 +1,95 @@
...
+
+from settings import get_settings
+
+_INSTANCE_ID_FILENAME = "backend_instance_id"
+
+_cached_id: str | None = None
+_lock = Lock()
+
+
+def _instance_id_path() -> Path:
```
MarkRedeman: In #1030 I'm also adding an `instance_id`, but with a different meaning: it tracks the instance of a server and gets refreshed whenever the server restarts.

I guess we should rename either of these to avoid naming confusion?

maxxgx: `instance_id` is also used by the `RuntimeZenohServer` to differentiate events, though, it's not a user-facing concept

application/backend/src/services/ssh/provisioning.py
AlbertvanHouten: Changed to frozen now

application/backend/src/core/backend_instance.py
AlbertvanHouten: renamed to backend_instance_id

maxxgx (approved):

ivanzati (approved):
Thank you, Albert!

PR #1015, opened 2026-08-19
feat(ssh-trainer): startup reattach recovery and orphan sweep wiring

AlbertvanHouten:
A training run can outlive the Studio process. If the backend restarts mid-job the container on the remote host keeps running, so Studio has to find and reclaim it rather than orphan it, kill it, or start a second one alongside it.

## What changed

Recovery runs from `TrainingWorker.setup()` **before** the generic orphan-job abort, so SSH jobs get container-aware recovery first.

- **`verify_reattach`** classifies a persisted job's container without holding a tunnel open — running, owned by this `backend_instance_id`, matching the persisted digest — then re-verifies `/health` through a throwaway tunnel. Every failure is a distinct `ReattachFailureReason`, so "couldn't verify" is never conflated with "verified and wrong".
- **`docker_ops`** gains `inspect_container` and an `IMAGE_DIGEST_LABEL` recorded at launch, so drift is detectable without a second registry call.
- **`recover_ssh_jobs`** runs one startup pass: reconcile rows left by a crashed teardown, verify each non-terminal job's container, then sweep orphans per server **excluding everything just confirmed or left pending** — so a container about to be reattached to is never swept out from under it.
- **`JobProvisioningRepository.list_stale()`** backs the reconcile step.

## Design notes

A host-key failure during reattach fails closed and does **not** tear down the container. When SSH identity or container ownership can't be established, recovery performs no destructive cleanup. Containers owned by another Studio instance are never touched.

Tests cover every `ReattachFailureReason`, the orchestration (reattach-before-sweep, exclusion, stale-row reconciliation, per-server failure isolation), a real-DB `list_active`/`list_stale` test, and the worker wiring.

---

**Plan:** [PR 9 — Recovery, operational hardening, and observability](https://github.com/open-edge-platform/physical-ai-studio/blob/main/application/docs/feature_plans/remote-ssh-trainer-pr-plan.md#pr-9--recovery-operational-hardening-and-observability) · [backend plan](https://github.com/open-edge-platform/physical-ai-studio/blob/main/application/docs/feature_plans/remote-ssh-trainer-plan.md) · [full PR plan](https://github.com/open-edge-platform/physical-ai-studio/blob/main/application/docs/feature_plans/remote-ssh-trainer-pr-plan.md)

### DIFF
--- a/application/backend/src/services/ssh/recovery.py
+++ b/application/backend/src/services/ssh/recovery.py
@@ -46,6 +46,7 @@ if TYPE_CHECKING:
     from uuid import UUID
 
     from repositories.job_provisioning_repo import JobProvisioningRepository
+    from schemas.job_provisioning import JobProvisioning
     from services.job_service import JobService
     from services.remote_server_service import RemoteServerService
     from settings import Settings
@@ -75,6 +76,18 @@ _TRANSIENT_REASONS = frozenset(
         ReattachFailureReason.INSPECTION_FAILED,
     }
 )
+
+
+@dataclass(frozen=True, slots=True)
+class _RowVerdict:
+    """What one active row's processing decided, for the caller to tally and act on."""
+
+    outcome: str  # "confirmed" | "transient" | "failed"
+    # Server whose active set this job's id should be added to, so the
+    # per-server sweep does not reclaim its container. None only when the
+    # job was failed outright (its container, if any, is left for the sweep).
+    active_on_server: UUID | None
+    failure_reason: str | None = None
 
 
 @dataclass(frozen=True, slots=True)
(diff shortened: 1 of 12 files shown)


application/backend/src/services/ssh/recovery.py
```diff
@@ -0,0 +1,239 @@
...
+                )
+                failed += 1
+                failures_by_reason["server_not_registered"] += 1
+                continue
+
+            outcome = await provisioning_service.verify_reattach(row, server)
+
+            if outcome.ok:
+                logger.info("Confirmed reattach for job {} on server '{}'", row.job_id, server.name)
+                active_job_ids_by_server[server.id].add(row.job_id)
```
maxxgx: Don't you need to update the job status to pending for Studio to pick them up again?

application/backend/src/services/ssh/recovery.py
```diff
@@ -0,0 +1,239 @@
...
+
+            if outcome.reason in _TRANSIENT_REASONS:
+                logger.warning(
+                    "Reattach check for job {} on server '{}' was inconclusive ({}); leaving pending for retry: {}",
+                    row.job_id,
+                    server.name,
+                    outcome.reason,
+                    outcome.detail,
+                )
+                active_job_ids_by_server[server.id].add(row.job_id)
```
maxxgx: same as above

application/backend/src/services/ssh/recovery.py
AlbertvanHouten: Yes you are right. Fixed.

maxxgx (approved):

PR #1016, opened 2026-08-19
feat(ssh-trainer): backend enablement switch and fail-closed rollout gating

AlbertvanHouten:
## What changed

- `Settings.ssh_remote_trainer_enabled` (off by default) and `SSH_REMOTE_TRAINER_ENABLED` in `.env.example`.
- `core.security.ssh_network_exposure` with `is_loopback_host()`, `SshFeatureAvailability`, and a cached `get_ssh_feature_availability()` that fails the feature closed when it's enabled but bound to a non-loopback address.
- **Gating at three layers** — route (`require_ssh_feature_active`), submission (`JobService.submit_train_job`), and dispatch (`get_training_backend`) — so a job accepted while the feature was active can't silently run after a restart changes the config or bind address.
- Unauthenticated `GET /api/remote-servers/feature-status`, so the UI can explain unavailability without a circular dependency on the feature.
- `SshFeatureDisabledError` (503) whose message never names an alias, container, or other server detail, since it can reach an unauthenticated caller.
- `cli.serve.start_server` reconciles `Settings.host`/`.port` with the real uvicorn bind args and clears both caches; `lifespan` logs a `critical` warning and records `app.state.ssh_feature_availability` when enabled but exposed.
- `docs/ssh-remote-trainer.md`, linked from the backend README.

## Design notes

The check inspects the **actual bind configuration at process startup**, not a config value claiming loopback, so a UI-only toggle can't bypass it. This is defense-in-depth on top of — not a replacement for — the "not enabled without an auth model" policy.

---

**Plan:** [PR 10 — Backend enablement switch, documentation, and staged rollout](https://github.com/open-edge-platform/physical-ai-studio/blob/main/application/docs/feature_plans/remote-ssh-trainer-pr-plan.md#pr-10--backend-enablement-switch-documentation-and-staged-rollout) · [backend plan](https://github.com/open-edge-platform/physical-ai-studio/blob/main/application/docs/feature_plans/remote-ssh-trainer-plan.md) · [full PR plan](https://github.com/open-edge-platform/physical-ai-studio/blob/main/application/docs/feature_plans/remote-ssh-trainer-pr-plan.md)


application/backend/docs/explanation/ssh-remote-trainer.md
maxxgx, line 1: Can we move this doc to `application/backend/docs/explanation/`? We have this structure for library and framework. The other docs in the same folder are numbered and serve as guide so it feels weird to have this doc here

AlbertvanHouten: moved

maxxgx (approved):

AlexanderBarabanov (approved):

PR #1017, opened 2026-08-19
fix(ssh): close error-handling gaps and harden signature verification in remote-training recovery path

AlbertvanHouten:
Backend work for SSH-provisioned remote training: signature verification no longer needs `cosign` on the remote host, a server's live status/GPU-wait state is now real instead of placeholder, the SSH feature's on/off switch is gone (it's always active, fail-closed on network exposure), and several error-handling gaps and a Hugging Face token leak are closed.

## Signature verification: `cosign` binary -> `sigstore` PyPI package

Verification no longer runs a remote `cosign` command over SSH at all - it runs entirely in this backend process:

- New `services.ssh.oci_registry`: a minimal OCI Distribution client that resolves a tag to its digest and fetches a Sigstore-bundle referrer artifact for that digest (GHCR doesn't implement the OCI 1.1 Referrers API, so this uses the `sha256-<hex>`-tag fallback convention `cosign`/Buildx also use).
- New `services.ssh.sigstore_verify`: verifies the bundle's certificate chain and Rekor inclusion proof via `sigstore-python` against Sigstore's public-good Fulcio/Rekor/TUF infrastructure, checks the certificate identity/issuer (mirroring `cosign verify --certificate-identity-regexp`/`--certificate-oidc-issuer`), and confirms the signed in-toto statement is a plain signature (not an SBOM/provenance attestation) over the exact image digest. Raises `SignatureUnavailableError` (registry/Sigstore infra unreachable) separately from `SignatureVerificationError` (verification attempted and failed), so callers can choose which to treat as an escape hatch.
- `docker_ops.verify_image_signature` (fail-closed, blocks provisioning) and `preflight._check_signature` (advisory, Tier 2's non-blocking `IMAGE_SIGNATURE` check) both delegate to it. Removed the `ssh_require_cosign_verification` setting and every remote `cosign version`/`cosign verify` call - a remote trainer server no longer needs `cosign` installed at all.
- Preflight's `IMAGE_SIGNATURE` check now reports `REASON_SIGSTORE_UNAVAILABLE`/`method=sigstore` when the registry or Sigstore's own infrastructure is unreachable, instead of the misleading `tool_missing`/`cosign` it used to report for what is now a completely different failure mode.

## Live status, throttled polling, and first-use verification

- `RemoteServerService.get_status` assembles `/status` end-to-end instead of returning hardcoded placeholders. `get_active_for_server` plus the job's `extra_info.phase.state` report `in_use_by_job_id` and `waiting_for_gpu` - display-only, never a workflow gate.
- The Tier 1 probe behind `/status` is coalesced and throttled per server via `ssh_preflight_throttle_s`, so concurrent pollers share one in-flight dial instead of piling connections onto a server that's actively training. The shared in-flight `Task` (`asyncio.create_task`) is `asyncio.shield`ed so a cancelled caller (e.g. an HTTP request cancelled on client disconnect) can never cancel the probe out from under other concurrent pollers.
- New `RemoteServerService.ensure_verified`: a server whose Tier 2 check has never run (`last_check_status == "unknown"`) is verified automatically, once, the first time it's selected for a job, instead of rejecting the job and sending the user back to click "Test connection" first. A server whose last explicit check *failed* still requires deliberate re-verification - this never retries that automatically. Wired into `training_targets/ssh.py`'s job-submission prepare step.
- Tier 2's per-check detail (image resolution, signature, device probe, protocol) is now persisted (`RemoteServerDB.last_check_checks`, new migration) alongside the existing rolled-up summary, so the verification card can render the last run's detail after a page refresh instead of resetting to "Not verified yet".
- Remote-server preflight failures during create/update now name which check failed and its own detail (e.g. free disk space, the raw Docker/driver error) instead of just the check key and reason code; a hung preflight's `TimeoutError` now maps to the same `SshConnectionError` the rest of this API already uses instead of a bare 500.
- New `GET /api/remote-servers/aliases/{alias}/device-type`: best-effort accelerator autodetection (`nvidia-smi`, then `xpu-smi`/render-node fallback) to prefill the add-target form's device type. Never raises for a driverless or unreachable host - it reports a reason code instead, so the form falls back to asking the user.

## Removed the SSH remote-trainer master switch

The feature is now always active - there is no `SSH_REMOTE_TRAINER_ENABLED` setting or settings-page switch to turn it off. The fail-closed network-exposure check (`core.security.ssh_network_exposure`, run at startup and in three other independent places) is the one enforcement a user cannot opt out of; the lack of an authentication model is instead surfaced as an explicit warning in the UI when a user registers an SSH target. `services.ssh.recovery.recover_ssh_jobs` now also skips startup reattachment entirely (no-op) when the feature is inactive, so a leftover provisioning row never causes a restart to dial SSH or touch a container while failing closed.

## User-editable SSH provisioning settings

New `SshProvisioningSettings` group (connect/command/preflight/image-pull/readiness timeouts, GPU-wait giveup, minimum free disk) exposed via `PATCH /api/settings` under an `ssh` key and surfaced on the Studio settings page, persisted to the settings JSON file. The SSH config/`known_hosts` paths, trainer image registry, and cosign certificate identity/issuer stay environment-only, since the (unauthenticated) settings API must never be able to move what host or image Studio trusts. These settings are explicitly excluded from the environment/`.env` sources (`settings._EnvExclusionSource`) so a value saved through the API can never be silently overridden by a stale environment variable on the next restart.

## Image pull and GPU-busy checks

- `docker_ops.pull_image` now skips a redundant `docker pull` when Tier 2's own background pull (detached via `nohup` so it survives that connection closing) already fetched the digest, waits for an in-flight background pull instead of racing it with a second concurrent pull of identical content, and prunes other locally cached images of the same repository once the needed digest is confirmed present (`prune_stale_images`, best-effort) so a long-lived server doesn't accumulate every digest it has ever pulled.
- The CUDA GPU-busy check switched from compute-process presence (`nvidia-smi --query-compute-apps`) to the same memory-utilization heuristic XPU already used, and both now query device index 0 explicitly (`-i 0` / `-d 0`) - the device the trainer container actually uses - rather than an ambiguous default.

## Hugging Face token scoping

The Hugging Face token no longer travels through anything persisted to disk: `JobStore` caches each job's token in memory only (`stash_secret`/`take_secret`/`discard_secret`), `SubmitJobRequest.hf_token` is `exclude=True` so it's never written to the trainer's SQLite row, and `training.job._hf_token_env` scopes `os.environ["HF_TOKEN"]` to the single training run via a context manager that restores whatever was previously set - jobs can run concurrently in the same process, so a token set for one must never leak into another's environment. `RemoteTrainingBackend.submit_job` now sends the token as a separate top-level `hf_token` field alongside `spec`, rather than folding it into `run_options`. A new shared `resolve_hf_token()` helper (falling back to the legacy `HF_TOKEN` env var) is used by the local backend, the remote/SSH backends, and `api.policies.check_huggingface_access`.

## Error-handling gaps

- `recover_ssh_jobs`' reattach loop no longer lets one job's exception abort recovery - and the orphan sweep - for every other job; the failure is now logged with its full traceback (`logger.exception`) and the job is requeued for a later retry instead.
- Host-key/alias failures from `verify_reattach`'s tunnel-probe connection are now caught and classified instead of escaping unhandled.

## Legacy data compatibility

- `JobMapper.from_schema` backfills `training_target` for jobs persisted before that field existed on `TrainJobPayload` (inferred from the presence of `remote_trainer_id`), and strips now-forbidden remote-only extra keys from legacy local payloads - without mutating the ORM row's payload dict in place, so a read mapping can never mark the entity dirty or risk an unintended flush.

## Docs, logging, and cleanup

- Corrected stale docstrings/docs surfaced in review: `RemoteServerService`'s module docstring now reflects that `get_status()` (unlike create/update) does dial SSH via a throttled probe, `ssh_patch_to_flat` documents falling back to field defaults (not an environment value), and `docs/remote-trainer.md`/`docs/explanation/ssh-remote-trainer.md` describe the always-active feature, the new settings group, and how the Hugging Face token now reaches an SSH-provisioned trainer.
- Quieted chatty third-party loggers (`asyncssh`, `httpx`, `httpcore`) to `WARNING`.
- Removed settings from `application/backend/.env.example` that are no longer necessary.

## Tests and CI

- `tests/integration/test_ssh_containerized_preflight.py` drives the real `SshTransport` and `run_tier1_preflight` against a throwaway Alpine `sshd` container with a generated ed25519 key - no mocks, covering the real handshake/host-key/auth behavior the hermetic suite can't. Marked `integration`/`slow`, skipped when Docker is unavailable.
- `tests/services/ssh/test_sigstore_verify_integration.py` hits the real `ghcr.io` registry and Sigstore's public-good infrastructure to verify this project's actual published trainer image signatures. Added `ghcr.io`, `pkg-containers.githubusercontent.com` (ghcr.io's manifest/blob redirect target), `fulcio.sigstore.dev`, `rekor.sigstore.dev`, `timestamp.sigstore.dev`, and `tuf-repo-cdn.sigstore.dev` to the backend unit-tests job's harden-runner egress allowlist - matching the set already allowlisted for cosign-signing jobs in `docker-publish.yml`/`trainer-images.yml` - so this job stops failing with blocked-egress errors.
- Added `sigstore>=4.0,<5` as a backend dependency; new Alembic migration for `remote_servers.last_check_checks`.

---

**Plan:** [PR plan](https://github.com/open-edge-platform/physical-ai-studio/blob/main/application/docs/feature_plans/remote-ssh-trainer-pr-plan.md)  [backend plan](https://github.com/open-edge-platform/physical-ai-studio/blob/main/application/docs/feature_plans/remote-ssh-trainer-plan.md)


application/backend/src/core/restart.py
```diff
@@ -0,0 +1,27 @@
...
+    """Send this process SIGTERM so the process supervisor restarts it.
+
+    The FastAPI lifespan (`core.lifecycle.lifespan`) stops workers on
+    receiving the signal, then re-executes the process
+    (`core.lifecycle._restart_process`) if
+    `HealthService.plugin_restart_required` was set beforehand. Callers must
+    set that flag first - this function only requests the shutdown half of
+    that sequence.
+    """
+    os.kill(os.getpid(), signal.SIGTERM)
```
MarkRedeman: Probably can be replaced by the restart endpoint that I added for the robot plugins?

AlbertvanHouten: indeed, reused

application/backend/src/services/remote_server_service.py
```diff
@@ -90,3 +113,84 @@ async def record_check_result(self, remote_server_id: UUID, result: PreflightRes
...
+            )
+            _status_checks[server.id] = task
+
+            def _clear_inflight(done: asyncio.Task[PreflightResult], server_id: UUID = server.id) -> None:
+                if _status_checks.get(server_id) is done:
+                    del _status_checks[server_id]
+
+            task.add_done_callback(_clear_inflight)
+
+        result = await task
```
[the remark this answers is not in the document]
AlbertvanHouten: Fixed in b63afb8c7ffaf41e8e2a7d566610ea9d52853d21 — the docstring already documented the shielding intent but the code hadn't caught up; `_tier1_status` now awaits `asyncio.shield(task)` so a cancelled caller can't cancel the shared probe.

application/backend/src/services/remote_server_service.py
```diff
@@ -8,15 +8,38 @@
 dials SSH.
```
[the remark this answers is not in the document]
AlbertvanHouten: Fixed in b63afb8c7ffaf41e8e2a7d566610ea9d52853d21 — updated the module docstring to reflect that `get_status()` dials SSH via a throttled Tier 1 probe.

application/backend/src/services/ssh/recovery.py
```diff
@@ -111,6 +112,83 @@ async def _reconcile_stale_rows(
...
+
+    try:
+        outcome = await provisioning_service.verify_reattach(row, server)
+    except Exception as error:  # one job's failure must not abort recovery of the rest
+        logger.error(
+            "Reattach check for job {} on server '{}' raised unexpectedly; leaving pending for retry: {}",
+            row.job_id,
+            server.name,
+            error,
+        )
```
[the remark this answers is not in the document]
AlbertvanHouten: Fixed in b63afb8c7ffaf41e8e2a7d566610ea9d52853d21 — restored the try/except around `verify_reattach` (it had been dropped in a later refactor) using `logger.exception(...)` so the traceback is preserved, and requeues the job as transient instead of letting the exception propagate.

application/backend/src/services/ssh/sigstore_verify.py
```diff
@@ -0,0 +1,181 @@
...
+
+    def verify(self, cert: Certificate) -> None:
+        """Verify `cert` against the policy. Raises `SigstoreVerificationError` on failure."""
+        san_ext = cert.extensions.get_extension_for_class(SubjectAlternativeName).value
+        candidates = set(san_ext.get_values_for_type(RFC822Name))
+        candidates.update(san_ext.get_values_for_type(UniformResourceIdentifier))
+        if not any(self._pattern.fullmatch(candidate) for candidate in candidates):
+            raise SigstoreVerificationError(
+                f"no certificate SAN matched {self._pattern.pattern!r} (got {sorted(candidates)!r})"
+            )
```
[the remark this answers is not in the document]
AlbertvanHouten: Fixed in b63afb8c7ffaf41e8e2a7d566610ea9d52853d21 — wraps SAN extraction in try/except ExtensionNotFound and re-raises as SigstoreVerificationError.

application/backend/src/services/ssh/sigstore_verify.py
```diff
@@ -0,0 +1,181 @@
...
+    * `predicateType` must be `cosign`'s plain-signature predicate, not an
+      SBOM or provenance attestation - which a registry stores in the
+      identically-shaped Sigstore-bundle wrapper, and which `oci_registry`'s
+      artifact-type filter does not by itself distinguish from a signature.
+    * The statement's subject digest must match the image actually being
+      verified, confirming the referrer lookup found the right artifact.
+    """
+    statement = json.loads(payload)
+    if statement.get("predicateType") != _COSIGN_SIGN_PREDICATE_TYPE:
+        raise SignatureVerificationError(f"unexpected predicate type {statement.get('predicateType')!r}")
```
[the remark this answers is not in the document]
AlbertvanHouten: Fixed in b63afb8c7ffaf41e8e2a7d566610ea9d52853d21 - _check_statement now catches JSON decode/type errors and rejects non-dict statements, raising SignatureVerificationError in both cases.

application/backend/src/api/policies.py
```diff
@@ -26,6 +26,7 @@
     # download. When adding/changing a policy, inspect its `from_pretrained`,
     # `hf_hub_download`, and `Auto*from_pretrained` calls and list every default
     # repository that Studio training needs here.
+    "pi0": (("google/paligemma-3b-pt-224", True),),
```
MarkRedeman: We don't show pi0 in the UI so I think it would be better to omit this

application/backend/src/api/settings.py
```diff
@@ -41,5 +52,11 @@ async def get_user_settings() -> UserSettingsResponse:
 @router.patch("")
 async def update_user_settings(update: SettingsUpdate) -> UserSettingsResponse:
     """Persist the supplied fields and return effective settings."""
-    merge_user_settings(update.model_dump(exclude_unset=True))
+    patch = update.model_dump(exclude_unset=True)
+    if "ssh" in patch:
```
MarkRedeman, line 60: Let's fix this in a follow up.

application/backend/src/core/logging/setup.py
```diff
@@ -26,6 +26,15 @@
 
 global_log_config = LogConfig()
 
+# Chatty at INFO (e.g. one line per SSH channel/HTTP request); quiet to WARNING.
+_NOISY_LOGGERS = ("asyncssh", "httpx", "httpcore")
+
+
+def _quiet_noisy_loggers() -> None:
```
MarkRedeman, line 34: Wouldn't setting them to `WARNING` make it more noisy? I.e. shouldn't it be `logging.DEBUG`?

application/backend/src/core/security/ssh_network_exposure.py
```diff
@@ -9,11 +9,14 @@
 agent, not just the registered servers. It is safe only on a single-user
 localhost workstation.
 
-This module is the runtime enforcement of that boundary, on top of - not a
-replacement for - the operator turning the feature on in the first place
-(`Settings.ssh_remote_trainer_enabled`, off by default) and the documentation
-in `docs/ssh-remote-trainer.md`. Even when explicitly enabled, the feature
-fails closed if the backend is bound to anything but a loopback address.
+The feature is always active - there is no settings-page or environment
```
MarkRedeman: Generally in this pr: let's drop the old references to having used a feature flag. It's an implementation detail tht is no longer relevant.

application/backend/src/training/job.py
```diff
@@ -190,56 +214,55 @@ def run_training_job(
...
-
-    _ensure_checkpoint_exists(trainer, cache_dir)
-    _publish(cache_dir, output_dir)
-
-    export_policy = _export_policy(spec, policy, output_dir)
-    _detach_trainer(export_policy, trainer)
-    del trainer, datamodule, policy
-    _release_memory()
-    _export(export_policy, output_dir, report)
+    with _hf_token_env(spec.run_options.hf_token):
```
MarkRedeman, line 282: Perhaps one thing to look at later: instead of assigning the hf token during the full job, we could do this only around:
```
    with _hf_token_env(spec.run_options.hf_token):
        trainer.fit(model=policy, datamodule=datamodule)
```

application/backend/src/settings.py
```diff
@@ -65,15 +65,144 @@ def empty_token_is_unset(cls, value: str | None) -> str | None:
...
+    """JSON settings source restricted to user-configurable settings."""
 
     def __call__(self) -> dict[str, Any]:
         data: dict[str, Any] = super().__call__()
-        return {key: value for key, value in data.items() if key in _USER_CONFIG_GROUPS}
+        allowed = (*_USER_CONFIG_GROUPS, *_user_config_scalar_keys())
+        return {key: value for key, value in data.items() if key in allowed}
+
+
+class _EnvExclusionSource(PydanticBaseSettingsSource):
```
MarkRedeman, line 199: This part feels a bit complicated. The comment mentions ssh settings, but I don't see that directly in the implementation

MarkRedeman (dismissed):

application/backend/src/core/logging/setup.py
AlbertvanHouten: It now only logs warning level logs for those 3 libs, not raising the lower levels. I have clarified the above comment.

MarkRedeman (approved):

application/backend/src/services/training_backends/__init__.py
AlexanderBarabanov, line 56: As far as I understand, Studio still supports a non-SSH `remote_trainer_url` provided by the user. Since we're now sending `hf_token` to the training backend, if a user configures this feature with `http://...`, the token won't be protected in transit.

Please consider adding validation that the URL requires `https://`, or if `http://` needs to remain supported - add a warning message in the UI and explicitly document this for user awareness.

AlexanderBarabanov (approved):
Thanks, one suggestion, could be addressed in a separate PR.

PR #1019, opened 2026-08-20
feat(plugin): extend studio plugin with ui schema

MarkRedeman:
> **This PR is extracted from #1003**

Extend the `RobotCatalogDefinition` with `category`, `source` and `preview_thumbnail`. These fields will mainly be used by the new robot picker that lists robots with a thumbnail previewing them.

The `source` has the following options: `source: Literal["internal", "first_party", "external"] = "external"`, I'm not entirely sure yet about these values, but the main point is to distinguish plugins that are installed by default in Studio and plugins that we maintain as the physicalai team, these are plugins where we guarantee that they are expected to work.

We also add a new `ui_schema` module, this exposes functions used by plugins when defining fields. An example usage of this is in #1003 where the UI uses these to automatically generate form fields when creating new robots.
As an example suppose a plugin defines,
```python
from physicalai_studio_plugin import robot_payload_ui
from pydantic import BaseModel, ConfigDict, Field


class MyRobotPayload(BaseModel):
    connection_string: str = Field(
        default="",
        title="Connection",
        description="Select a discovered port or enter one manually.",
    )
    serial_number: str = Field(
        default="",
        title="Serial number",
        description="Optional hardware serial identifier.",
    )
    torque_enabled: bool = Field(
        default=True,
        description="Enable motor torque during startup.",
    )

    model_config = ConfigDict(
        json_schema_extra={
            **robot_payload_ui(
                [
                    {
                        "kind": "section",
                        "id": "connection",
                        "title": "Connection",
                        "description": "Use Refresh to rescan ports.",
                        "items": [
                            {
                                "kind": "info",
                                "title": "Before connecting",
                                "text": "Power on the robot and clear the workspace.",
                            },
                            {
                                "kind": "connection",
                                "label": "Robot connection",
                                "device_discovery": True,
                                "identify": True,
                                "manual_entry": True,
                                "bind": {
                                    "connection": "connection_string",
                                    "serial_number": "serial_number",
                                },
                            },
                            {
                                "kind": "info",
                                "text": "USB hubs can change assigned port names after reboot.",
                                "variant": "warning",
                            },
                        ],
                    },
                    {
                        "kind": "section",
                        "id": "behavior",
                        "title": "Behavior",
                        "items": [{"kind": "field", "name": "torque_enabled"}],
                    },
                
            )
        }
    )
```

Then the UI creates a form where the user sees a small info panel, can select a usb entry (serial number or port number), another info and then a new section showing a toggle for enabling/disabling torque.
This quite flexible and in my initial UI implementation I've made this recursive, so the UI can handle something like,
```
class TrossenBimanualPayload(BaseModel):
    """Connection configuration for Trossen bimanual robots."""

    left_payload: TrossenSingleArmPayload
    right_payload: TrossenSingleArmPayload
```

The UI will then read that schema and can show two distinctive form segments for both the left and right arm.

### DIFF
--- a/application/plugin/src/physicalai_studio_plugin/catalog.py
+++ b/application/plugin/src/physicalai_studio_plugin/catalog.py
@@ -70,6 +70,8 @@ class RobotCatalogDefinition(Generic[_PayloadT]):
     type: str
     display_name: str
     role: Literal["follower", "leader"]
+    category: str = "Other"
+    source: Literal["internal", "first_party", "external"] = "external"
     robot_builder: BuildRobotCallable | None = None
     robot_payload: _PayloadModelT | None = None
     asset: RobotAsset | None = None
--- a/application/plugin/src/physicalai_studio_plugin/catalog.py
+++ b/application/plugin/src/physicalai_studio_plugin/catalog.py
@@ -70,8 +70,6 @@ class RobotCatalogDefinition(Generic[_PayloadT]):
     type: str
     display_name: str
     role: Literal["follower", "leader"]
-    category: str = "Other"
-    source: Literal["internal", "first_party", "external"] = "external"
     robot_builder: BuildRobotCallable | None = None
     robot_payload: _PayloadModelT | None = None
     asset: RobotAsset | None = None
--- a/application/backend/src/robots/catalog/so101.py
+++ b/application/backend/src/robots/catalog/so101.py
@@ -6,10 +6,15 @@ from pathlib import Path
 from typing import TYPE_CHECKING, Literal
 
 from loguru import logger
-from physicalai.robot import SO101
-from physicalai.robot.so101 import SO101Calibration, SO101JointCalibration
+from physicalai.robot.so101 import SO101, SO101Calibration, SO101JointCalibration
 from physicalai.robot.so101.constants import TICKS_PER_REVOLUTION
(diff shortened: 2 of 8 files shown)


application/plugin/src/physicalai_studio_plugin/catalog.py
```diff
@@ -70,6 +70,8 @@ class RobotCatalogDefinition(Generic[_PayloadT]):
     type: str
     display_name: str
     role: Literal["follower", "leader"]
+    category: str = "Other"
+    source: Literal["internal", "first_party", "external"] = "external"
```
AlbertvanHouten: Are internal and first_party not basically the same thing?

AlbertvanHouten (dismissed):

application/plugin/src/physicalai_studio_plugin/catalog.py
MarkRedeman: My initial thought was that we would have So101 and WidowX as our "internal", robots that are installed by default.
While first party will be ones that our team is able to support (have hardware access to)

MarkRedeman: Have thought about it  a bit more and think I will remove the source in a follow up pr. At the moment the UI we have doesn't make a good distinction between these sources (+ external plugins can "fake" the source)

maxxgx (approved):

PR #1023, opened 2026-08-20
feat: add unavailable robot response

MarkRedeman:
> This PR is extracted from https://github.com/open-edge-platform/physical-ai-studio/pull/1003

With the new robot plugin setup it will be possible for users to have robots stored in their database that may no longer be available (due to the plugin being uninstalled).
Before the backend would throw a validatoin error from `RobotAdapter.validate_python`, with this PR we are being more lenient: if the robot type is unknown then we return an `UnavailableRobot`.

The UI is updated to show some very basic error messages: I expect that the issue this PR addresses will mainly be seen by developers working on plugins, so for now we won't invest too much time yet on tihs.

| **Dataset recording** | **Dataset viewing** |
|------------|-----------|
|  screenshot-2026-08-20_12-53-21          | screenshot-2026-08-20_12-53-08          |
| **Environment**   | **Robot list**  |
|  screenshot-2026-08-20_12-52-57          |  screenshot-2026-08-20_12-52-44         |

### DIFF
--- a/application/backend/src/api/robots.py
+++ b/application/backend/src/api/robots.py
@@ -4,7 +4,7 @@ from uuid import UUID
 from fastapi import APIRouter, Depends, status
 
 from api.dependencies import get_project_id, get_robot_id, get_robot_service
-from schemas.robot import Robot, RobotWithConnectionState
+from schemas.robot import Robot, RobotWithConnectionState, UnavailableRobot, UnavailableRobotWithConnectionState
 from services import RobotService
 
 router = APIRouter(prefix="/api/projects/{project_id}/robots", tags=["Project Robots"])
@@ -16,7 +16,7 @@ ProjectID = Annotated[UUID, Depends(get_project_id)]
 async def list_project_robots(
     project_id: ProjectID,
     robot_service: Annotated[RobotService, Depends(get_robot_service)],
-) -> list[Robot]:
+) -> list[Robot | UnavailableRobot]:
     """Fetch all robots."""
     return await robot_service.get_robot_list(project_id)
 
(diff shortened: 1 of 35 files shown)


application/backend/src/schemas/robot.py
```diff
@@ -27,6 +30,17 @@
 RobotAdapter: TypeAdapter = _registry.get_robot_adapter()
 
 
+class UnavailableRobot(BaseRobot):
+    """A persisted robot whose catalog plugin is not currently installed."""
+
+    type: str
+    payload: dict[str, Any]
+    unavailable: Literal[True] = True
```
AlbertvanHouten, line 38: Why is this a separate class? Can't we add this unavailable logic to the BaseRobot class?

MarkRedeman: For this I prefer the extra robot as it makes it more explicit: all other robots are also extensions of `BaseRobot` with their `type` being the descriminative type.

Another thing is that it helps make this distinction:
```
Robot = _registry.make_robot_type()

ReadableRobot = Robot | UnavailableRobot
```

where `Robot` is the union of all robots that have been registered (both internally defined and plugins) while `ReadableRobot` also allows this `UnavailalbeRobot`. 
So `Robot` defines one that we can interact with, while `ReadableRobot` is mainly used in our REST API response. If we'd change this then we would need to check if the robot is available in many other places in the backend and plugins.

application/backend/src/api/robots.py
```diff
@@ -16,7 +16,7 @@
 async def list_project_robots(
     project_id: ProjectID,
     robot_service: Annotated[RobotService, Depends(get_robot_service)],
-) -> list[Robot]:
+) -> list[Robot | UnavailableRobot]:
```
AlbertvanHouten: This return type could be just ReadableRobot.

AlbertvanHouten (dismissed):

application/backend/src/api/robots.py
```diff
@@ -45,7 +51,7 @@ async def get_project_robot(
     project_id: Annotated[UUID, Depends(get_project_id)],
     robot_id: Annotated[UUID, Depends(get_robot_id)],
     robot_service: Annotated[RobotService, Depends(get_robot_service)],
-) -> Robot:
+) -> Robot | UnavailableRobot:
```
AlbertvanHouten:
```suggestion
) -> ReadableRobot:
```

RHeckerIntel (approved):

PR #1024, opened 2026-08-20
feat: Upgrade to dictionary based payload

ArendJanKramer:
## Description

This PR updates the Studio/backend camera fingerprint”handling to use JSON-encoded dictionary payloads (to match updated physicalai / camera stack behavior), and refreshes dependency locks and sources across the repo.

Changes:

- Switch camera fingerprints from legacy string IDs to JSON-serialized dict payloads in backend APIs/utilities and adjust tests accordingly.
- Update physicalai dependency source to a new git revision (and update related uv.lock files).
- Update lockfile metadata (e.g., exclude-newer) and bump the root lock’s Python requirement.

## Related Issues
Screenshot from 2026-09-01 13-36-28


application/backend/src/utils/camera_factory.py
```diff
@@ -45,6 +46,14 @@
...
 }
 
+def _get_fingerprint_dict(fingerprint_str: str) -> dict[str, Any]:
+    try:
+        d = json.loads(fingerprint_str)
+        if isinstance(d, dict):
+            return d
+    except ValueError as error:  # Ignore errors, we'll fall back to legacy
+        pass
+    raise ValueError("Unable to parse fingerprint into a dict")
```
maxxgx: How is this falling back to legacy since it raises just after this line?

Adding a log here will be useful

application/backend/src/api/camera.py
```diff
@@ -32,15 +32,19 @@ def _formats_to_response(raw: list[tuple[int, int, int]]) -> list[SupportedCamer
 @cache
 def _query_formats(driver: str, fingerprint: str) -> list[SupportedCameraFormat]:
     """Query real formats from a device via physicalai.capture."""
+    try:
+        f = json.loads(fingerprint)
```
maxxgx: Maybe it's time to rename `fingerprint`? Could be `camera_payload` / `hardware_payload`?

ArendJanKramer: I find fingerprint more descriptive tbh, payload in camera context would suggest the pixel data to me

application/backend/src/schemas/project_camera.py
```diff
@@ -103,7 +103,7 @@ class USBCamera(BaseCamera):
             "example": {
                 "name": "front_camera",
                 "driver": "usb_camera",
-                "fingerprint": "USB\\VID_1234&PID_5678:0",
+                "fingerprint": '{"serial":"12345678AB"}',
```
MarkRedeman, line 106: Can we change the fingerprint to be a proper dict / json object, and possibly add some extra type safety so that in the UI we can check things like `fingerprint.serial` instead of it having to manually parse the string as json?

ArendJanKramer: Not yet, next pr. It will touch a lot of code, as the api spec also changes

application/backend/src/api/camera.py
```diff
@@ -32,19 +32,26 @@ def _formats_to_response(raw: list[tuple[int, int, int]]) -> list[SupportedCamer
 @cache
 def _query_formats(driver: str, fingerprint: str) -> list[SupportedCameraFormat]:
     """Query real formats from a device via physicalai.capture."""
+
+    if driver == "basler":
+        # TODO: Replace with cached Basler hardware discovery once implementated in physicalai.capture
+        return _formats_to_response([(640, 480, 30), (768, 480, 30), (1920, 1200, 30)])
+
+    try:
+        f = json.loads(fingerprint)
```
maxxgx:
```suggestion
        hardware_payload = json.loads(fingerprint)
```
Please don't use single letter variables

application/backend/src/api/camera.py
```diff
@@ -32,19 +32,26 @@ def _formats_to_response(raw: list[tuple[int, int, int]]) -> list[SupportedCamer
 @cache
 def _query_formats(driver: str, fingerprint: str) -> list[SupportedCameraFormat]:
     """Query real formats from a device via physicalai.capture."""
+
+    if driver == "basler":
+        # TODO: Replace with cached Basler hardware discovery once implementated in physicalai.capture
+        return _formats_to_response([(640, 480, 30), (768, 480, 30), (1920, 1200, 30)])
```
maxxgx, line 38: these formats placeholder are fine. The TODO is a bit misleading as basler api allows for any WxH combo within its sensor limitation, and same goes with FPS. 

We could probably generate a list of supported formats from `physicalai.capture`, can you updated the TODO for future reference?

maxxgx:
Can you post a screenshot to show how it looks after removing the fingerprint from the UI?

maxxgx (approved):

PR #1025, opened 2026-08-20
feat: include application footer in app layout

MarkRedeman:
> This PR is extracted from https://github.com/open-edge-platform/physical-ai-studio/pull/1003

Having the footer been shown in the application layout allows users to see training jobs without being inside of a project.

We will later also extend this footer with an option to warn the user that they need to reboot their server after installing a new plugin.

image

### DIFF
--- a/application/ui/src/components/app-footer/app-footer.tsx
+++ b/application/ui/src/components/app-footer/app-footer.tsx
@@ -4,13 +4,10 @@ import { Manifest } from '@geti-ui/ui/icons';
 import { JobStatus } from '../../features/jobs/footer/job-status';
 import { LogsDialog } from '../../features/logs/logs-dialog';
 
-export const AppFooter = ({ gridArea = 'footer', compact = false }: { gridArea?: string; compact?: boolean }) => {
+export const AppFooter = ({ gridArea = 'footer' }: { gridArea?: string }) => {
     return (
         <View
             gridArea={gridArea}
-            height='size-400'
-            minHeight='size-400'
-            maxHeight='size-400'
             borderTopColor={'gray-300'}
             borderTopWidth={'thin'}
             borderBottomColor={'gray-75'}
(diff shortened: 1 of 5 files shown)


application/ui/src/components/app-footer/app-footer.tsx
```diff
@@ -0,0 +1,46 @@
...
+import { JobStatus } from '../../features/jobs/footer/job-status';
+import { LogsDialog } from '../../features/logs/logs-dialog';
+
+export const AppFooter = ({ gridArea = 'footer', compact = false }: { gridArea?: string; compact?: boolean }) => {
+    return (
+        <View
+            gridArea={gridArea}
+            height='size-400'
+            minHeight='size-400'
+            maxHeight='size-400'
```
dwesolow: why height, mix and max are needed if we specify height in the grid?

application/ui/src/components/app-footer/app-footer.tsx
```diff
@@ -0,0 +1,46 @@
...
+            borderTopColor={'gray-300'}
+            borderTopWidth={'thin'}
+            borderBottomColor={'gray-75'}
+            borderBottomWidth={'thin'}
+            backgroundColor={'gray-75'}
+            paddingX='size-100'
+            paddingY='size-0'
+        >
+            <Flex alignItems='center' justifyContent='space-between' gap='size-100' height='100%'>
```
dwesolow: is this flex necessary?

application/ui/src/components/app-footer/app-footer.tsx
```diff
@@ -0,0 +1,46 @@
+import { ActionButton, DialogTrigger, Divider, Flex, Icon, View } from '@geti-ui/ui';
+import { Manifest } from '@geti-ui/ui/icons';
+
+import { JobStatus } from '../../features/jobs/footer/job-status';
+import { LogsDialog } from '../../features/logs/logs-dialog';
+
+export const AppFooter = ({ gridArea = 'footer', compact = false }: { gridArea?: string; compact?: boolean }) => {
```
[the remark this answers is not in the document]
dwesolow: +1

application/ui/src/components/app-footer/app-footer.tsx
MarkRedeman: I've removed it in the last commit. Before I had a button that was conditionally shown and justified to the end, but will re-add that once needed.

dwesolow (approved):

PR #1031, opened 2026-08-21
feat(backend): fix compile model  and executorch export

AlbertvanHouten:
## Description

**Policy Export:**
* Fixed a bug in `mixin_policy.py` where exporting with `to_executorch()` would crash if the model was on CUDA/XPU by ensuring a deep-copied, CPU-resident version of the model and input sample is used for export, leaving the original untouched.

**Docker Trainer Environment:**
* Updated `Dockerfile.trainer` to explicitly install `build-essential` and `g++` packages as these are necessary to compile a model


library/src/physicalai/export/mixin_policy.py
```diff
@@ -612,10 +613,22 @@ def to_executorch(
...
+        # ExecuTorch doesn't support CUDA/XPU tensors (segfaults instead of
+        # raising), so trace a CPU copy. Move the model to CPU before
+        # deepcopy (not after) to avoid briefly holding two accelerator-
+        # resident copies; original device/mode are restored below.
+        original_device = self.device
+        was_training = self.model.training
+        self.model.to("cpu")
+        try:
+            export_model = copy.deepcopy(self.model).eval()
+        finally:
```
MarkRedeman, line 630: Wouldn't the out of memory result in a hard crash? For instance for some XPU systems with low memory that's shared with both GPU and CPU?

AlbertvanHouten: executorch exports are currently only supported for ACT models which don't have a very high chance to run OOM  I think. I could add some logic here that checks if there is enough remaining memory before trying this operation and otherwise temporarily save the model to disk and then run the export. What do you think?

alfieroddan: Pi05's isn't far off just FYI - https://github.com/open-edge-platform/physical-ai-studio/pull/851.

AFAIK

`to_executorch` has delegation options - [`portable`, `xnnpack`, `openvino`].

portable: CPU
xnnpack: optimized CPU
openvino: Intel CPU/GPU/NPU through a custom OpenVINO-enabled ExecuTorch runtime

CUDA isn't implemented yet - requires a CUDA partioner.

Perhaps we don't need a copy of the model? Just move to eval and cpu, export and then return to train / device if needed?

AlbertvanHouten: I tried without the copy and it seems to work fine so lets go with that approach.  Either way the memory could be a problem for larger models as executorch is making a deep copy regardless: https://github.com/pytorch/executorch/blob/v1.4.1/exir/backend/backend_api.py#L114

MarkRedeman (approved):

PR #1039, opened 2026-08-23
feat: libero full benchmark with reproducible pi05

alfieroddan:
## Description

- overhaul README to show results with reproducible commands
- change default pi05 from v44
- LIBERO full benchmark config
- shard is more general


Daankrol (approved):
Nice!

PR #1040, opened 2026-08-23
docs: new policy design

alfieroddan:
## Description

New policy design.

Main changes to current:

**1. Explicit Features**  
Input and output features now live directly in the model config. `dataset_stats` only provides optional normalization data.

**2. One Initialization Path**  
Every construction route calls `_initialize_from_config()`. It creates the model and processors, loads weights, and applies modifications.

**3.Config Mapping**  
Model should be able to init from_config.

**4. Eager and Lazy Setup**  
Policies initialize eagerly from supplied features or lazily from dataset features. Feature mismatches fail instead of silently rebuilding the model.

**5. Pretrained and Checkpoint Loading**  
Pretrained weights load after constructing the final architecture. Checkpoints restore their saved config without downloading external weights.

**6. Clear Ownership**  
The config is owned by the policy but can fully describe the model. The policy owns training and weight management. The dataset owns feature order and normalization statistics.

**7. Consistent Export**  
Processors and export schemas use the same ordered feature contract as the model. Export no longer needs to reconstruct features from `dataset_stats`.


Daankrol:
I like this new design, really nice work. The original docs we had on policies has drifted alot from the code. (might be a good idea in general to occasionally recheck our docs).
My notes and thoughts:

## model construction and dataset stats
decouple datasets stats from the models. Input and output features should live in the model/policy in a serializable config, normalization stats should be kept in the dataset. I like the idea of having explicit features and not deriving this from data. No routing of feature contract through `datasets_stats`. 
(we have 226 (!) references across 8 policies, export mixin and tests.)
Policy should have clear ownership on this. 

A single point of policy construction (through config) is a good change. No extra paths that are harder to maintain. A single route for the Studio/backend to init a policy. A single source of truth for base model hparams is an easy way to expose this in the Studio UI. We currently have 4 (as far as i know) ways to construct policies and no guards of reloading a model that already has loaded weights.  

## Mixins
Not yet sure how we should handle certain Mixins. This design does not really cover that. The ownership table has 6 owners. Snapflow, PEFT and RTC don't really fit any of them. Each one is a opt-in feature that targets config, policy and model. So mixins need ownership too. 

"Resolve a complete config, materialize once" doesn't survive real policy lifecycles. For example, Snapflow turns on at step 30k, RTC gets toggled at export time, and LoRA changes the model itself after the pretrained weights (of base model) have loaded. Also some mixins have model specific parameters, they are not just plug and play for all models. e.g. LoRA target layers differ per architecture. 

LoRA changes state_dict keys, so it must be replayed at reconstruct time. SnapFlow only changes the architecture after a certain amount of training, so it doesn't. The doc treats all config fields identically and compares them with != on restore. That check wrongly combines "can this checkpoint load" with "are these the same settings", which is why a phase-2 SnapFlow checkpoint or a dropout change would read as an architecture mismatch.

I think we have two options that still match most of your design decisions:

1. Two configs. Model config = features and architecture, checkpointed, does reconstruction. Policy config = recipe plus the capability mixins. Restore compares only the model config. Cost: more moving parts, and the model gets its capability flags set by the policy rather than its own constructor.

2. One model config, mixins mix into it. Least code. Cost: optimizer LR and SnapFlow alpha get called "model config", and every flag change looks like an architecture change. Not 100% clear how that would look like


Great overall restructuring. I like the ownership table, not a bad idea to create some rules based on this in our create-policy skill. This skill offcourse needs a full rewrite. Let me know what you think about the Mixin configs.

library/docs/explanation/policy/template.py
```diff
@@ -0,0 +1,862 @@
...
+    def _apply_model_modifications(self) -> None:
+        model = self._require_model()
+
+        if self.gradient_checkpointing:
+            model.gradient_checkpointing_enable()
+
+        if self.use_lora:
+            model.enable_lora()
+
+    def _from_hf(
```
daniil-lyakhov, line 143: As _from_hf doesn't initialize the model but builds a config, I find this name a bit confusing here

daniil-lyakhov:
Hi @alfieroddan, thank you for the proposal! 

Your points are clear and make sense, but as such a big modification is ongoing, it is complicated for me to cover the whole picture together.
Could you please make following comparisons 
1) How a policy being built now - all the possible scenarios (setup, eager and etc) and how the policy is build in your proposal - to make it clear, what init functions was before and what how functions will work after
2) And the same thing for the Features - where do they live now and where will they live after - in cases of init from checkpoint or config or others

Maybe we can even visualize it with blocks - so that the improvements will be clear and visible on a picture
Thanks!

eugene123tw (commented):
Thanks Alfie for organizing this design. It is a strong starting point, and I think with a bit more separation we can make it much easier for future policy authors to follow.

A few questions and suggestions:

- `sample_input`, `inputs_schema`, `outputs_schema`, and `extra_export_args` are currently properties. In practice these are computed values with validation and failure modes, so debugging can become ambiguous when they fail indirectly. I suggest migrating them to explicit methods so call sites are clearer and stack traces are easier to follow.

- The current naming does not clearly communicate intent. For example, `inputs_schema` sounds generic, but it is mainly used for export tracing and manifest wiring, not training-time logic. Could we rename or document these APIs to make lifecycle intent explicit.

- For `template.py`, can we keep only the essentials in the main flow:
  1. model forward, compute_loss, predict_action_chunk
  2. policy forward, predict_action_chunk, training_step, configure_optimizers
  3. config dataclass
  The current template includes many helpers, but it is not obvious which are required versus optional.

- We should explicitly document required and optional interfaces for new policies:
  1. which methods must be implemented for a minimal working policy
  2. which are optional
  3. when to implement `observation_delta_indices` and `action_delta_indices`, and why

- Optional features such as LoRA and gradient checkpointing would be clearer in dedicated docs rather than embedded in the core template, which is currently a lot to absorb for first-time contributors.

- I propose splitting policy docs into focused pages:
  1. Policy Architecture
  2. Required Interfaces
  3. How-to: implement a new policy
  4. Advanced Patterns: LoRA or PEFT mixin, checkpoint restoration, export customizations
  5. Export API reference explaining what each hook does and when to override it, including inputs_schema, outputs_schema, extra_export_args, to_openvino, to_onnx, sample_input, and _get_default_export_input_sample

library/docs/explanation/policy/README.md
```diff
@@ -1,219 +1,585 @@
...
+A native policy normally separates configuration, model computation, processing, and
+Lightning integration:
 
 ```text
 policy_name/
-├── config.py   # Dataclass configuration
-├── model.py    # PyTorch nn.Module
-└── policy.py   # Lightning module wrapper
+|-- config.py        # Serializable model configuration
+|-- model.py         # PyTorch model
```
samet-akcay, line 24: we might optionally need additional files here to have as lean `policy.py` as possible. for example `optimizer.py`, `checkpoint.py` etc. `Policy` implementation, in my mind, is a lean orchestrator that uses lightning hooks and combine bits and pieces together

library/docs/explanation/policy/README.md
```diff
@@ -1,219 +1,585 @@
...
 
-## Base Class
+`compute_loss()` returns a loss tensor with gradients and a metrics dictionary with
+at least a `"loss"` key. `compute_val_loss()` may reuse the training loss.
+
+The model also implements `observation_delta_indices`, `action_delta_indices`, and
+`reward_delta_indices`. Data loading uses these properties to select the temporal
+context required by the model.
+
+### Policy
```
samet-akcay, line 92: as I mentioned above, perhaps we could highlight that the policy implementation should be as lean as possible? So like it would be really easy to skim over the policy implementation and understand what's happening?

samet-akcay: and I believe we do not utilize lightning hooks as much as we should be. For example we have `_initialize_model` etc, which can be replaced with `configure_model`? --need to double check the behavior though

library/docs/explanation/policy/README.md
```diff
@@ -1,219 +1,585 @@
...
+@dataclass
+class MyModelConfig(Config):
+    input_features: list[Feature]
+    output_features: list[Feature]
+    hidden_size: int = 1024
+    chunk_size: int = 32
+    n_action_steps: int = 32
+```
+
+Architecture fields stay flat so `FromConfig` can map them directly to matching model
```
samet-akcay, line 136: minor comment, after this pr openvinotoolkit/physicalai#227, `FromConfig` is deprecated, we could use `FromConfigMixin` from jsonargparse directly

https://github.com/openvinotoolkit/physicalai/blob/5ada732cbc02e15345fb22b0e908165ae62932e7/src/physicalai/config/mixin.py#L123

library/docs/explanation/policy/README.md
```diff
@@ -1,219 +1,585 @@
...
+constructor arguments:
+
+```python
+class MyModel(Model, FromConfig):
```
samet-akcay, line 140:
```suggestion
class MyModel(FromConfigMixin, Model):
```

library/docs/explanation/policy/README.md
```diff
@@ -1,219 +1,585 @@
...
+    cls,
+    config: MyModelConfig,
+    *,
+    optimizer_lr: float = 1e-4,
+) -> "MyPolicy":
+    policy = cls(
+        n_action_steps=config.n_action_steps,
+        optimizer_lr=optimizer_lr,
+    )
+    policy._initialize_from_config(config)
```
samet-akcay, line 246: is this helper method needed? Can its content not be part of `from_config` directly? So this would be one less helper method in the policy implementation?

library/docs/explanation/policy/README.md
```diff
@@ -1,219 +1,585 @@
...
+
+The postprocessor reverses output normalization in the same order. For multiple
+action features, concatenation and slicing must follow `config.output_features`
+exactly.
+
+A typical policy delegates runtime behavior as follows:
+
+```python
+def forward(self, batch: Observation):
+    model = self._require_model()
```
samet-akcay, line 359: what does _require_model do?

library/docs/explanation/policy/README.md
```diff
@@ -1,219 +1,585 @@
...
+    actions = model.predict_action_chunk(
+        self._prepare_batch(batch, require_actions=False)
+    )
+    return self._postprocessor(actions)
 ```
 
-## Mixins
+## Export
+
+Export-capable policies continue to inherit `ExportablePolicyMixin`. The mixin owns
```
samet-akcay, line 378: again, for the sake of a lean policy implementation..

we could perhaps mention that we could have policy-specific export mixin if the implementation is complex? Like `MolmoACT2ExportMixin(ExportablePolicyMixin)` or `Rldx1ExportMixin(ExportablePolicyMixin)`?

library/docs/explanation/policy/template.py
```diff
@@ -0,0 +1,862 @@
...
+    
+        policy._initialize_from_config(config)
+        return policy
+
+    def _require_model(self) -> NewPolicyModel:
+        if not isinstance(self._model, NewPolicyModel):
+            raise RuntimeError("Policy model is not initialized")
+        return self._model
+
+    def _initialize_from_config(
```
samet-akcay, line 113: can you suggest some alternative names that would be inline with lightning model hooks? so our implementation looks consistent?

library/docs/explanation/policy/template.py
samet-akcay, line 1: this template is here until everyone agrees on the contract, and you merge molmo ?

alfieroddan:
https://github.com/open-edge-platform/physical-ai-studio/pull/949#discussion_r3916202929

We should only trim action chunks / dim outside of model - the model should return full chunk otherwise RTC and other mixins may not work.

library/docs/explanation/policy/template/base.py
```diff
@@ -0,0 +1,108 @@
...
+            return
+
+        resolved_config = NewPolicyModelConfig.from_dict(config_data)
+        if self._config is not None:
+            if self._config != resolved_config:
+                raise ValueError("Checkpoint feature contract does not match the initialized policy")
+            return
+
+        self._config = resolved_config
+        self.configure_model()
```
alfieroddan, line 108: Just a thought.

In the new design we would remove model specific modifications like lora, gradient checkpointing out of the config. I agree with this.

If we use lightning to resume training and/or load training state we must make sure to keep these hparams.

PR #1051, opened 2026-08-27
fix(snapflow): compile-friendly FM/CD loss split + --weights_from warm start

Daankrol:
## Summary
- Makes the SnapFlow FM/CD batch split compile-friendly: replaces the data-dependent Bernoulli-mask-and-`nonzero()` split with a static-size random permutation split (`n_fm = round(alpha * bsize)`, `torch.randperm`), avoiding `torch.compile` graph breaks/recompiles on every step.
- Adds a `--weights_from` CLI option to `physicalai fit`/etc. for warm-starting a model's weights from a checkpoint without resuming full training state (optimizer/LR schedule/global step), which is needed for the SnapFlow phase-1 -> phase-2 handoff when `train_expert_only` changes the optimizer's trainable parameter set.
- Updates SnapFlow distillation configs/docs to use `--weights_from` instead of `--fit.ckpt_path` for the phase-2 handoff, and documents the compile-friendly split rationale.

## Testing
- `uv run pytest tests/unit/policies/test_snapflow.py tests/unit/train/test_callbacks.py tests/unit/cli/test_cli.py -q` → 102 passed

## Notes
- This branch merges the SnapFlow mixin refactor already landed on `main` (`physicalai.policies.mixins`) — the compile-friendly split logic was ported into that new location.

### DIFF
--- a/library/configs/physicalai/pi05_snapflow_distillation.yaml
+++ b/library/configs/physicalai/pi05_snapflow_distillation.yaml
@@ -94,7 +94,7 @@ data:
 trainer:
   # Weights-only warm start via --weights_from starts step/epoch counting at 0,
   # so this is the phase-2 budget alone (not phase-1 + phase-2).
-  max_steps: 30000
+  max_epochs: 2
   accelerator: gpu
   devices: 1
   precision: 32
--- a/library/configs/physicalai/pi05_snapflow_distillation.yaml
+++ b/library/configs/physicalai/pi05_snapflow_distillation.yaml
@@ -9,18 +9,24 @@
 #   # Phase 1 — standard flow-matching training
 #   physicalai fit --config configs/physicalai/pi05.yaml
 #
-#   # Phase 2 — SnapFlow distillation, VLM frozen
+#   # Phase 2 — SnapFlow distillation, VLM frozen, weights-only warm start
 #   physicalai fit \
 #       --config configs/physicalai/pi05_snapflow_distillation.yaml \
-#       --fit.ckpt_path ./experiments/lightning_logs/version_0/checkpoints/last.ckpt \
-#       --trainer.max_steps 30000
+#       --weights_from ./experiments/lightning_logs/version_0/checkpoints/last.ckpt
 #
-# Note that `--fit.ckpt_path` is a full Lightning *resume*: it restores the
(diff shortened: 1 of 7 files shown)


library/configs/physicalai/pi05_snapflow_distillation.yaml
```diff
@@ -78,6 +92,8 @@ data:
     data_format: "physicalai"
 
 trainer:
+  # Weights-only warm start via --weights_from starts step/epoch counting at 0,
+  # so this is the phase-2 budget alone (not phase-1 + phase-2).
   max_steps: 30000
```
AlbertvanHouten: This should still be updated to epochs to match the original pi05 config we have. Same for smolvla

library/docs/how-to/training/snapflow_distillation.md
```diff
@@ -335,7 +342,7 @@ policy.enable_snapflow(alpha=0.5, lambda_=0.1, num_inference_steps=1)
 | `best_teacher_monitor`   | `None`                   | Disambiguates which monitored `ModelCheckpoint` to restore from when more than one is configured.                               |
 | `scope_best_to_phase`    | `true`                   | Reset best-checkpoint tracking at the boundary, since `val/loss` is not comparable across `num_inference_steps`.                |
 | Phase-1 budget           | 5-10 epochs              | Fewer for a warm-started VLA than for from-scratch training.                                                                    |
-| Phase-2 budget           | ~3-5 epochs (~30k steps) | Short because the target-time embedding is zero-initialised.                                                                    |
+| Phase-2 budget           | ~2-3 epochs (~30k steps) | Short because the target-time embedding is zero-initialised.                                                                    |
```
AlbertvanHouten:
```suggestion
| Phase-2 budget           | ~2-3 epochs | Short because the target-time embedding is zero-initialised.                                                                    |
```

library/configs/physicalai/pi05_snapflow_distillation.yaml
Daankrol: @copilot Change to max 2 epochs for this Snapflow phase. Don't use steps here.

AlbertvanHouten (dismissed):

alfieroddan (dismissed):

MarkRedeman (approved):

PR #1052, opened 2026-08-28
feat: add restart server button to footer

MarkRedeman:
This PR adds a new UI component that gets shown when the UI detects that the server requires a restart.
When the user clicks the 'Restart server" button we show a dialog which, when confirmed, remains open until the server has finished restarting.
To check that the server successfully restarted we compare the `instance_id` added to the health endpoint in #1030

image

### DIFF
--- a/application/ui/src/components/app-footer/app-footer.tsx
+++ b/application/ui/src/components/app-footer/app-footer.tsx
@@ -3,6 +3,7 @@ import { Manifest } from '@geti-ui/ui/icons';
 
 import { JobStatus } from '../../features/jobs/footer/job-status';
 import { LogsDialog } from '../../features/logs/logs-dialog';
+import { RestartRequiredBanner } from '../../features/system/restart-required-banner';
 
 export const AppFooter = ({ gridArea = 'footer' }: { gridArea?: string }) => {
     return (
@@ -16,25 +17,28 @@ export const AppFooter = ({ gridArea = 'footer' }: { gridArea?: string }) => {
             paddingX='size-100'
             paddingY='size-0'
         >
-            <Flex alignItems={'center'} height='100%' gap='size-100' minWidth={0}>
-                <View overflow={'hidden'}>
-                    <DialogTrigger type='fullscreen'>
-                        <ActionButton
-                            isQuiet
-                            UNSAFE_style={{
-                                paddingRight: 'var(--spectrum-global-dimension-size-100)',
-                            }}
-                        >
-                            <Icon>
-                                <Manifest />
-                            </Icon>
-                            Logs
-                        </ActionButton>
-                        {(close) => <LogsDialog close={close} />}
-                    </DialogTrigger>
-                </View>
-                <Divider orientation='vertical' size='S' />
-                <JobStatus />
+            <Flex alignItems='center' justifyContent='space-between' gap='size-100' height='100%'>
+                <Flex alignItems={'center'} height='100%' gap='size-100' flex={1} minWidth={0}>
+                    <View overflow={'hidden'}>
+                        <DialogTrigger type='fullscreen'>
+                            <ActionButton
+                                isQuiet
+                                UNSAFE_style={{
+                                    paddingRight: 'var(--spectrum-global-dimension-size-100)',
+                                }}
+                            >
+                                <Icon>
+                                    <Manifest />
+                                </Icon>
+                                Logs
+                            </ActionButton>
+                            {(close) => <LogsDialog close={close} />}
+                        </DialogTrigger>
+                    </View>
+                    <Divider orientation='vertical' size='S' />
+                    <JobStatus />
+                </Flex>
+                <RestartRequiredBanner />
             </Flex>
         </View>
     );
--- a/application/ui/src/features/system/restart-required-banner.tsx
+++ b/application/ui/src/features/system/restart-required-banner.tsx
@@ -1,0 +1,54 @@
+import { Button, Flex, Text, View } from '@geti-ui/ui';
+
+import { useRestartServerMutation, useRestartState } from './restart-state';
+
+const statusText: Record<string, string> = {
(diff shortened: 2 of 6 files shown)


application/ui/src/components/app-footer/app-footer.tsx
```diff
@@ -16,25 +17,28 @@ export const AppFooter = ({ gridArea = 'footer' }: { gridArea?: string }) => {
...
-                <Divider orientation='vertical' size='S' />
-                <JobStatus />
+            <Flex alignItems='center' justifyContent='space-between' gap='size-100' height='100%'>
+                <Flex alignItems={'center'} height='100%' gap='size-100' flex={1} minWidth={0}>
+                    <View overflow={'hidden'}>
+                        <DialogTrigger type='fullscreen'>
+                            <ActionButton
+                                isQuiet
+                                UNSAFE_style={{
+                                    paddingRight: 'var(--spectrum-global-dimension-size-100)',
```
camiloHimura, line 27:
```suggestion
                                    paddingRight: dimensionValue('size-100'),
```

application/ui/src/features/system/restart-required-banner.tsx
```diff
@@ -0,0 +1,41 @@
...
+                    {isRestarting ? 'Restarting the server…' : 'Restart the server to activate the plugin changes.'}
+                </Text>
+                <Button
+                    variant='primary'
+                    style='fill'
+                    isDisabled={isRestarting}
+                    onPress={openRestartPrompt}
+                    UNSAFE_style={{
+                        minHeight: '24px',
+                        paddingInline: 'var(--spectrum-global-dimension-size-200)',
```
camiloHimura, line 32:
```suggestion
                        minHeight: dimensionValue('size-300'),
                        paddingInline: dimensionValue('size-200'),
```

application/ui/src/features/system/restart-state.tsx
```diff
@@ -0,0 +1,213 @@
...
+        <Dialog>
+            <Heading>Restart server now?</Heading>
+            <Divider />
+            <Content>
+                <Flex direction='column' gap='size-150'>
+                    <Text>Plugin changes require a server restart to become active.</Text>
+                    {activeTrainingJobCount > 0 ? (
+                        <Text>
+                            Restarting now will interrupt {activeTrainingJobCount} active training job
+                            {activeTrainingJobCount === 1 ? '' : 's'}.
```
camiloHimura, line 65: Geti has a `pluralize` helper in `application/ui/src/shared/util.ts`. It could be useful in this repo as well.

application/ui/src/features/system/restart-state.tsx
camiloHimura, line 1: I think this file should have the `-provider` suffix.

application/ui/src/features/system/restart-state.tsx
```diff
@@ -0,0 +1,213 @@
...
+        meta: { skipInvalidation: true },
+    });
+    const { data: jobs = [] } = $api.useQuery('get', '/api/jobs');
+    const healthQuery = $api.useQuery('get', '/api/health', {
+        retry: false,
+        staleTime: 0,
+        refetchOnWindowFocus: false,
+        refetchOnReconnect: 'always',
+    });
+    const health = healthQuery.data as HealthResponse | undefined;
```
dwesolow, line 107: Is `health` not type safe based on the API types?

application/ui/src/features/system/restart-state.tsx
```diff
@@ -0,0 +1,213 @@
...
+    }, [healthQuery, isRestarting, restartMutation]);
+
+    useEffect(() => {
+        const instanceId = health?.instance_id;
+        if (instanceId === undefined) {
+            return;
+        }
+
+        if (lastInstanceId.current !== undefined && instanceId !== lastInstanceId.current) {
+            queryClient.clear();
```
dwesolow, line 157: why is this needed?

application/ui/src/features/system/restart-state.tsx
```diff
@@ -0,0 +1,213 @@
...
+            setIsRestarting(false);
+            setRestartRequested(false);
+            setRestartPromptOpen(false);
+        }
+        lastInstanceId.current = instanceId;
+    }, [health?.instance_id, queryClient]);
+
+    useEffect(() => {
```
dwesolow, line 165: could we move this logic to `refetch` in `healthQuery`?

application/ui/src/features/system/restart-state.tsx
```diff
@@ -0,0 +1,213 @@
...
+                    <Button variant='secondary' onPress={onDismiss}>
+                        Later
+                    </Button>
+                ) : null}
+            </ButtonGroup>
+        </Dialog>
+    </DialogContainer>
+);
+
+export const RestartStateProvider = ({ children }: { children: ReactNode }) => {
```
dwesolow, line 90: Why does this need to be a provider? I see that `useRestartState` is used only in one place atm.

application/ui/src/features/system/restart-state.tsx
MarkRedeman: It's mainly a safety procedure: when the server restarts this might make some of our queries stale, for instance due to a project having a robot installed whose plugin was previously uninstalled.
In such case the API returns `unavailable: true` and may lose out of other info. After installing a plugin and restarting the server his may chance, hence the need for a full "reboot" of the UI as well.

Will update the code with a comment explaining this.

application/ui/src/features/system/restart-state.tsx
MarkRedeman: Will be used in a follow up PR, by the code that installs plugins.

dwesolow (approved):

MarkRedeman:
@dwesolow and @camiloHimura I'm going to make a follow up PR with the fixes from your suggestions.
This way I can prevent some merge conflicts with another PR.

PR #1055, opened 2026-08-28
feat(rldx1): add graph-safe OpenVINO export path and runtime-aligned RLDX-1 preprocessing

eugene123tw:
## Description
This PR adds first-class export support for RLDX-1 in Physical AI Studio on branch feat/rldx1-ov, including a graph-safe export path for OpenVINO/ONNX-style tracing constraints, runtime-aligned preprocessing metadata, and schema/stat handling needed for upstream RLWRLD checkpoints.

Compared against main, the branch primarily introduces:
- Graph-safe RLDX-1 backbone/text/vision export wrappers so export no longer depends on non-traceable eager paths.
- RLDX-1 export helper and stats helper utilities for stable input schema construction and manifest generation.
- Explicit feature merge support so missing visual metadata in upstream checkpoints can be supplied by users.
- Export mixin and backend improvements for cleaner export argument handling and input naming robustness.
- Unit test coverage for graph-safe text/vision behavior, export sample merge logic, and export argument handling.

## Why
RLDX-1 uses data-dependent operations in eager forward (for example dynamic token/span handling and rope-related construction) that are not directly traceable for deployment export flows. This PR introduces a graph-safe export-time model view and tightens policy export plumbing so Studio can produce deployment-ready OpenVINO artifacts for Runtime consumption.

## Main changes
- Added graph-safe export modules for RLDX-1 backbone/text/vision and integrated them into policy export flow.
- Added export-only helpers for:
  - Token composer parameter generation
  - Reference token span construction
  - fp32 casting for safe tracing
  - Export sample trimming/padding
- Added stats/schema helpers to:
  - Merge user-provided input_features/output_features into dataset stats
  - Infer number of visual views
  - Resolve feature shape robustly
- Updated RLDX-1 policy export metadata to emit Runtime preprocessor specs (normalize, rldx1, tokenizer, token composer, rope) with synchronized num_views/num_frames/max_state_dim.
- Added/updated unit tests around graph-safe behavior and export regression points.

## Coupled Runtime PR
This change is tightly coupled with Runtime work in:
https://github.com/openvinotoolkit/physicalai/pull/257

Please review and merge in coordination, since Studio export metadata and Runtime loading/preprocessing behavior must stay aligned.

## Usage example
    policy = Rldx1(
        pretrained_name_or_path="RLWRLD/RLDX-1-FT-LIBERO",
        input_features={
            "front_view": Feature(ftype=FeatureType.VISUAL, shape=(3, 256, 256), name="front_view"),
            "left_wrist_view": Feature(ftype=FeatureType.VISUAL, shape=(3, 256, 256), name="left_wrist_view"),
        },
    )

    policy.export("rldx1_libero_finetuned_hf_ov", backend="openvino")
    policy.to_openvino("rldx1_libero_finetuned_hf_ov")

    model = InferenceModel(
        export_dir="./rldx1_libero_finetuned_hf_ov",
        device="CPU",  # or "GPU" for Intel iGPU/dGPU
        callbacks=[Rldx1VtcWindowCallback(video_length=4, video_stride=2)]
    )

    benchmark = LiberoBenchmark(
        task_suite="libero_10",   # --env.task=libero_10
        task_ids=[6],             # --env.task_ids=[6] — reproduces libero_10_6
        num_episodes=5,          # --env.num_episodes=10
        max_steps=300,            # match the 500 steps seen in the failing run
        video_dir="./tmp_scripts/ov_libero_eval_videos",
        record_mode="all",
    )
    results = benchmark.evaluate(model)
    print(results.summary())

## Important notes and current limitations
- For upstream RLWRLD trained checkpoints, users must provide input_features with visual shapes when constructing Rldx1. Upstream checkpoint stats do not carry complete visual feature metadata needed for export.
- When running via InferenceModel, users must set:
  callbacks=[Rldx1VtcWindowCallback(video_length=4, video_stride=2)]
  because RLDX-1 inference depends on this temporal window behavior.
- Today, callbacks cannot be declared in manifest.json, so this requirement remains an explicit runtime caller responsibility.

## Validation
- Added/updated unit tests for:
  - Graph-safe text export logic
  - Graph-safe vision parity/restore behavior
  - Dynamic export sample merge behavior
  - Export mixin argument handling and naming behavior


library/src/physicalai/policies/rldx1/components/backbone/graph_safe_rldx1.py
```diff
@@ -0,0 +1,162 @@
...
+BACKBONE_FEATURES = rldx1_constants.BACKBONE_FEATURES
+EMBODIMENT_ID = rldx1_constants.EMBODIMENT_ID
+IMAGE_GRID_THW = rldx1_constants.IMAGE_GRID_THW
+INPUT_IDS = rldx1_constants.INPUT_IDS
+PIXEL_VALUES = rldx1_constants.PIXEL_VALUES
+POSITION_IDS = rldx1_constants.POSITION_IDS
+STATE = rldx1_constants.STATE
+
+
+class GraphSafeRldx1Model(nn.Module):
```
samet-akcay, line 67: what are the pros/cons of wrapping the modules vs directly modifying them to make them export friendly

library/src/physicalai/policies/rldx1/policy.py
```diff
@@ -745,3 +829,454 @@ def _get_warmup_hold_action(self, batch: Observation, env_action_dim: int) -> to
...
+    @staticmethod
+    def get_supported_export_backends() -> list[str | ExportBackend]:
+        """Get a list of export backends supported by policy.
+
+        This method returns a list of supported export backends as strings.
+
+        Returns:
+            list[str | ExportBackend]: A list of supported export backends.
+        """
+        return [ExportBackend.TORCH, ExportBackend.OPENVINO]
```
samet-akcay: no onnx, intentional?

library/src/physicalai/policies/rldx1/policy.py
```diff
@@ -745,3 +829,454 @@ def _get_warmup_hold_action(self, batch: Observation, env_action_dim: int) -> to
 
         # Fallback for tasks without compatible state-to-action mapping.
         return torch.zeros(batch_size, env_action_dim, dtype=torch.bfloat16, device=self.device)
+
+    @staticmethod
+    def get_supported_export_backends() -> list[str | ExportBackend]:
```
samet-akcay: the export logic in this policy.py file becomes a bit large. 

the purpose of `ExportPolicyMixin` was to encapsulate this logic, but this seems not possible with this policy. 

would it be possible to implement rldx specific export mixin?

samet-akcay: I did a bit of quick LLM fact check, here is the result?

```python
   class Rldx1(Rldx1ExportMixin, Policy):
       ...
 ```

 with:

 ```python
   class
 Rldx1ExportMixin(ExportablePolicyMixin):
       ...
 ```

 The resulting method resolution order is:

 ```text
   Rldx1
   ├── Rldx1ExportMixin
   ├── ExportablePolicyMixin
   └── Policy
 ```

 So Rldx1ExportMixin inherits all generic export
 capabilities and overrides only the
 RLDX-specific parts.

 Example

 ```python
   # rldx1/export.py

   from physicalai.export import
 ExportablePolicyMixin, ExportBackend


   class
 Rldx1ExportMixin(ExportablePolicyMixin):
       """RLDX-1-specific export behavior."""

       @staticmethod
       def get_supported_export_backends() ->
 list[str | ExportBackend]:
           return [
               ExportBackend.TORCH,
               ExportBackend.ONNX,
               ExportBackend.OPENVINO,
           ]

       @property
       def inputs_schema(self):
           # RLDX-specific schema
           ...

       @property
       def outputs_schema(self):
           # RLDX-specific schema
           ...

       @property
       def extra_export_args(self):
           # RLDX-specific Runtime components
 and exporter settings
           ...

       def
 _get_default_export_input_sample(self):
           # RLDX-specific VTC preprocessing and
 static prompt construction
           ...

       def _build_graph_safe_model(self, ...):
           ...

       @contextmanager
       def _graph_safe_export_model(self, ...):
           ...

       @torch.no_grad()
       def to_openvino(self, output_path,
 input_sample=None, **export_kwargs):
           if input_sample is None:
               input_sample =
 self._get_default_export_input_sample()

           with
 self._graph_safe_export_model(...):
               return super().to_openvino(
                   output_path,
                   input_sample=input_sample,
                   **export_kwargs,
               )

       @torch.no_grad()
       def to_onnx(self, output_path,
 input_sample=None, **export_kwargs):
           if input_sample is None:
               input_sample =
 self._get_default_export_input_sample()

           with
 self._graph_safe_export_model(...):
               return super().to_onnx(
                   output_path,
                   input_sample=input_sample,
                   **export_kwargs,
               )
 ```

 Then policy.py becomes:

 ```python
   from physicalai.policies.rldx1.export import
 Rldx1ExportMixin


   class Rldx1(Rldx1ExportMixin, Policy):
       ...
 ```

File: library/src/physicalai/policies/rldx1/components/backbone/graph_safe_rldx1.py (hunk 1 above)
Comment (eugene123tw) [reply library/src/physicalai/policies/rldx1/components/backbone/graph_safe_rldx1.py:67]: I suggest keeping the GraphSafe wrapper approach for now (ported from RLDX1 upstream), rather than directly modifying the base RLDX1 modules.

Pros:
1. It isolates export-only traceability work from training and eager inference behavior, which lowers regression risk.
2. It solves tracing blockers cleanly (position embedding interpolation, M-RoPE index path, VTC compression span logic) without changing core model semantics.
3. It improves export determinism and speed by precomputing static tensors once (e.g `pos_embed`)

Cons:
1. We now have dual paths (base + GraphSafe), so drift is a maintenance cost.
2. Any functional change in the base backbone may need a mirrored GraphSafe update.
3. Static assumptions (prompt/image layout used during precompute) must stay aligned with runtime/export contract.

Why not directly patch base modules right now:
1. Higher risk of changing training/eager behavior while trying to make code export-friendly.
2. Larger validation surface and slower iteration.
3. Some backend/operator limitations are not solved by refactoring base code anyway.

Going forward:
1. Use wrappers for export-specific graph constraints or static-shape baking.
2. Push changes into base modules only when they are behavior-preserving, useful for all paths, and pass full training/inference parity.
3. If a fix is universally beneficial and stable, we can later upstream/merge that part back into the base path and shrink wrapper scope.

Parity checklist (to manage drift):
1. Eager vs GraphSafe output parity on Libero (with tolerances).
2. Exported model parity vs eager policy on action outputs.

File: library/src/physicalai/policies/rldx1/policy.py [hunk 4]
```diff
@@ -270,7 +299,7 @@ def __init__(  # noqa: PLR0913
         self.save_hyperparameters(ignore=["config"])
 
         self.model: Rldx1Model | None = None  # type: ignore[assignment]
-        self._preprocessor: Rldx1Preprocessor | None = None
+        self._preprocessor: torch.nn.Module = cast("torch.nn.Module", None)
```
samet-akcay, line 310: is this change a must ?

samet-akcay (commented):
I see you create the following file for generic utils: `library/src/physicalai/policies/utils/features.py`

There is also additional utils files for rldx1 specifically such as 
- `library/src/physicalai/policies/rldx1/stats_helpers.py`
- `library/src/physicalai/policies/rldx1/pretrained_utils.py`
- `library/src/physicalai/policies/rldx1/export_helpers.py`

Would it be an idea to follow the same convention as `library/src/physicalai/policies/utils/`, where you have a util directory under rldx like `library/src/physicalai/policies/rldx1/utils`. So you would have something like:

- `library/src/physicalai/policies/rldx1/utils/stats.py`
- `library/src/physicalai/policies/rldx1/utils/pretrain.py`
- `library/src/physicalai/policies/rldx1/utils/export.py`

This would also reduce the number of files we see in library/src/physicalai/policies/rldx1?

library/src/physicalai/policies/rldx1/policy.py
eugene123tw: Pyrefly complains a lot, so had to cast it.

library/src/physicalai/policies/rldx1/policy.py
eugene123tw: added

library/src/physicalai/policies/rldx1/policy.py
eugene123tw: Moved to export.py and add a `Rldx1ExportMixin`

samet-akcay (approved):
lgtm thanks

PR #1057, opened 2026-08-31
feat: integrate robot plugin schema into UI form

MarkRedeman:
This PR completes the integration of robot plugins into Studio's UI.
Two big changes include:
1. the form's state has been changed from,
```tsx
type SetRobotFormContextType = {
    setActiveType: (type: ConfigurableRobotType) => void;
    updateFormData: <K extends ConfigurableRobotType>(
        schemaType: K,
        update: Partial<FormDataForSchema[K]> | ((prev: FormDataForSchema[K]) => FormDataForSchema[K])
    ) => void;
};
```

to
```tsx
type RobotFormState = {
    activeType: string | undefined;
    name: string;
    payload: RobotPayload;
};
```

Before `FormDataForSchema` was strictly typed based on the available robots. With the new design this is no longer possible as users can install new plugins at runtime.
So instead the UI tries to keep its payload compatible with the selected robot's form.
We do this by leveraging a robot catalog's schema, available from `GET /api/robots/catalog/{robot_type}/schema`. This is the other big change.
2. The form fields that are shown when configuring a robot are no longer hardcoded (* except for SO101 and bimanual so101), instead the UI parses the `/api/robots/catalog/{robot_type}/schema` api and generates form elements. Each plugin can add a `x-physicalai-ui` field to their schema that informs the UI wether it needs to show a simple `TextField`, or a more complicated `ConnectionField`.

In additoin this PR also comes with a small tweaks that I've kept in as otherwise merge conflicts would be more complicated:
1. When changing a robot type we now change the robot name if the user didn't customize it
2. extend `robot_field` with a `advanced_configuration` option. These advanced options are currently hidden. In `schema-form.tsx` we have a "Show advanced options" to show them, but the UX is not great yet so for now this option is hidden.


application/ui/src/components/form-preview-layout.tsx
```diff
@@ -20,7 +20,7 @@ export const FormPreviewLayout = ({ form, preview, previewProps }: FormPreviewLa
     return (
         <Grid
             areas={['form preview']}
-            columns={[minmax('size-6000', 'auto'), minmax(0, '1fr')]}
+            columns={['size-6000', minmax(0, '1fr')]}
```
MarkRedeman, line 23: The help text of some of the plugins I've been testing made the sidebar very big. I feel like s`size-6000` probably is good enough for our current UX.

application/ui/src/features/robots/robot-form/robot-schema/components/connection-field.tsx
```diff
@@ -0,0 +1,134 @@
...
+                'The robot could not be identified. Make sure it is powered on and not already in use, then try again.'}
+        </InlineAlert>
+    );
+};
+
+export const ConnectionField = ({ robotType, payload, options, onChange }: ConnectionFieldProps) => {
+    const discover = useDiscoverRobotsQuery(robotType);
+    const identify = useCatalogIdentifyMutation();
+    const connectionKey = options.bind.connection;
+    const serialNumberKey = options.bind.serial_number;
```
dwesolow, line 90: do we need to normalize serial number here?

application/ui/src/features/robots/robot-form/robot-schema/schema-form.tsx
```diff
@@ -0,0 +1,212 @@
...
+            {item.title !== undefined && <Heading level={4}>{item.title}</Heading>}
+            {item.description !== undefined && <Text>{item.description}</Text>}
+            <SchemaFormItems {...props} items={item.items} renderUnownedFields={false} />
+        </Flex>
+    );
+};
+
+const SchemaFormItems = ({ items, renderUnownedFields, ...props }: SchemaFormItemsProps) => {
+    const ownedFields = fieldNamesOwnedByItems(items);
+    const unownedFields = Object.entries(props.properties).filter(([name]) => !ownedFields.has(name));
```
dwesolow: if `renderUnownedFields` is false then we don't need to compute `unownedFields`

application/ui/src/features/robots/robot-form/robot-schema/schema-form.tsx
```diff
@@ -0,0 +1,212 @@
...
+        if (Object.keys(payload).length !== 0) return;
+        const defaults = schemaDefaults(properties, definitions);
+        if (Object.keys(defaults).length !== 0) setPayload(defaults);
+    }, [definitions, payload, properties, setPayload]);
+
+    const isFieldVisible: IsFieldVisible = (name, field, fieldRequired) => {
+        const resolvedField = resolveReference(field, definitions);
+        const fieldUi = resolvedField['x-physicalai-ui'];
+
+        const isRequired = fieldRequired.has(name) || (!isUiItems(fieldUi) && fieldUi?.required === true);
```
dwesolow: this check for `isRequired` is present in a few places, imo worth extracting to separate function

application/ui/src/features/robots/robot-form/form.tsx
```diff
@@ -36,50 +46,53 @@ export const RobotType = () => {
...
-        case 'Trossen_Bimanual_WidowXAI_Follower':
-            formFields = <BiManualWidowxAIFormFields />;
-            break;
-        // Bimanual SO101 is a plugin robot type not present in the backend catalog; kept for persisted robots.
-        case 'BimanualSO101_Follower' as ConfigurableRobotType:
-        case 'BimanualSO101_Leader' as ConfigurableRobotType:
-            formFields = <BimanualSO101FormFields />;
-            break;
-    }
+    useEffect(() => {
```
dwesolow, line 53: why `autoFocus` on line 69 is not sufficient?

application/ui/src/features/robots/robot-form/form.tsx
```diff
@@ -36,50 +46,53 @@ export const RobotType = () => {
...
     );
 };
 
+const SelectedRobotFields = ({ activeType }: { activeType: string }) => {
+    const schema = useRobotCatalogSchema(activeType);
+    const isBimanualSO101 = activeType === 'BimanualSO101_Follower' || activeType === 'BimanualSO101_Leader';
+    return isBimanualSO101 ? (
+        <BimanualSO101FormFields />
+    ) : (
+        schema.data && <SchemaForm schema={schema.data as Parameters<typeof SchemaForm>[0]['schema']} />
```
dwesolow: that is suspense query so `data` should be already defined

dwesolow (dismissed):

application/ui/src/features/robots/robot-form/robot-schema/components/connection-field.tsx
```diff
@@ -121,18 +121,24 @@ export const ConnectionField = ({ robotType, payload, options, onChange }: Conne
...
-                        <Refresh />
-                    </Icon>
-                </ActionButton>
-                {options.identify && (
-                    <ActionButton
-                        onPress={() => identify.mutate({ params: { path: { robot_type: robotType } }, body: payload })}
-                        isDisabled={identify.isPending}
-                    >
-                        Identify
+                <View>
```
dwesolow, line 124: I guess that the issue might be with the shrinking, so `flexShrink={0}` might be root cause issue instead of `View`

application/ui/src/features/robots/robot-form/form.tsx
MarkRedeman: I also want the name field to get focus after selecting a different robot type.
So the first time the form is opened you select a robot type and this sets the name, but adds focus so you can easily change it.
If you did change it then we don't rename the robot after selecting a different type and also don't refocus.
If you did not change the name but do change the type then we rename to that type's name and place focus on the field again.

dwesolow (approved):

PR #1058, opened 2026-08-31
feat: OpenVINO INT8 Weight Compression post hook support

daniil-lyakhov:
## Description

OpenVINO INT8 Weight Compression Post Hook support

The Compression post hook allows to apply INT8 symmetric weight compression to any policy, replacing the original model  inplace


library/src/physicalai/export/mixin_policy.py
```diff
@@ -716,6 +716,7 @@ def export(
         output_path: PathLike | str,
         backend: ExportBackend | str,
         input_sample: dict[str, torch.Tensor] | None = None,
+        post_export_hooks: list[Callable[[str], None]] | None = None,
```
sovrasov, line 719: I think this new export feature can be at least a tiny bit documented in `library/docs`

library/src/physicalai/export/hooks.py
```diff
@@ -0,0 +1,66 @@
...
+"""Post-export hooks for model compression and optimization."""
+
+import logging
+import tempfile
+from pathlib import Path
+
+logger = logging.getLogger(__name__)
+
+
+def compress_weights_openvino_int8_sym(model_path: str) -> None:
```
sovrasov, line 13: At least one example on hot to utilize that hook is required

library/tests/unit/export/test_mixin_export.py
```diff
@@ -759,3 +760,118 @@ def test_returns_none_without_sample_input(self):
         wrapper = ExportWrapper(model)
 
         assert wrapper._get_default_export_input_sample() is None
+
+
```
sovrasov, line 764: Why 3 lines?

daniil-lyakhov: Fixed

library/src/physicalai/export/mixin_policy.py
daniil-lyakhov: Updated the library/docs/explanation/export/README.md and library/docs/how-to/export/export_inference.md

library/src/physicalai/export/hooks.py
daniil-lyakhov: https://github.com/open-edge-platform/physical-ai-studio/pull/1058/changes#diff-9da8623459cc165e7d8320f9ab867a7a37530d23ba54bdd41fdfbb81cceae7f0R80-R96

sovrasov (approved):

PR #1061, opened 2026-09-01
feat: install robot plugins via UI

MarkRedeman:
This PR adds a new `/plugins` page (hidden under the `plugins` feature flag, use `setFeatureFlag('plugins', true)` to enable)
This page allows the user to install a hardcoded list of plugins. Later we can also extend this to allow users to search plugins from pypi and/or install manually.

image

**Opens / follow ups**:
- What list of plugins do we want to expose by default? I'm thinking bimanual SO101 and rebot (and when we extract the stararm robot from the rebot plugin we can add stararm). The lekiwi (we need to update our teleoperation logic), lerobot (exposes too many robots, making UX bad) and mujoco (requires manually starting the mujoco simulation) plugins each require more work.
- When deploying with docker the user may lose their plugins when doing `docker compose down`. This will be fixed in a followup PR by adding a restoration cli + ui
- When installing a plugin instead of a simple "loading..." button we can show a dialog and have the server stream in installation logs
- Currently the plugin api doesn't allow to remove a plugin which is in use by 1 or more robots, this constraint can be removed. (thanks to #1023)
- Need to tweak the UI a bit more, we can probably simplify it more


application/ui/src/features/plugins/plugins.test.tsx
```diff
@@ -0,0 +1,136 @@
...
+        );
+        const user = userEvent.setup();
+
+        render(<PluginsView />, { route: '/plugins', path: '/plugins' });
+
+        await user.click(await screen.findByRole('button', { name: 'Install' }));
+
+        expect(await screen.findByText('Plugin changes require a server restart to become active.')).toBeVisible();
+        expect(screen.getByRole('button', { name: 'Restart now' })).toBeVisible();
+        await user.click(screen.getByRole('button', { name: 'Later' }));
```
dwesolow, line 71: should we test what is visible after clicking `Later`?

application/ui/src/features/plugins/plugins.test.tsx
```diff
@@ -0,0 +1,136 @@
...
+    });
+
+    it('restarts after confirming the restart prompt', async () => {
```
dwesolow, line 74: IMO this test could be merged with the "opens a restart prompt after installing a plugin"

application/ui/src/features/plugins/plugins-disabled.tsx
```diff
@@ -0,0 +1,12 @@
+import { Flex, Heading, Text, View } from '@geti-ui/ui';
+
+export const PluginsDisabledView = () => {
```
dwesolow, line 3: I see we already handle a case that when plugins ff is disabled then plugins should not be visible. Do we still need this component?

AlbertvanHouten (approved):

PR #1065, opened 2026-09-01
feat(runtime): add session status report

maxxgx:
Runtime sessions are detached processes that outlive the API that spawned them, so their lifecycle is independent of the backend. Today a leftover session is invisible and can only be cleared by hunting down its pid. This makes them visible and stoppable.

Backend:
- `GET /api/runtime/sessions`, `GET /sessions/count`, `POST /sessions/{name}/stop`. All host-wide rather than project-scoped: the lock dir is host-wide, and a session can outlive its robot row.
- State is rebuilt from the lock directory and each session's metadata queryable, so sessions started before a backend restart are still reachable. Probes run concurrently and never subscribe, so listing a session does not reset its idle timeout.
- Statuses: `starting / running / stopped / error / unreachable`. Unreachable sessions stay in the list, since one holding a robot that won't answer is the most likely to need attention.
- Stop is graceful (SIGTERM through the normal idle-timeout teardown), verifies the lock was released before reporting success, and is idempotent.
- Fixed model reload: refreshing inference page does not reload model again

UI:
- Footer entry showing N sessions, hidden when none. Popover lists activity, uptime, cameras, and the idle-shutdown countdown.
- We can pin to keep the list open while working. The popover is non-modal, so the page stays interactive.
- Stop is an inline confirm on the row, scoped so one in-flight stop doesn't disable the others.
- Robots list gets a per-robot session light; the recording page now renders the footer too.
- Two poll rates: the count every 5s (a directory read, mounted on every page via the footer) and the list every 2s (opens a transport session per runtime session, mounted by the popover and by the robots list).

PR size overview -> 2078 lines changed:
- UI: 700 (34%)
- UI tests: 541 (26%)
- Backend tests: 479 (23%)
- Backend: 358 (17%)

https://github.com/user-attachments/assets/e0354370-ddfc-491b-9831-56ecddbb5b59


application/ui/mockups/runtime-sessions.tsx
MarkRedeman, line 1: Leftover?

maxxgx: Used it to play around to quickly check the ux. I think we can keep it since we already have this mockups dir

maxxgx: Removed it because of broken CI

RHeckerIntel (approved):

PR #1066, opened 2026-09-01
feat: Xiaomi Robotics 0 Model Support

daniil-lyakhov:
## Summary

* Introduces XR0, a flow-matching Vision-Language-Action policy
* Generalizes the export subsystem's hook API to support the graph-baking and input-renaming that XR0's OpenVINO export requires.
* Model is runnable on Intel GPUs in FP16/FP32 precisions, BF16 OpenVINO support WIP
* Model training runs, but the convergence is WIP

## The XR0 model (summary by opus)
XR0 is a flow-matching VLA built on a Qwen3-VL-4B vision-language backbone with a DiT action expert (rectified-flow action head):

Backbone: Qwen/Qwen3-VL-4B-Instruct, with 3D MRoPE positional encoding over interleaved image + text tokens.
Action head: 16-layer DiT decoder (hidden 1024, 8 KV heads, head dim 128) that iteratively refines an action chunk from noise via Euler integration (5 inference steps by default).
I/O: predicts an action chunk (default horizon 30) conditioned on a single current observation (multi-camera images + state + task prompt).

New package  (config / model / policy / dit / preprocessor / qwen3_vlm / export_openvino), a xr0.yaml config, and full unit + OpenVINO parity test coverage.

### 1. Global pre-export hooks + unified hook orchestration

export() now accepts both `pre_export_hooks` and `post_export_hooks`.

`pre_export_hooks` — run before tracing/conversion; mutate the model in place (e.g. bake constants into the graph, swap ops for export-friendly variants).
`post_export_hooks` — run after the artifact is written; receive the export path and may rewrite the file in place (e.g. INT8 compression).

XR0 uses a pre-hook (_bake_ingraph_export) to bake its in-graph position-id computation before tracing.

### 2. Declarative hooks on ExportParameters

ExportParameters are extended by pre_export_hooks and post_export_hooks fields so a policy can declare its export-time graph updates alongside its other export params.

### 3. OpenVINO input renaming (input_name_map)

OpenVINOExportParameters are extended by `input_name_map`. After conversion, _rename_openvino_inputs renames graph input ports so they align with the keys emitted by preprocessor components.

## Testing
* New XR0 unit suites (config, dit, model, policy, preprocessor, qwen3_vlm, pretrained_utils, export_openvino) + test_xr0_openvino_parity.py.

* Export hook tests updated to route through export() and assert hook ordering (pre → convert → post) and a path passed to post-hooks.


library/src/physicalai/export/backends.py
```diff
@@ -60,12 +81,20 @@ class ONNXExportParameters(ExportParameters):
...
+            to rename the converted graph's input tensors before saving. Useful to
+            align the graph ports with the keys emitted by preprocessor components
+            (e.g. an ``ov_tokenizer`` producing ``tokenized_prompt``).
+    """
 
     export_tokenizer: bool = False
     outputs: list[str] = field(default_factory=lambda: ["action"])
     compress_to_fp16: bool = False
     via_onnx: bool = False
+    input_name_map: dict[str, str] = field(default_factory=dict)
```
daniil-lyakhov: @sovrasov, should input_name mapping be a post hook there?
It would allow to remove additional parameter but will require additional save/load of the IR.
What do you think?

library/configs/physicalai/xr0.yaml
```diff
@@ -0,0 +1,75 @@
...
+    # Input / output structure
+    n_obs_steps: 1
+    chunk_size: 30
+    n_action_steps: 30
+    max_state_dim: 32
+    max_action_dim: 32
+
+    # Ordered camera view names embedded into the prompt
+    camera_views:
+      - base
```
sovrasov: Do we need some sort of reorder mechanism in case if dataset has inverted order of views, or those just names?

library/configs/physicalai/xr0.yaml
```diff
@@ -0,0 +1,75 @@
...
+
+data:
+  class_path: physicalai.data.lerobot.LeRobotDataModule
+  init_args:
+    repo_id: "lerobot/aloha_sim_transfer_cube_human"
+    train_batch_size: 16
+    data_format: "physicalai"
+
+trainer:
+  max_epochs: 100
```
sovrasov, line 68: PAS UI targets roughly 8 epochs for fine-tuning. Does this policy really require that many steps?

library/src/physicalai/export/backends.py
sovrasov: Why do we need to know this map in advance?
Ideally, the input names are picked from the input sample, and we don't need to rename them.
OV can arbitrary mutate input names, but in that case we can not know the resulting names in advance

library/configs/physicalai/xr0.yaml
```diff
@@ -0,0 +1,75 @@
...
+      - 256
+
+    # Flow matching parameters
+    num_inference_steps: 5
+
+    # Finetuning settings
+    gradient_checkpointing: true
+    freeze_vision_encoder: false
+    # Frequency-domain auxiliary loss (enabled in the original recipe)
+    enable_freq: true
```
Daankrol, line 48: i see that not freezing the VLM will result in a rather huge VRAM requirement. Have you experimented with finetuning the model while freezing VLM?

library/configs/physicalai/xr0.yaml
```diff
@@ -0,0 +1,75 @@
...
+
+    # Optimizer
+    optimizer_lr: 1.0e-4
+    optimizer_weight_decay: 0.1
+    optimizer_grad_clip_norm: 1.0
+
+    # Scheduler
+    scheduler_warmup_steps: 2000
+    scheduler_decay_steps: 30000
```
Daankrol, line 57: Starting LR seems rather high when comparing to the schuler decay lr. Decay steps don't seem to match the full training loop. Also 100 epochs seems a rather big. Other models use around 5-8 epochs. Setting decay_steps to None might also be an option.

library/src/physicalai/export/mixin_policy.py
```diff
@@ -881,3 +1011,26 @@ def _postprocess_openvino_model(ov_model: openvino.Model, output_names: list[str
...
+            ignored so partial maps are safe.
+    """
+    if not name_map:
+        return
+    for old_name, new_name in name_map.items():
+        try:
+            port = ov_model.input(old_name)
+        except RuntimeError:
+            continue
+        port.get_tensor().set_names({new_name})
```
Daankrol: This Mixin is getting quiet big. Maybe we can move some helpers to a utils file

library/src/physicalai/policies/xr0/config.py
```diff
@@ -0,0 +1,197 @@
...
+    Attributes:
+        vlm_model_id: HuggingFace id of the Qwen3-VL backbone. Defaults to
+            ``"Qwen/Qwen3-VL-4B-Instruct"``.
+        vlm_attn_implementation: Attention backend for the VLM. Defaults to
+            ``"flash_attention_2"``.
+        dtype: Precision for model weights. Options: ``"bfloat16"``,
+            ``"float16"``, ``"float32"``. Defaults to ``"bfloat16"``. Use
+            ``"float16"`` when exporting to OpenVINO for Intel GPU: the GPU
+            plugin cannot build the OpenCL kernel for bf16 attention permutes,
+            while the f16 kernel builds correctly.
```
Daankrol, line 33: do we want to drop some tips in the logger to warn the user?

library/src/physicalai/policies/xr0/config.py
```diff
@@ -0,0 +1,197 @@
...
+        training_repeat: Per-sample training repeat factor. Defaults to 4.
+        enable_freq: Add the frequency-domain loss term. Defaults to True.
+        prefix_mask_prob: Probability of masking a prefix token during training.
+            Defaults to 0.5.
+        async_train: Randomly condition on an action prefix during training.
+            Defaults to False.
+        image_resolution: Target image resolution (height, width). Defaults to
+            (256, 256).
+        camera_views: Ordered camera view names embedded into the prompt.
+            Defaults to ("base", "wrist_left").
```
Daankrol: do we need a rename/remap method when users use different camera names/order?

library/pyproject.toml
```diff
@@ -116,6 +116,14 @@ rldx1 = [
 # Neural Network Compression Framework (optional, for post-export quantization hooks)
 nncf = ["nncf>=3.0.0"]
 
+# Xiaomi-Robotics-0 (XR0) foundation model
+# Requires a Qwen3-VL backbone (transformers>=5.5.0 ships the qwen3_vl models)
+xr0 = [
+    "transformers>=5.5.0,<5.6.0",
```
Daankrol: maybe we can move transformers to the base requirements, seems like we are gonna need this for most models.

Daankrol (commented):
Thats a big one! Looking good! 

Can you elaborate on the statement "model training runs but convergence is wip" ?

daniil-lyakhov:
@sovrasov @Daankrol, thanks for the comments!

The view names parameters was actually a symptom of a bigger problem - the XR0 exported model had the system prompt and some parts of the image preprocessing baked to the exported graph - introducing a lot of code and confusion without visible benefits. 

I've aligned the studio and runtime preprocessors, removing the preprocessed system prompt and a part of the preprocessor from the exported graph, should be much cleaner now

library/src/physicalai/policies/xr0/config.py
daniil-lyakhov: Good catch, please see https://github.com/open-edge-platform/physical-ai-studio/pull/1066#issuecomment-5603492247

library/configs/physicalai/xr0.yaml
daniil-lyakhov: Good catch, please see https://github.com/open-edge-platform/physical-ai-studio/pull/1066#issuecomment-5603492247

library/src/physicalai/export/mixin_policy.py
daniil-lyakhov: Turns out this function could be just dropped, removed the renaming logic by patching forward input names during the export

library/src/physicalai/export/backends.py
daniil-lyakhov: Good catch, I've removed the IR renaming logic by replacing it with an forward input parameters mapping during the export

library/src/physicalai/policies/xr0/config.py
daniil-lyakhov: Actially, it is allright to export in bf16, the only limitation we have - we should specify precision hint f16/f32 specifically for intel GPU

library/configs/physicalai/xr0.yaml
daniil-lyakhov: The config is removed as it was invalid

library/configs/physicalai/xr0.yaml
daniil-lyakhov: The config is removed as it was invalid, didn't try freezing VLM during the training yet

library/configs/physicalai/xr0.yaml
daniil-lyakhov: The config is removed as it was invalid

library/pyproject.toml
sovrasov: I think this should be done in a standalone PR after a brief discussion.
In some cases we don't need transformers (like ACT training or dataset management), the question is whether we prioritize them

library/src/physicalai/data/lerobot/datamodule.py
```diff
@@ -441,18 +441,20 @@ def _ensure_quantile_stats(
...
         # Propagate computed quantiles to the train adapter
-        for key, feat_stats in full_ds.meta.stats.items():
-            if "q01" in feat_stats and key in train_lerobot_ds.meta.stats:
-                train_lerobot_ds.meta.stats[key]["q01"] = feat_stats["q01"]
-                train_lerobot_ds.meta.stats[key]["q99"] = feat_stats["q99"]
+        train_stats = train_lerobot_ds.meta.stats or {}
+        for key, feat_stats in (full_ds.meta.stats or {}).items():
+            if "q01" in feat_stats and key in train_stats:
+                train_stats[key]["q01"] = feat_stats["q01"]
+                train_stats[key]["q99"] = feat_stats["q99"]
```
daniil-lyakhov, line 448: New dependency requires to update uv.lock, which updates pyrefly. With the latest pyrefly I have to fix hits in this unrelated file

daniil-lyakhov:
@sovrasov @Daankrol @samet-akcay, rebase is successful, ready for the rereview

library/src/physicalai/policies/xr0/config.py
```diff
@@ -0,0 +1,187 @@
...
+    enable_freq: bool = True
+    prefix_mask_prob: float = 0.5
+    async_train: bool = False
+
+    image_resolution: tuple[int, int] = (256, 256)
+    tokenizer_max_length: int = 256
+
+    gradient_checkpointing: bool = True
+    compile_model: bool = False
+    compile_mode: str = "max-autotune"
```
sovrasov: Does that work well? If not tested  yet, `default` looks safer

library/src/physicalai/policies/xr0/config.py
```diff
@@ -0,0 +1,187 @@
...
+    normalization_mode: Literal["MEAN_STD", "QUANTILES"] = "QUANTILES"
+
+    optimizer_lr: float = 1.0e-4
+    optimizer_betas: tuple[float, float] = (0.9, 0.95)
+    optimizer_eps: float = 1e-8
+    optimizer_weight_decay: float = 0.1
+    optimizer_grad_clip_norm: float = 1.0
+
+    scheduler_warmup_steps: int = 2_000
+    scheduler_decay_steps: int | None = 30_000
```
sovrasov: We default this parameter to `None` after the recent changes. Have a look at pi05 implementaiton

library/src/physicalai/policies/xr0/config.py
daniil-lyakhov: Didn't tested yet, put default as the default value

library/src/physicalai/policies/xr0/config.py
daniil-lyakhov: Done

library/src/physicalai/policies/xr0/model.py
```diff
@@ -0,0 +1,690 @@
...
+        return None
+
+    @property
+    def action_delta_indices(self) -> list[int]:
+        """Relative indices of the predicted action chunk.
+
+        Returns:
+            ``list(range(action_len))``.
+        """
+        return list(range(self.action_shape[-2]))
```
sovrasov: Should it be equal to explicit chunk size variable?

library/src/physicalai/policies/xr0/model.py
```diff
@@ -0,0 +1,690 @@
...
+        *,
+        return_loss: bool,
+    ) -> torch.Tensor | dict[str, torch.Tensor] | tuple[torch.Tensor, torch.Tensor]:
+        """VLM encode -> MRoPE continuation -> rectified-flow train / inference.
+
+        Returns:
+            The predicted actions (inference) or the loss dict (training).
+        """
+        prefix_length = batch.pop("prefix_length", 0)
+        seed = batch.pop("seed", None)
```
sovrasov: Ideally, we should not inject any hyperparameters to the batch. What's the reason of having it here?
Are there any alternative ways to pass them?

library/src/physicalai/policies/xr0/model.py
```diff
@@ -0,0 +1,690 @@
...
+                return torch.randn_like(action)
+            return torch.randn(action.shape, dtype=torch.float32, device=action.device).to(action.dtype)
+
+        if seed is None or self.training:
+            return _draw()
+
+        seed_val = int(seed.flatten()[0].item()) if isinstance(seed, torch.Tensor) else int(seed)
+        cpu_rng_state = torch.get_rng_state()
+        gpu_rng_state = torch.cuda.get_rng_state(action.device) if action.is_cuda else None
+        torch.manual_seed(seed_val)
```
sovrasov: Would this affect the global torch scope?

library/src/physicalai/policies/xr0/model.py
```diff
@@ -0,0 +1,690 @@
...
+            position_ids = self._repeat_tensor(position_ids, dim=1)
+            action = self._repeat_tensor(action)
+            prefix = self._repeat_tensor(prefix)
+            action_mask = self._repeat_tensor(action_mask)
+            state_embed = self._repeat_tensor(state_embed)
+            attn_mask = self._repeat_tensor(attn_mask)
+            past_key_values = self._repeat_past_key_values(past_key_values)
+
+        position_embeds = self.rotary_emb(action, position_ids)
+        noise = self._sample_noise(action, seed)
```
sovrasov: Looks like we can not disable noise sampling. Maybe this was the root cause of nondeterministic results on libero? I'd suggest fix the related logic

library/src/physicalai/policies/xr0/model.py
daniil-lyakhov: Chunk size is a part of action_shape, always equal to `action_shape[0]`. If not - the whole model would fail.
I changed the -2 to 0, and suggest to keep it as it is to avoid making dependent variables
https://github.com/open-edge-platform/physical-ai-studio/pull/1066/changes#diff-5a8aca7a17e9cd734bc9e5f85bcfde3c3cb6618c73be976562fafb80ad5d8c4aR414

library/src/physicalai/policies/xr0/model.py
daniil-lyakhov: This is copy of the existing XR0 implementation 
https://github.com/XiaomiRobotics/Xiaomi-Robotics-0/blob/main/xr0/mibot/models/VLA/XR0.py#L721

`prefix_lenght` as a batch key was redundant, removed.
`seed` was used only in tests, removed, tests are refactored

library/src/physicalai/policies/xr0/model.py
daniil-lyakhov: It would, I refactored `_run` to take noise as an input parameter instead of dealing with seed. Now XR0 and Pi05 are similar in this regard

library/src/physicalai/policies/xr0/model.py
daniil-lyakhov: Refactored to have `noise` as a `_run` parameter - same way as for the Pi05

library/src/physicalai/policies/xr0/model.py
sovrasov: Does that mean an api user now has a toggle to disable/enable noise, like in pi05?

library/src/physicalai/policies/xr0/preprocessor.py
```diff
@@ -0,0 +1,679 @@
...
+            tensor = batch[key]
+            if tensor.ndim == _TEMPORAL_IMAGE_NDIM:  # (B, T, C, H, W) -> last frame
+                tensor = tensor[:, -1]
+            per_view.append(tensor)
+
+        batch_size = per_view[0].shape[0]
+        images: list[list[Image.Image]] = []
+        for sample in range(batch_size):
+            sample_images = [
+                resize_image(_to_pil(view[sample]), factor=self.image_factor, max_pixels=self.image_max_pixels)
```
sovrasov: Maybe not for now (training is still not polished), but overall we don't use pil in such pipelines: `torch.funcitonal.interpolate` allows batch processing on GPU, which becomes important when we process big batches during training: now it all falls to CPU

library/src/physicalai/policies/xr0/preprocessor.py
```diff
@@ -0,0 +1,679 @@
...
+# deploy/server.py), e.g. "wrist_left" -> "Left-Wrist" so the prompt reads
+# "# Left-Wrist View". A plain capitalize would wrongly yield "Wrist Left".
+_VIEW_TITLES = {
+    "base": "Base",
+    "wrist_left": "Left-Wrist",
+    "wrist_right": "Right-Wrist",
+}
+
+
+def view_title(view: str) -> str:
```
sovrasov: What's the rule of thumb when we declare internal functions/constants in this file? Are non-underscored entities used outside of this file?

library/src/physicalai/policies/xr0/preprocessor.py
```diff
@@ -0,0 +1,679 @@
...
+            "attention_mask": encoded["attention_mask"].to(device),
+            "pixel_values": encoded["pixel_values"].to(device),
+            "image_grid_thw": encoded["image_grid_thw"].to(device),
+            "state": self._prepare_state(batch, device),
+        }
+
+        if ACTION in batch and batch[ACTION] is not None:
+            action, action_mask = self._prepare_action(batch[ACTION], device, state=batch[STATE])
+            out[ACTION] = action
+            out["action_mask"] = action_mask
```
sovrasov, line 515: Declare it as a constant in this file and utilize in xr0 module?

library/src/physicalai/policies/xr0/model.py
daniil-lyakhov: No, noise always on, only for testing purposes for now

library/tests/unit/policies/xr0/test_policy.py
```diff
@@ -0,0 +1,290 @@
+# Copyright (C) 2026 Intel Corporation
```
sovrasov, line 1: Do unit test consume any significant time, like pi05 tests for instance?

daniil-lyakhov: 16 secs on Intel(R) Core(TM) i9-10980XE CPU @ 3.00GHz

library/src/physicalai/policies/xr0/preprocessor.py
daniil-lyakhov: Renamed all helper functions to start with `_`

library/src/physicalai/policies/xr0/preprocessor.py
daniil-lyakhov: Moved to pytorch

library/src/physicalai/policies/xr0/preprocessor.py
daniil-lyakhov: Don't get it, please give an example snipped

library/src/physicalai/policies/xr0/model.py
```diff
@@ -0,0 +1,679 @@
...
+    ) -> tuple[torch.Tensor, torch.Tensor, torch.Tensor]:
+        """Pop ``action`` / ``action_mask`` / ``state`` from the batch.
+
+        Zero-filled defaults are returned for inference when they are absent.
+
+        Returns:
+            Tuple of ``(action, action_mask, state)``.
+        """
+        device = batch["input_ids"].device
+        if "action" in batch:
```
sovrasov: Please use the global `ACTION` constant here

library/src/physicalai/policies/xr0/preprocessor.py
sovrasov: In this file: `ACTION_MASK` = "action_mask"
In `model.py`: `from .preprocessor import ACTION_MASK`

library/src/physicalai/policies/xr0/model.py
daniil-lyakhov: Good catch, done

library/src/physicalai/policies/xr0/preprocessor.py
daniil-lyakhov: Done

sovrasov (approved):

PR #1068, opened 2026-09-01
feat(snapflow): surface SnapFlow distillation in Studio training and UI

Daankrol:
## Summary

SnapFlow self-distillation (`library/src/physicalai/train/callbacks.py::SnapFlowPhaseCallback`) existed in the library but was never wired into Studio: `application/backend` never enabled it, and nothing in `application/ui` referenced it. This PR wires it end to end: opt-in training, checkpoint handling, export/inference, and overview visibility.

## Backend

- `TrainingJobSpec.snapflow_start_epoch` and payload fields `snapflow_enabled` / `snapflow_distill_epochs` describe a two-phase run: standard flow matching, then SnapFlow distillation for the trailing epochs. The payload validates the epoch budget and checks that the policy actually implements `enable_snapflow()` (Pi0.5, SmolVLA only).
- A run that reaches the distillation phase keeps both checkpoints side by side: `model.ckpt` (flow matching, untouched) and `snapflow-model.ckpt` (distilled). Neither is renamed or deleted.
- `resolve_checkpoint()` is the single place that decides which checkpoint represents a model, preferring the distilled one when present.
- Export now reloads from the resolved checkpoint before running backends, instead of exporting the live end-of-fit policy, which holds the final-epoch weights and not necessarily the best distilled epoch on disk. Studio's inference path consumes exported backend artifacts, so this is what makes exported artifacts (and therefore inference) reflect the distilled model.
- Remote submission omits the new spec field on an ordinary run, so a non-distillation job stays compatible with a trainer image that predates SnapFlow.
- `Model.properties` exposes a typed `snapflow_enabled` field (no migration needed), and `TrainingSummary` reports the distillation budget.
- Resume-from-base-model intentionally keeps reading the plain checkpoint, not the distilled one: SnapFlow permanently freezes the VLM and switches to 1-step sampling once activated, baked into the checkpoint's hparams, so resuming from it would silently carry that into a run that never asked for SnapFlow.

## UI

- Opt-in "SnapFlow distillation" checkbox + distillation-epochs field in the train dialog, shown only for flow-matching policies.
- A SnapFlow badge on both the training-jobs and models overview tables.
- Distillation summary (enabled + epoch count) in the model detail panel.

## Testing

- Backend: full suite passes (`uv run pytest -m "not slow and not integration and not requires_download"`), `ruff check`/`ruff format --check` clean, `pyrefly check` clean.
- UI: `type-check`, `lint`, `cyclic-deps-check`, `format:check`, `test:unit`, and `build` all pass. `openapi-spec.json`/`.d.ts` are regenerated locally but gitignored per `application/ui/.gitignore`, so they aren't part of this diff.

## Not in scope for this PR

- Downloading the two checkpoints separately; both currently ship together in the whole-model ZIP.
- Exposing the SnapFlow loss hyperparameters (alpha/lambda) in the UI. Studio always uses the published defaults.

## Screenshots

image

image

image


application/backend/src/schemas/job.py
```diff
@@ -140,6 +142,72 @@ def resolve_training_limit(self) -> "TrainJobPayloadBase":
...
+
+        if self.policy.lower() not in SNAPFLOW_POLICIES:
+            msg = (
+                f"SnapFlow distillation is not available for policy {self.policy!r}; "
+                f"it requires a flow-matching policy ({sorted(SNAPFLOW_POLICIES)})."
+            )
+            raise ValueError(msg)
+
+        max_epochs = self.max_epochs or _DEFAULT_MAX_EPOCHS
+        if self.snapflow_distill_epochs >= max_epochs:
```
AlbertvanHouten: If snapflow epochs are part of max_epochs we cannot train for more snapflow epochs than normal training epochs. It also seems that if we increase snapflow epochs by 1 then we need to additionally update the max epochs by 1 to ensure training still runs for the same length. Maybe we should considering making these settings completely separate.

Daankrol:
I'm performing an end-to-end test via Studio. Lets wait for that before we merge.

application/backend/src/schemas/job.py
Daankrol: Fixed!

Daankrol:
End to end test is completed. The export is made on the final Snapflow checkpoint (as it should). Performance is good and comparable to library cli training on same dataset.

application/ui/src/features/models/shared/snapflow-badge.tsx
```diff
@@ -0,0 +1,23 @@
...
+/**
+ * Marks a training job or model whose checkpoint is SnapFlow-distilled.
+ *
+ * Renders nothing when it is not, so callers can drop it into a row
+ * unconditionally instead of repeating the guard.
+ */
+export const SnapflowBadge = ({ isEnabled }: { isEnabled: boolean | undefined }) => {
+    if (!isEnabled) {
+        return null;
+    }
```
AlbertvanHouten: Wondering if we need a badge for this. If we start adding a badge for every advanced training option it might be a bit much

application/backend/src/schemas/job.py
```diff
@@ -69,6 +70,7 @@ def validate_index_for_device_type(self) -> "TrainingDevice":
 
 
 _DEFAULT_MAX_EPOCHS = 5
```
AlbertvanHouten: I think we should rename this to `_DEFAULT_TRAIN_EPOCHS ` now that max epochs can be extended with snapflow epochs

RHeckerIntel (approved):

PR #1076, opened 2026-09-02
fix(pi05,rldx1,smolvla): ignore unknown keys in pretrained config.json

### LINKED ISSUE #1069: [Bug]: Failed load pi05 in pi05.policy
### Description

when I tried to fine tune pi05, I met an error; after debug, seems the issue happened when pi05.policy.py load the pi05_base model. It can't parse the json file correctly.

jsonargparse._namespace.NSKeyError: Group 'object' does not accept option 'input_features.observation.images.base_0_rgb.type'

### Steps to Reproduce

Here is the debug code to reproduce the issue
```python
import hashlib
import json
import traceback
from pathlib import Path

from huggingface_hub import hf_hub_download
from physicalai.policies.pi05.policy import Pi05

REPO_ID = "lerobot/pi05_base"

print("=== Environment ===")
print("Pi05 implementation:", Path(__import__("physicalai.policies.pi05.policy", fromlist=["__file__"]).__file__).resolve())

config_path = Path(
    hf_hub_download(
        repo_id=REPO_ID,
        filename="config.json",
        repo_type="model",
    )
)

config = json.loads(config_path.read_text(encoding="utf-8"))

print("=== Pretrained model ===")
print("repo:", REPO_ID)
print("config:", config_path)
print("config sha256:", hashlib.sha256(config_path.read_bytes()).hexdigest())
print("input_features:", json.dumps(config.get("input_features"), indent=2))
print("output_features:", json.dumps(config.get("output_features"), indent=2))

print("\n=== Calling Pi05._from_hf() ===")
print("This is the loader called by Pi05.__init__().")

try:
    # _from_hf() performs the same config loading and parsing as Pi05.__init__().
    # It does not access instance state before the failing config parse, so a
    # bare instance is sufficient for this focused reproduction.
    loader = object.__new__(Pi05)
    loader._from_hf(REPO_ID)
    print("UNEXPECTED: Pi05 configuration was accepted")
except Exception as exc:
    print("\nEXPECTED FAILURE:")
    print(f"{type(exc).__name__}: {exc}")
    traceback.print_exc()
```

### Expected Behavior

load pi05_base successfully.

### Actual Behavior

argparse.ArgumentError: Group 'object' does not accept option 'input_features.observation.images.base_0_rgb.type'

### Component

Library (VLA framework)

### Environment

```markdown
Ubuntu 24.04
I also run the git pull to use the latest physical-ai-studio
I also re-downloaded the pi05_base from huggingface.
```

### Logs / Error Output

```shell

```

### Additional Context

_No response_

Daankrol:
### What

Loading any pretrained Pi05/Rldx1/SmolVLA checkpoint (e.g. `lerobot/pi05_base`) fails with:

```
jsonargparse._namespace.NSKeyError: Group 'object' does not accept option 'input_features.observation.images.base_0_rgb.type'
```

### Why

`_from_hf()` in all three policies calls `<Name>Config.from_dict(hf_config, strict=False)`. Pretrained `lerobot`-format `config.json` files always carry keys with no matching `Config` field (`input_features`, `output_features`, `repo_id`, `license`, `push_to_hub`, ...). `strict=False` is supposed to tolerate that, but doesn't: the runtime's jsonargparse-based `Config.from_dict` rejects unknown keys regardless of `strict`. This is a regression in runtime-owned `physicalai.config` introduced by #1024's dictionary-based-payload rewrite — reported and fixed upstream at [openvinotoolkit/physicalai#251](https://github.com/openvinotoolkit/physicalai/issues/251) / [openvinotoolkit/physicalai#253](https://github.com/openvinotoolkit/physicalai/pull/253).

### Fix

Added `physicalai.policies.utils.pretrained.known_config_fields_only(config_cls, data)`, which filters a raw config dict down to the dataclass's known fields (with debug logging of what's dropped), and used it in `Pi05._from_hf`, `Rldx1._from_hf`, and `SmolVLA._from_hf` instead of relying on `strict=False` alone. This workaround can be removed once Studio's `physicalai` pin picks up the upstream fix (docstring links both).

Also documented the pattern in the `physicalai-train-adding-a-policy` skill so a future policy that loads pretrained weights doesn't hit the same issue.

### Testing

- Verified against the real cached `lerobot/pi05_base` checkpoint: reproduced the exact reported error on `main`, confirmed it loads cleanly with this fix.
- `library/tests/unit/policies/utils/test_pretrained.py` (new): 16 tests parametrized across `Pi05Config`/`Rldx1Config`/`SmolVLAConfig` — drops junk keys, preserves known fields, doesn't mutate input, logs drops, rejects non-dataclasses.
- `library/tests/unit/policies/test_pi05.py`: strengthened the `_from_hf` regression test with the real-world junk-key set and assertions on file-sourced values that differ from defaults (so it can't pass under an over-aggressive filter).
- Full pi05 / checkpoint-loading / smolvla / rldx1 suites: 245 passed.
- `uvx prek run` clean (ruff, ruff format, mypy, Pyrefly, bandit) on all changed files.

Fixes #1069

### DIFF
--- a/library/src/physicalai/policies/pi05/policy.py
+++ b/library/src/physicalai/policies/pi05/policy.py
@@ -30,6 +30,7 @@ from physicalai.export.backends import (
 from physicalai.policies.base import Policy
 from physicalai.policies.mixins import SnapFlowPolicyMixin
 from physicalai.policies.mixins.peft import PeftPolicyMixin
+from physicalai.policies.utils.pretrained import known_config_fields_only
 from physicalai.train.schedulers import cosine_decay_with_warmup_scheduler
 from physicalai.train.utils import reformat_dataset_to_match_policy
 
@@ -544,8 +545,11 @@ class Pi05(PeftPolicyMixin, SnapFlowPolicyMixin, ExportablePolicyMixin, Policy):
             if detected is not None:
                 hf_config["normalization_mode"] = detected
 
-        # strict=False: ignore legacy config.json keys not present in Pi05Config
-        config = Pi05Config.from_dict(hf_config, strict=False)
+        # Drop keys not present on Pi05Config before parsing. Pretrained lerobot
+        # config.json files carry extra fields (e.g. input_features, output_features,
+        # architectures, model_type) that from_dict() cannot silently ignore; see
+        # known_config_fields_only() for details.
+        config = Pi05Config.from_dict(known_config_fields_only(Pi05Config, hf_config))
 
         # --- build dataset_stats from HF artefacts ---
         dataset_stats = _extract_dataset_stats(hf_config, preprocessor_file, preprocessor_dir)
--- a/library/src/physicalai/policies/rldx1/policy.py
+++ b/library/src/physicalai/policies/rldx1/policy.py
@@ -57,6 +57,7 @@ from physicalai.policies.base import Policy
 from physicalai.policies.rldx1.config import Rldx1Config
 from physicalai.policies.rldx1.model import Rldx1Model
 from physicalai.policies.rldx1.pretrained_utils import extract_dataset_stats, retrieve_safetensors_shards
+from physicalai.policies.utils.pretrained import known_config_fields_only
 from physicalai.train.schedulers import cosine_decay_with_warmup_scheduler
 from physicalai.train.utils import reformat_dataset_to_match_policy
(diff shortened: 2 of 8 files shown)


alfieroddan (approved):
Thanks!

PR #1077, opened 2026-09-02
fix: Memory leak in websocket camera

RHeckerIntel:
When recording for a very long time you would hit memory issues with cameras.

In JS, an awaited recursive call isn't tail-call optimized: each level keeps the outer async function suspended, holding its promise, closures, and locals alive until the whole chain unwinds

### DIFF
--- a/application/ui/src/features/cameras/websocket-camera.tsx
+++ b/application/ui/src/features/cameras/websocket-camera.tsx
@@ -23,23 +23,22 @@ const CameraCanvas = ({ camera, width, height }: { camera: SchemaProjectCamera; 
 
         processingRef.current = true;
         try {
-            const bitmap = await createImageBitmap(blobData);
-            const canvas = canvasRef.current;
-            const ctx = canvas?.getContext('2d', { alpha: false });
+            let currentBlob: Blob | null = blobData;
 
-            if (canvas && ctx) {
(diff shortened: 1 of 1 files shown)


application/ui/src/features/cameras/websocket-camera.tsx
```diff
@@ -23,23 +23,22 @@ const CameraCanvas = ({ camera, width, height }: { camera: SchemaProjectCamera;
...
-            const bitmap = await createImageBitmap(blobData);
-            const canvas = canvasRef.current;
-            const ctx = canvas?.getContext('2d', { alpha: false });
+            let currentBlob: Blob | null = blobData;
 
-            if (canvas && ctx) {
-                ctx.drawImage(bitmap, 0, 0, canvas.width, canvas.height);
-                setIsLoading(false);
-            }
+            while (currentBlob) {
```
MarkRedeman, line 28: Perhaps one alternative would be to use `requestanimationframe` here?

MarkRedeman (approved):

application/ui/src/features/cameras/websocket-camera.tsx
RHeckerIntel: After some thinking I dont think the requestAnimationFrame would fit well. It could well exceed the refreshrate of images from the backend. Might aswell just use that as timer.

PR #1078, opened 2026-09-02
feat: Implements hotkey editor in settings page

RHeckerIntel:
Allow for custom hotkeys for the user.
Hotkeys are stored in settings.json.
UI can then useHotkey for a specific key (e.g. recording.start)

Bindings page:
image

Recording page:
image

### DIFF
--- a/application/ui/src/features/hotkeys/key-combo.ts
+++ b/application/ui/src/features/hotkeys/key-combo.ts
@@ -1,0 +1,52 @@
+import { HOTKEY_ACTION_IDS, HOTKEY_ACTIONS, HotkeyActionId } from './hotkey-actions';
+
+const MODIFIER_KEYS = new Set(['Control', 'Alt', 'Shift', 'Meta']);
+
+const DISPLAY_KEY_LABELS: Record<string, string> = {
+    ArrowRight: '→',
+    ArrowLeft: '←',
+    ArrowUp: '↑',
+    ArrowDown: '↓',
+    ' ': 'Space',
+};
+
+const modifierPrefix = (e: Pick<KeyboardEvent, 'ctrlKey' | 'altKey' | 'shiftKey' | 'metaKey'>): string[] => {
+    const modifiers: string[] = [];
+    if (e.ctrlKey) modifiers.push('Ctrl');
+    if (e.altKey) modifiers.push('Alt');
+    if (e.shiftKey) modifiers.push('Shift');
+    if (e.metaKey) modifiers.push('Meta');
+    return modifiers;
+};
+
+// Modifiers come from the boolean event flags, never from `e.key`'s casing -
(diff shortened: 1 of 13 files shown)


Daankrol (approved):
Nice!

application/ui/src/routes/datasets/record/recording-viewer.tsx
```diff
@@ -39,24 +42,20 @@ export const RecordingViewer = () => {
...
-                }
-            }
-        };
+    const { data: settings } = $api.useSuspenseQuery('get', '/api/settings');
+    const bindings = getEffectiveBindings(settings.hotkeys.bindings);
 
-        window.addEventListener('keydown', handleKeyDown);
-        return () => window.removeEventListener('keydown', handleKeyDown);
-    }, [state.is_recording, saveEpisode, startEpisode, discardEpisode, task]);
+    useHotkey(bindings['recording.start_episode'], () => startEpisode.mutate(task), !state.is_recording && task !== '');
```
MarkRedeman, line 48: One thing that might be nice as follow up would be to move this hook into a `KeybindingButton` (or some other name..?)
So then that button would be responsible for calling `useHotKey` can can use its own disabled state and `onPress` for the disabled state and action of the hotkey.
The hotkey could then also show its hotkey combo internally. (or perhaps only show the combo on hover if the hotkey is not the default value, mainly to make it so that a key like "CTRL + SHIFT + S" would not take up too much space?)

application/ui/src/features/hotkeys/key-combo.ts
```diff
@@ -0,0 +1,52 @@
...
+    return modifiers;
+};
+
+// Modifiers come from the boolean event flags, never from `e.key`'s casing -
+// Caps Lock alone also produces an uppercase `e.key` with `shiftKey: false`, so
+// trusting casing would make that indistinguishable from an actual Shift press.
+const normalizeBaseKey = (key: string): string => (key.length === 1 ? key.toUpperCase() : key);
+
+/** Canonical combo string for a keydown event, or null for a bare modifier press. */
+export const keyboardEventToCombo = (e: KeyboardEvent): string | null => {
```
MarkRedeman, line 28: Should we also prevent users from using keybindings that are in use by the browser?
This likely will be tricky to detect (I believe not all browsers do the same), might be good as a follow up.

MarkRedeman (approved):

application/ui/src/features/hotkeys/key-combo.ts
RHeckerIntel: Heh, that would probably be very very obvious (unless you switch os or something) since during fill in you would trigger that event.

MarkRedeman: Yeah agree. I vaguely remember that we did do this check on Geti, but it's something that we can fix if we find it's necessary.

PR #1081, opened 2026-09-02
Max/fix model reload

maxxgx:


maxxgx:
commit included in https://github.com/open-edge-platform/physical-ai-studio/pull/1065

PR #1082, opened 2026-09-02
feat(data): enable pinned memory and persistent workers for train/val dataloaders

Daankrol:
Training and eval-loss validation DataLoaders now pin memory and keep worker processes alive between epochs, instead of doing neither like before. Both settings resolve sensibly instead of being hardcoded.

## Changes
- Added a `pin_memory` parameter to `DataModule` and `LeRobotDataModule`, applied to both the training DataLoader and the eval-loss validation DataLoader. It defaults to `"auto"`, resolved lazily to `torch.accelerator.is_available()`, so memory gets pinned automatically when a CUDA/XPU/etc. accelerator is present, and PyTorch's "pin_memory=True but no accelerator found" warning stays quiet on CPU-only setups (this repo ships a `cpu` extra). You can still override it with an explicit bool.
- Added `persistent_workers` (train, default `True`) and `val_persistent_workers` (validation, default `False`). Both get disabled automatically when `num_workers` resolves to `0`, since `torch.utils.data.DataLoader` requires `num_workers > 0` for persistent workers. Validation defaults to off because it runs far less often than training, so keeping its worker processes alive for the whole run buys little and adds extra long-lived processes. It also sidesteps the documented instability from combining `pin_memory=True`, `persistent_workers=True`, and `reload_dataloaders_every_n_epochs > 0` (pytorch/pytorch#91252).
- `LeRobotDataModule`'s lerobot-format `train_dataloader` override honors all three settings.
- `DataModule.__init__` still accepts positional arguments. A `*` keyword-only marker went in and came back out after review feedback flagged it as a breaking change.

## Why
Validation batches weren't pinning memory, so host-to-GPU transfers during eval-loss validation were slower than they needed to be. Separately, Lightning warns that `persistent_workers=True` should be set on `train_dataloader`/`val_dataloader` to avoid re-spawning workers every epoch; this turns it on for training by default and leaves validation opt-in.

## Testing
Earlier test runs in this PR were done against an XPU-flavored `torch` build that a bare `uv run` had silently installed on this CUDA machine, so the `cv2` import failures reported at the time got mislabeled as pre-existing. Both problems are sorted now, and this run uses a correctly provisioned `cu128` environment (`torch.cuda.is_available()` confirmed `True`):
- `uv sync --frozen --extra tests --extra cu128 && uv run --no-sync pytest tests/unit/data/test_datamodules.py tests/unit/datamodules/test_lerobot.py tests/unit/datamodules/test_val_split.py -q` → 51 passed.
- `tests/unit/eval/test_rollout_metric.py::TestRolloutIntegration::test_lightning_training_integration` fails both with and without this PR's changes (checked against the pre-PR base commit), so it's a pre-existing issue unrelated to this change.
- `uv run --no-sync ruff check` / `ruff format --check` on all changed files: no new lint issues.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -34,6 +34,7 @@ Physical AI Studio is the training-side repo for the Physical AI workflow: colle
 
 ## Contribution Notes
 
+- **Before pushing any commit, run `prek run --all-files` (scoped to the changed area, e.g. `prek run --all-files library/`) and make sure it exits clean.** This is what CI's `code-quality` job actually runs; ad hoc `ruff check`/`ruff format --check`/`mypy` calls are not equivalent and can miss real failures (e.g. hook args like ruff's `--fix`, or rules like `unused-noqa` that only show up when a directive stops being necessary). If `prek` modifies files, re-run it until clean and include those changes in the commit.
 - Use Conventional Commits for PR titles and commits.
 - Sign commits when committing changes.
 - Follow `docs/development/coding-standards.md` for repo-wide coding standards.
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -34,7 +34,6 @@ Physical AI Studio is the training-side repo for the Physical AI workflow: colle
 
 ## Contribution Notes
 
-- **Before pushing any commit, run `prek run --all-files` (scoped to the changed area, e.g. `prek run --all-files library/`) and make sure it exits clean.** This is what CI's `code-quality` job actually runs; ad hoc `ruff check`/`ruff format --check`/`mypy` calls are not equivalent and can miss real failures (e.g. hook args like ruff's `--fix`, or rules like `unused-noqa` that only show up when a directive stops being necessary). If `prek` modifies files, re-run it until clean and include those changes in the commit.
 - Use Conventional Commits for PR titles and commits.
 - Sign commits when committing changes.
 - Follow `docs/development/coding-standards.md` for repo-wide coding standards.
--- a/library/src/physicalai/data/datamodules.py
+++ b/library/src/physicalai/data/datamodules.py
@@ -134,6 +134,7 @@ class DataModule(LightningDataModule):
     def __init__(
         self,
         train_dataset: Dataset,
+        *,
         train_batch_size: int = 16,
         num_workers: int | Literal["auto"] = "auto",
         val_gym: Gym | None = None,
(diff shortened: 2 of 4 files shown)


Daankrol:
Addressed in 7dc6723: dropped the `*` keyword-only marker (it wasn't intentional — I only added it to satisfy a ruff lint rule for the new boolean flags) and suppressed `FBT001`/`FBT002` on `pin_memory`/`persistent_workers` instead. `DataModule` remains constructible positionally as before; added a regression test (`TestDataModuleBackwardCompatibility::test_positional_construction`) to cover this.

Daankrol:
Pushed 30ecd37 revising the defaults based on further review:

- `pin_memory` now defaults to `"auto"` (resolved lazily to `torch.accelerator.is_available()`) instead of a hardcoded `True`, so CPU-only environments don't get a permanent `pin_memory=True but no accelerator found` warning on every DataLoader construction.
- Split `persistent_workers` into a train-side flag (`True` by default) and a new `val_persistent_workers` (`False` by default), since keeping validation's worker processes alive for the whole training run has limited benefit and adds memory overhead disproportionate to how often the val loader is actually used.

Also note: the test results I reported earlier in this PR were run against an environment where `uv run` had silently installed an XPU-flavored torch build on this CUDA machine (a footgun with this repo's fallback-index config), and I mischaracterized the resulting `cv2` import failures as "pre-existing". Re-ran everything on a correctly-provisioned `cu128` env — see the updated PR description for accurate results.

alfieroddan (approved):
Looks good - excited for the speedups.

The only concern after some research is users with low RAM. In general if you're training policies on studio I think we'd expect enough RAM.

But perhaps the smaller sized policies (ACT, SmolVLA etc) may have users with lower RAM.

Could we add a small note in agents.md or the docs explaining when to turn off pin_memory and persistent workers?

Daankrol:
@alfieroddan thats actually a really nice idea. I was wondering how to explain this to users when they get OOM but why not just let the agents.md file handle this. good idea. 

And yeah, in general I think we should expect users to have quite some RAM when doing physical AI on local machine.

Daankrol:
@alfieroddan added a note on this instead of AGENTS.md, since that file is repo-map stuff and users training locally wouldn't read it anyway. Put the RAM tradeoff in `library/docs/explanation/data/datamodules.md` and a one-liner in the `working-with-datasets` skill's debugging section, so it shows up for both users and agents digging into OOM issues.

PR #1083, opened 2026-09-03
feat: expose LoRA/DoRA fine-tuning in the training UI, fix compiled PEFT export

Daankrol:
Adds LoRA/DoRA fine-tuning controls to the backend training API and the model training dialog, following the same pattern used for SnapFlow. Support is limited to Pi0 and Pi0.5, the only policies whose config mixes in `PeftConfigMixin`. Also fixes a crash exporting a compiled, LoRA-adapted policy that surfaced while testing this feature end to end.

## Backend
- `TrainingJobSpec` and `TrainJobPayloadBase` gain `lora_enabled`, `lora_rank`, `lora_alpha`, `lora_dropout`, and `lora_use_dora`, validated against the policies that support PEFT.
- `build_policy` forwards these to the policy constructor and bumps the learning rate about 10x for LoRA runs, matching the existing `pi05_lora.yaml` example config.
- `Model.properties` records whether a model was LoRA/DoRA fine-tuned, surfaced through computed fields and `TrainingSummary`.
- The remote training backend omits the new fields from the request body when LoRA is off, so older trainer deployments still accept the payload.

## UI
- A LoRA fine-tuning section (rank, alpha, dropout, DoRA toggle) in the training dialog's advanced settings, shown only for supported policies.
- Job rows, model rows, and the model detail page show a LoRA/DoRA badge and the relevant hyperparameters.

## Library: fix compiled PEFT export

Exporting a LoRA-adapted Pi05/Pi0 policy from the backend crashed with `KeyError: 'action'` whenever `compile_model` was enabled. `PeftPolicyMixin.export()` traced a `copy.deepcopy()` of the model with the adapters merged in, but `copy` treats plain functions as atomic: policies that rebind a compiled method on the instance (`self.forward = torch.compile(self.forward)`) got a copy whose `forward` still dispatched into the original, still-training, still-unmerged module, so `eval()` on the copy never took effect.

- `ExportablePolicyMixin` gains `_export_ready_model()`, a context manager that strips instance-level `torch.compile` wrappers and forces `eval()` for the duration of tracing, restoring both afterward. Wired into `to_onnx`, `to_openvino`, and `to_executorch`.
- `PeftPolicyMixin.export()` now uses a new `merged_lora_scope()` helper that merges adapters into the live model in place (snapshotting only the targeted layers to CPU) instead of deep-copying the whole model, then restores the tuner wrappers and exact weights on exit. This also avoids doubling a multi-billion-parameter model on the accelerator during export.

## Testing
Added backend unit tests for spec/payload validation, learning-rate scaling, and the remote-backend field exclusion, plus UI tests for dialog visibility and badge rendering. For the export fix: new coverage for `_export_ready_model`, `merged_lora_scope`, and an end-to-end regression exercising `PeftPolicyMixin` + `ExportablePolicyMixin` with a compiled, training-mode, LoRA-injected model through real ONNX export.

- Backend: 1185 tests pass, ruff check/format clean, mypy clean.
- UI: 205 tests pass, lint/type-check/format clean, production build succeeds.
- Library: 981 unit tests pass (1 unrelated pre-existing failure confirmed present on main), 2 slow LoRA/DoRA export integration tests pass, `prek` clean.

## Screenshots 

image

image

image

image

image

image


Daankrol:
Lets wait with merging until end to end test is done

Daankrol:
End to end test runs succesfully

application/ui/src/features/models/train-model-dialog/training-parameters.tsx
```diff
@@ -170,5 +193,101 @@ export const TrainingParameters = ({
...
+                                    <Heading>LoRA alpha</Heading>
+                                    <Content>
+                                        <Text>
+                                            Scaling numerator (scaling = alpha / rank). Leave empty to default to the
+                                            rank (scaling = 1.0). Increase for a stronger adaptation signal.
+                                        </Text>
+                                    </Content>
+                                </ContextualHelp>
+                            }
+                        />
```
RHeckerIntel: Confusing default of undefined. Would it be possible to just use the default number? Else a toggle and a number for convenience perhaps.

library/src/physicalai/policies/pi0/policy.py
```diff
@@ -403,16 +407,32 @@ def validation_step(self, batch: Gym, batch_idx: int) -> dict[str, float]:  # ty
...
+
         Returns:
             Optimizer configuration dict.
         """
         # Get trainable parameters
         params = [p for p in self.parameters() if p.requires_grad]
 
+        lr_multiplier = self.config.lora_lr_multiplier
+        learning_rate = self.config.learning_rate * lr_multiplier
+        if self.config.lora_enabled:
```
sovrasov, line 423: Why not applying the multiplier only in case if lora is enabled? Corresponding branching is here anyway.
Having an extra property with related logic embedded doesn't look transparent: it looks exactly the same as other fields of config, but actually is conditioned on the other config fields.

library/src/physicalai/export/mixin_policy.py
```diff
@@ -397,13 +433,13 @@ def to_onnx(
...
             arg_name = self._get_forward_arg_name()
 
-            self.model.eval()
-            self._onnx_core_export_step(
-                model_path=model_path,
-                input_sample=input_sample,
-                arg_name=arg_name,
-                **extra_export_kwargs,
-            )
+            with self._export_ready_model():
```
sovrasov: Maybe a nitpick, but why don't use the manager as a decorator on related methods?
That'd reduce already messy indents

application/backend/src/services/training_backends/remote.py
```diff
@@ -343,8 +343,15 @@ async def submit_job(self, context: TrainingContext) -> uuid.UUID:
...
+
+        # Older remote trainers may run a TrainingJobSpec (extra="forbid") predating the
+        # lora_* fields; only send them when a LoRA run was actually requested, so an
+        # ordinary run's body still validates against an old trainer.
+        excluded = {"run_options"}
+        if not spec.lora_enabled:
+            excluded.update({"lora_enabled", "lora_rank", "lora_alpha", "lora_dropout", "lora_use_dora"})
         body: dict[str, Any] = {
-            "spec": spec.model_dump(mode="json", exclude={"run_options"}),
+            "spec": spec.model_dump(mode="json", exclude=excluded),
```
AlbertvanHouten: Lets exclude this backward compatibility logic since we haven't officially released yet.

application/backend/src/schemas/job.py
```diff
@@ -142,6 +142,40 @@ def resolve_training_limit(self) -> "TrainJobPayloadBase":
         description="Training precision ('32-true', 'bf16-mixed')",
     )
     compile_model: bool = Field(default=False, description="Enable torch.compile for supported policies")
+    lora_enabled: bool = Field(
```
MarkRedeman, line 145: Perhaps for later we can have a `LoraConfiguration` dataclass that looks like,
```python
@dataclass(frozen=True)
class LoraConfig:
    rank: int = Field(
        default=32,
        ge=1,
        le=256,
        description="LoRA rank (dimension of the low-rank decomposition). Ignored unless lora_enabled.",
    )
    alpha: int | None = Field(
        default=None,
        ge=1,
        description=(
            "LoRA scaling numerator (scaling = lora_alpha / lora_rank). None defaults to lora_rank "
            "(scaling = 1.0). Ignored unless lora_enabled."
        ),
    )
    dropout: float = Field(
        default=0.05,
        ge=0.0,
        lt=1.0,
        description="Dropout probability applied to LoRA adapter inputs. Ignored unless lora_enabled.",
    )
    use_dora: bool = Field(
        default=False,
        description=(
            "Use DoRA (Weight-Decomposed Low-Rank Adaptation) instead of plain LoRA; typically improves "
            "quality at low ranks at the cost of slightly more compute/memory. Ignored unless lora_enabled."
        ),
    )
```

and reuse this same structure for the `TrainingSummary` schema.

In the job we can then have,
```python
class TrainJobPayloadBase(BaseModel):
    lora: LoraConfig | None = Field(
        default=None, description=(
        "Fine-tune with LoRA/DoRA instead of full fine-tuning: freezes the base model and trains "
        f"small low-rank adapters, using much less memory. Only available for {sorted(PEFT_POLICIES)}."
        )
    )
```

so that enabling / disabling lora is inferred by the `lora` field not being None

application/backend/src/services/dataset_import_dir_service.py
MarkRedeman, line 1: These changes don't seem relavant to this PR.

It does seem useful so let's keep it in, but next time it would be better to extract it to a separate PR.

application/ui/src/features/models/model-details/training-configuration.tsx
```diff
@@ -32,6 +32,21 @@ export const TrainingParameters = ({ summary }: { summary: SchemaModelDetailResp
                     <DetailRow name='Validation split' value={summary.val_split} />
                 )}
                 {summary.device_type && <DetailRow name='Device' value={summary.device_type} />}
+                {summary.lora_enabled !== null && summary.lora_enabled !== undefined && (
```
MarkRedeman, line 35: These could all be encapsulated in a single` summary.lora_enabled` (or with the dataclass suggestion from my other comment `summary.lora !== null`

application/ui/src/features/models/train-model-dialog/train-model-dialog.tsx
```diff
@@ -80,9 +81,16 @@ export const TrainModelDialog = ({ baseModel, close, defaultMaxEpochs = 5 }: Tra
     const [autoScaleBatchSize, setAutoScaleBatchSize] = useState<boolean>(bestDevice?.type === 'cuda');
     const [precision, setPrecision] = useState<Key | null>(bestDevice?.type === 'cuda' ? 'bf16-mixed' : '32-true');
     const [compileModel, setCompileModel] = useState<boolean>(false);
+    const [loraEnabled, setLoraEnabled] = useState<boolean>(false);
```
MarkRedeman, line 84: Similar to a comment I made on Albert's PR recently, instead of having many different `useState`s its often a bit nicer to combine them into a single `useState` and possibly extract to a named hook:
```
const useLoraState = () => {
    const [loraEnabled, setLoraEnabled] = useState<null | {
        rank: number,
        alpha: number | null,
        dropout: number,
        useDora: boolean
    }>(null);
    
    return [lora, setLora]
}

// in your component
// ...
const [lora, setLora] = useLoraState()
```

I feel like AI wants to add too many `useState`s in components which can blow up quite quickly and make it hard to read in the future.

application/backend/src/services/dataset_import_dir_service.py
Daankrol: agreed, sorry for adding it here.

application/backend/src/schemas/job.py
Daankrol: I agree, and these are not policy specific as far as I can see. Good for follow up PR.

application/ui/src/features/models/train-model-dialog/train-model-dialog.tsx
Daankrol: This is where my limited UI knowledge stops though. Could we add this to some skill?

samet-akcay (approved):
lgtm, but would be better to get an expert review from application perspective

PR #1094, opened 2026-09-07
fix(ui): Improve navigation and robot loading after robot deletion

dwesolow:
## Description

Before when the user would remove a robot that they've selected the route would remain the same, refreshing the page would result in a global "Something was wrong" error without option to select a different robot etc.
Now we don't allow user to open route that has a robot that no longer exists.

After this PR is merged we can close https://github.com/open-edge-platform/physical-ai-studio/pull/1053.

### DIFF
--- a/application/ui/src/router.tsx
+++ b/application/ui/src/router.tsx
@@ -279,6 +279,33 @@ export const router = createBrowserRouter([
                                     },
                                     {
                                         path: paths.project.robots.show.pattern,
+                                        loader: async ({ params }) => {
+                                            const { robot_id, project_id } = params;
+
+                                            if (project_id === undefined || robot_id === undefined) {
+                                                return redirect(paths.projects.index({}));
+                                            }
+
+                                            const { error } = await fetchClient.GET(
+                                                '/api/projects/{project_id}/robots/{robot_id}',
+                                                {
+                                                    params: {
+                                                        path: {
+                                                            robot_id,
+                                                            project_id,
+                                                        },
+                                                    },
+                                                }
+                                            );
+
+                                            if (
+                                                error !== undefined &&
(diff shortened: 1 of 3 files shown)


application/ui/src/router.tsx
```diff
@@ -279,6 +279,33 @@ export const router = createBrowserRouter([
                                     },
                                     {
                                         path: paths.project.robots.show.pattern,
+                                        loader: async ({ params }) => {
```
camiloHimura, line 284: no, move this to a loader utility file so it can be tested in isolation, and keep the router file as a short, general configuration file.

camiloHimura (dismissed):

application/ui/src/router.tsx
```diff
@@ -279,6 +279,33 @@ export const router = createBrowserRouter([
...
+                                            const { robot_id, project_id } = params;
+
+                                            if (project_id === undefined || robot_id === undefined) {
+                                                return redirect(paths.projects.index({}));
+                                            }
+
+                                            const { error } = await fetchClient.GET(
```
MarkRedeman: One minor problem with this is that we'd load the get robot API once here, but then forget about it and reload it via `$api.useQuery`.

Perhaps instead you can use `$api.queryOptions`,
```
                                    // Note: throws on error
                                    const data = await queryClient.ensureQueryData(
                                        $api.queryOptions('get', '/api/projects/{project_id}/robots/{robot_id}', {
                                            params: { path: { robot_id, project_id, }, }
                                        })
                                    )
```

MarkRedeman (approved):

application/ui/src/router.tsx
dwesolow: Updated

PR #1095, opened 2026-09-07
fix: Reset selected episode when dataset is changed

### LINKED ISSUE #1091: [Bug]: switching dataset but the playback window on the left is not updated
### Description

when switch between datasets, I can see the episode column on the right is updated correctly. I can also see the backend log has GET /api/dataset/<dataset ID>/epiosdes/0 http/1.1" 200.
However the epidos playback window on the left side is not updated. Click the play button, it will still play the first episode of the previous dataset; 
This issue can be workaround by refresh the whole web page.

### Steps to Reproduce

1. click the dataset tab to see the first episode loaded on the playback window on the left
2. click another dataset tab. you will see dataset episode column is updated but playback is not updated

### Expected Behavior

Playback is updated

### Actual Behavior

Still playback the episode from previous dataset

### Component

UI (React frontend)

### Environment

```markdown

```

### Logs / Error Output

```shell

```

### Additional Context

_No response_

dwesolow:
## Description

When for dataset A episode 10th is selected, we change dataset to B, episode 10th is still selected. This PR fixes that issue. The other issues mentioned in the ticket were fixed in https://github.com/open-edge-platform/physical-ai-studio/pull/1075.

Fix #1091

### DIFF
--- a/application/ui/src/routes/datasets/dataset-viewer.tsx
+++ b/application/ui/src/routes/datasets/dataset-viewer.tsx
@@ -1,4 +1,4 @@
-import { useMemo, useState } from 'react';
+import { useMemo, useRef, useState } from 'react';
 
 import {
     ActionButton,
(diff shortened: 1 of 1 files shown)


camiloHimura (approved):

MarkRedeman (approved):
Later: let's look into adding `episode_index` into the route url. I think that might help prevent these bugs as the datasets tabs could directly link to the 0 episode index? (or redirect to it after loading the dataset episodes?)

PR #1101, opened 2026-09-08
feat: show advanced configuration options in robot forms

MarkRedeman:
This PR allows the user to open up the "Advanced configuration" options for a robot. These are configuration options that users are unlikely to need unless they're debugging or want more control on their robot.
One important distinction that this PR makes is that the schema form can render form fields inside of sections, which are used if for instance a payload references another payload,
```python
class TrossenBimanualPayload(BaseModel):
    """Connection configuration for Trossen bimanual robots."""

    left_arm: TrossenSingleArmPayload
    right_arm: TrossenSingleArmPayload
```

In this case if these payload contain any advanced options then we show a disclosure panel inside of the section.
In other cases (no section) we simply show the panel at the bottom of the form.

| **Advanced options for rebot** | **Multiple sections** |
|------------|-----------|
|  image  |  image  |

### DIFF
--- a/application/ui/src/features/robots/robot-form/robot-schema/schema-form.module.css
+++ b/application/ui/src/features/robots/robot-form/robot-schema/schema-form.module.css
@@ -1,0 +1,20 @@
+.advancedDisclosure {
+    padding: 0;
+    margin: 8px 0;
+    --spectrum-accordion-icon-gap: var(--spectrum-global-dimension-size-100);
+    --spectrum-accordion-item-padding: var(--spectrum-global-dimension-size-50);
+}
+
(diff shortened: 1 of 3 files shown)


application/ui/src/features/robots/robot-form/robot-schema/schema-form.module.css
```diff
@@ -0,0 +1,20 @@
+.advancedDisclosure {
+    padding: 0;
+    margin: 8px 0;
```
dwesolow, line 3: np. spectrum var here and in other classes in this file

dwesolow (approved):

PR #1104, opened 2026-09-08
feat(smolvla): align SnapFlow CD-sample padding exemption with Pi0.5

Daankrol:
SmolVLA's SnapFlow consistency-distillation (CD) branch regresses onto a self-generated teacher velocity instead of the dataset action, so padded action steps carry no bad supervision there. Pi0.5 already exempts these CD samples from the action-padding mask (#919); SmolVLA discarded `cd_idx` there to preserve its pre-SnapFlow masking behavior, which was a deliberate scope decision at the time, not a correctness argument against doing it.

This wires `cd_idx` through `VLAFlowMatching.forward` -> `SmolVLAModel.compute_loss`, mirroring `Pi05Model._flow_matching_loss`. Only affects SnapFlow-enabled training (`snapflow_enabled` defaults to `False`).

Testing: added `test_snapflow_distillation_steps_ignore_the_pad_mask` to `test_smolvla.py`, mirroring the existing Pi0.5 test. Full `tests/unit/policies/` suite (220 tests) passes locally.

### DIFF
--- a/library/src/physicalai/policies/smolvla/model.py
+++ b/library/src/physicalai/policies/smolvla/model.py
@@ -215,13 +215,18 @@ class SmolVLAModel(Model):
         lang_tokens = batch[TOKENIZED_PROMPT]
         lang_masks = batch[TOKENIZED_PROMPT_MASK]
         loss_dict: dict[str, torch.Tensor | float] = {}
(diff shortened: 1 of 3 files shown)


alfieroddan (approved):
Thanks!

PR #1120, opened 2026-09-10
feat(ui): Improve models page to show only running jobs in the page, failed/cancelled models are visible in the dialog

dwesolow:
## Description

image


application/ui/src/features/models/models-page.tsx
```diff
@@ -31,20 +39,30 @@ export const ModelsPage = () => {
...
+                                    All jobs ({jobs.length})
+                                </Button>
+                            ) : undefined
+                        }
+                    />
                 ) : (
                     <Flex direction={'column'} flex={1} gap={'size-300'} minHeight={0}>
-                        <Flex justifyContent={'end'}>
+                        <Flex justifyContent={'space-between'} alignItems={'center'}>
+                            <Button variant='secondary' onPress={() => setAllJobsOpen(true)} isDisabled={!hasJobs}>
```
MarkRedeman: Could we also check here if there are any jobs?

It's possible that the user does have a model but no jobs if they manually removed the completed job.

MarkRedeman (dismissed):

MarkRedeman (approved):

PR #1122, opened 2026-09-10
feat: expose image augmentation in the training UI

Daankrol:
Image augmentations were only reachable by hand-writing about 45 lines of `image_transforms` YAML, so runs started from the GUI never used them. This adds a ready-made pipeline class to the library and an "Augment images" checkbox in the training dialog's Advanced settings that switches it on.

## Changes

**Library.** `DefaultImageAugmentations(RandomChoice)` in `physicalai/transforms/image_transforms.py`, holding the six transforms and ranges from the existing docs table. `n_subset` and `random_order` stay configurable.

**Backend.** `augment_images: bool = False` on `TrainJobPayloadBase` and `TrainingJobSpec`, mapped in `build_spec` and applied at the one `LeRobotDataModule` construction site in `training/job.py`. `build_spec` is shared with the remote backend, so remote runs get it without extra plumbing. Also added to `TrainingSummary` for the model detail page.

**UI.** Checkbox with contextual help next to "Compile model", plus an "Image augmentation" row on the model detail page. The help text is deliberately hedged: it says augmentation *could* help on small or single-setup datasets, that it does not always improve results, and that epochs get slightly slower.

**Custom pipelines through `data.init_args.image_transforms` still work exactly as before.**
## Testing

- Library: 961 passed on `torch 2.11.0+cu128`, including 7 new tests for the pipeline
- Backend: 1234 passed, including coverage that the datamodule receives the transform when the flag is set and `None` when it is not
- UI: 285 passed, with two new tests asserting the checkbox reaches the request body; type-check, lint, format, cyclic-deps and production build all clean
- `prek run --from-ref origin/main --to-ref HEAD` clean

I regenerated the OpenAPI spec with `physicalai-studio gen-api`.

## Screenshots

The training dialog with the new "Augment images" checkbox under Advanced settings, checked and sitting next to "Compile model":

Train dialog with Augment images checked

The contextual help popover for it:

Augment images contextual help popover

### DIFF
--- a/application/ui/src/features/models/train-model-dialog/training-parameters.tsx
+++ b/application/ui/src/features/models/train-model-dialog/training-parameters.tsx
@@ -29,6 +29,8 @@ interface TrainingParametersProps {
     onPrecisionChange: (value: Key | null) => void;
     compileModel: boolean;
     onCompileModelChange: (value: boolean) => void;
+    augmentImages: boolean;
+    onAugmentImagesChange: (value: boolean) => void;
     isAutoScaleBatchDisabled: boolean;
     deviceType: string | undefined;
     /** False for policies without a flow-matching sampler; hides the SnapFlow controls entirely. */
@@ -52,6 +54,8 @@ export const TrainingParameters = ({
     onPrecisionChange,
     compileModel,
     onCompileModelChange,
+    augmentImages,
+    onAugmentImagesChange,
     isAutoScaleBatchDisabled,
     deviceType,
     isSnapflowSupported,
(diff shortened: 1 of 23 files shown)


application/ui/src/features/models/train-model-dialog/training-parameters.tsx
```diff
@@ -169,19 +173,38 @@ export const TrainingParameters = ({
...
+                </Flex>
+                <Flex direction='row' alignItems='center'>
+                    <Checkbox isEmphasized isSelected={augmentImages} onChange={onAugmentImagesChange}>
+                        Augment images
+                    </Checkbox>
+                    <ContextualHelp variant='info'>
+                        <Heading>Augment images</Heading>
+                        <Content>
+                            <Text>
+                                Randomly varies brightness, contrast, saturation, hue, sharpness and small rotations on
```
AlbertvanHouten, line 199: I don't think we should include rotations in this group of augmentations. Would make more sense to have a separate switch for that IMO.

alfieroddan (approved):
Thanks, looks good library side 😄

PR #1124, opened 2026-09-10
feat: Add SSH tunnel port forwarding support for direct trainer URLs

AlbertvanHouten:
This pull request adds support for optional SSH port-forward tunnels to remote trainers, allowing the backend to maintain a standing SSH tunnel for each configured trainer that requires one. This enables secure, local access to remote trainers via a specified local port, and includes full lifecycle management of these tunnels (creation, update, deletion, and startup/shutdown integration). The changes also introduce a new API for users to add SSH host aliases directly via the UI, and ensure that SSH tunnel configuration is only accepted when the SSH feature is enabled.

**Remote Trainer SSH Tunnel Support**

- Added new nullable fields (`ssh_host_alias`, `ssh_remote_port`, `ssh_local_port`) to the `remote_trainers` database table and corresponding schema/models to support SSH tunnel configuration for each trainer. 
- Implemented validation logic in the `RemoteTrainerCreate` schema to ensure tunnel configuration is coherent and only allowed when required fields are present.

**SSH Tunnel Lifecycle Management**

- Introduced a new module, `remote_trainer_tunnel_manager`, which manages the lifecycle of SSH tunnels for trainers, including starting all tunnels at application startup, syncing tunnels on create/update, and stopping tunnels on deletion or shutdown. 
- Updated `RemoteTrainerService` to invoke tunnel sync/stop operations whenever trainers are created, updated, or deleted, and to reject tunnel configuration if the SSH feature is disabled. 

**API Enhancements for SSH Host Aliases**

- Added a new API endpoint to allow users to create SSH host aliases (`/aliases`), enabling configuration of SSH hosts from the UI without manual SSH config editing. The endpoint verifies reachability and cleans up on failure. 

**SSH Tunnel Implementation Improvements**

- Enhanced the `SshTunnel` class to allow specifying a stable local port, ensuring that trainer URLs can reliably point to a fixed port on the loopback interface.

## Screenshots

image

image


application/backend/src/core/lifecycle.py
```diff
@@ -90,10 +93,20 @@ async def lifespan(app: FastAPI) -> AsyncGenerator[None]:
...
     await app.state.robot_manager.find_robots()
 
+    # Open the standing SSH tunnel for every configured direct trainer that
+    # wants one.
+    try:
+        async with get_async_db_session_ctx() as session:
+            remote_trainers = await RemoteTrainerService(session).list_remote_trainers()
+        await remote_trainer_tunnel_manager.start_all(remote_trainers)
+    except Exception:
+        logger.exception("Failed to start configured remote-trainer SSH tunnels")
```
MarkRedeman: This might take a long time and could make starting up the server a bit slower.
Is this something we could do in the background instead? E.g. start the connection in a background thread (maybe fastapi's background task can help here?) and don't wait for the result.

application/backend/src/schemas/remote_server.py
```diff
@@ -125,6 +125,33 @@ class SshHostAliasOption(BaseModel):
     user: str | None = None
 
 
+class SshHostAliasCreate(BaseModel):
+    """User-supplied fields to append a new ``Host`` stanza to ``~/.ssh/config``.
```
MarkRedeman: TIL:
> A [stanza](https://en.wikipedia.org/wiki/Stanza) is a grouped set of lines in a poem, separated from other groups by a blank line or indentation. [[1](https://www.twinkl.de/teaching-wiki/stanza)] (https://www.twinkl.de/teaching-wiki/stanza)

application/ui/src/features/remote-trainers/remote-trainers-table/remote-trainer-form/remote-trainer-form.tsx
```diff
@@ -51,9 +128,156 @@ export const RemoteTrainerForm = ({ remoteTrainer, close }: RemoteTrainerFormPro
...
+                                if (urlPort !== undefined) {
+                                    setSshRemotePort((current) => current ?? urlPort);
+                                    setSshLocalPort((current) => current ?? urlPort);
+                                }
+                            }}
+                        >
+                            Reach this trainer through an SSH tunnel
+                        </Switch>
+                        {sshTunnelEnabled && (
+                            <Flex direction='column' gap='size-100'>
```
MarkRedeman: Possibly extract this into a smaller component

application/ui/src/features/remote-trainers/remote-trainers-table/remote-trainer-form/remote-trainer-form.tsx
```diff
@@ -13,22 +32,80 @@ type RemoteTrainerFormProps = {
...
+    const [sshTunnelEnabled, setSshTunnelEnabled] = useState(Boolean(remoteTrainer?.ssh_host_alias));
+    const [sshHostSource, setSshHostSource] = useState<SshHostSource>('pick');
+    const [sshHostAlias, setSshHostAlias] = useState(remoteTrainer?.ssh_host_alias ?? '');
+    const [newAlias, setNewAlias] = useState('');
+    const [newHostname, setNewHostname] = useState('');
+    const [newPort, setNewPort] = useState<number | undefined>(22);
+    const [newUser, setNewUser] = useState('');
+    const [newIdentityFile, setNewIdentityFile] = useState('');
+    const [sshRemotePort, setSshRemotePort] = useState<number | undefined>(remoteTrainer?.ssh_remote_port ?? undefined);
+    const [sshLocalPort, setSshLocalPort] = useState<number | undefined>(remoteTrainer?.ssh_local_port ?? undefined);
```
MarkRedeman: Instead of having many different `useState`s I would prefer combining them into a logical group. So for instance,
```tsx
const useSshConfig = (remoteTrainer?: SchemaRemoteTrainer) => {
    return useState<SshConfig>({
        tunnelEnabled: Boolean(remoteTrainer?.ssh_host_alias),
        hostSource: 'pick',
        hostAlias: remoteTrainer?.ssh_host_alias ?? '',
        newAlias: '',
        newHostname: '',
        newPort: 22,
        newUser: '',
        newIdentityFile: '',
        remotePort: remoteTrainer?.ssh_remote_port ?? undefined,
        localPort: remoteTrainer?.ssh_local_port ?? undefined,
    });
}

export const RemoteTrainerForm = ({ remoteTrainer, close }: RemoteTrainerFormProps) => {
    const [name, setName] = useState(remoteTrainer?.name ?? '');
    const [url, setUrl] = useState(remoteTrainer?.url ?? '');
    const [sshConfig, setSshConfig] = useSshConfig(remoteTrainer);
// ...
```

that way code below that needs to update multiple fields can do this at once and you can give a stronger guarantee that all the state should be consistent with each other.

A downside of this approach is that you'd need to do a bit more bookkeeping when changing a single value at a time:
```tsx
                                     <NumberField
                                         label='Remote port'
-                                        value={sshRemotePort}
-                                        onChange={setSshRemotePort}
+                                        value={sshConfig.remotePort}
+                                        onChange={(value) =>
+                                            setSshConfig((current) => ({ ...current, remotePort: value }))
+                                        }
                                         minValue={1}
                                         maxValue={65535}
```

application/ui/src/features/remote-trainers/remote-trainers-table/remote-trainer-form/use-remote-trainer-form-mutation.ts
```diff
@@ -1,38 +1,61 @@
...
+        let resolvedValues = values;
+        if (newSshHost !== undefined) {
+            const result = await createSshHostAlias.mutateAsync({ body: newSshHost }).catch(() => undefined);
+
+            if (result === undefined) {
+                return;
+            }
+
+            resolvedValues = { ...values, ssh_host_alias: result.alias };
+        }
```
MarkRedeman, line 36: Perhaps for a longer term refactor I'd try to change the UX so that the user stores the Ssh host alias **before** creating the remote trainer, so leading to them needing to press a save button twice (or maybe we can do it automatically in the background)
That could then help simplify this part as we wouldn't need to add this `resolvedValues`

MarkRedeman (dismissed):
For UI comments: let's look into this in a follow up steps.

I guess the lifecycle start won't be long if you don't have any ssh remote trainers yet so not a big blocker either.
I had no clue what a stanza was.. maybe rewrite those docblocks.

application/ui/src/features/remote-trainers/remote-trainers-table/remote-trainer-form/use-remote-trainer-form-mutation.ts
AlbertvanHouten: yeah fair point. Will extract this logic in the future.

application/backend/src/schemas/remote_server.py
AlbertvanHouten: renamed to entry

MarkRedeman (approved):

PR #1133, opened 2026-09-14
feat(policies): add RLDX-1 support and runtime callback wiring

eugene123tw:
## Summary
This PR adds first-class RLDX-1 policy support across Studio backend and UI, and updates runtime config construction to attach RLDX-1 callback behavior when loading exported models. It also updates physicalai dependency pins to revision e68aa457 and refreshes lockfiles.

## What changed
- Backend policy registration:
1. Added rldx1 to supported policy class mapping and Hugging Face requirement checks in `policies.py`.
2. Added rldx1 to supported policy validation in `schemas.py`.
3. Added rldx1 to supported import policy set in `model_import_service.py`.

- Training job defaults:
1. Added RLDX-1 pretrained base checkpoint mapping (RLWRLD/RLDX-1-PT) in `job.py`.

- Runtime config behavior:
1. Added manifest-based policy detection via get_policy_name in `config_builder.py`.
2. Added conditional callback injection for rldx1 policy family in `config_builder.py`.
3. Ensures callbacks are only passed when required and available, with graceful fallback behavior.

- Tests:
1. Added coverage for manifest policy-name parsing and conditional callback wiring in `test_config_builder.py`.

- UI:
1. Added RLDX-1 option to training model selection in `policies.ts`.

- Dependency updates:
1. Updated physicalai git source revision to e68aa45740ecc72ec880498dece98bbb9a455529 in:
- `application/backend/pyproject.toml`
- `application/plugin/pyproject.toml`
- `library/pyproject.toml`
2. Regenerated lockfiles:
- `application/backend/uv.lock`
- `application/plugin/uv.lock`
- `library/uv.lock`

## Impact
- Enables RLDX-1 discoverability and selection in Studio workflows.
- Aligns runtime config generation with policy-specific callback needs for RLDX-1 exports.
- Synchronizes Studio with newer physicalai runtime dependency revision.

## Validation
- Added/updated unit tests for runtime callback wiring and manifest parsing in `test_config_builder.py`.


application/backend/src/runtime/config_builder.py
```diff
@@ -109,19 +117,95 @@ def policy_source_from_fragment(fragment: dict[str, Any]) -> PolicySource:
...
+    """Return optional runtime callbacks required by specific policy families."""
+    if policy_name != "rldx1":
+        return []
+
+    try:
+        from physicalai.inference.callbacks.rldx1_vtc import Rldx1VtcWindowCallback
+    except Exception:
+        return []
+
+    return [Rldx1VtcWindowCallback(video_length=4, video_stride=2)]
```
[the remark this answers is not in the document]
eugene123tw: For the reviewer:  Right now the callback is hardcoded to default VTC values, which is why non-default RLDX-1 exports can get the wrong temporal window at inference time.

For the demo, we are keeping the current hardcoded behavior. Immediately after, we will implement the proper fix by persisting `video_length` and `video_stride` in export metadata (via `manifest.json`) and constructing the callback from those exported values, so runtime preprocessing matches the exact settings used during training/export.

application/ui/src/features/models/train-model-dialog/train-model-dialog.tsx
```diff
@@ -126,6 +128,12 @@ export const TrainModelDialog = ({ baseModel, close, defaultMaxEpochs = 5 }: Tra
         }
     }, [activeDevice]);
 
+    useEffect(() => {
```
MarkRedeman: @RHeckerIntel can you later look into refactoring this so that we set this state when changing the selected model?

application/backend/src/training/job.py
```diff
@@ -173,6 +177,17 @@ def validate_snapflow(self) -> TrainingJobSpec:
...
+    @model_validator(mode="after")
+    def validate_compile_support(self) -> TrainingJobSpec:
+        """Reject ``compile_model`` for policies that do not support torch.compile."""
+        if self.compile_model and self.policy.lower() in _COMPILE_UNSUPPORTED_POLICIES:
+            msg = (
+                f"compile_model is not supported for policy {self.policy!r}; "
+                f"disable compile_model or choose a different policy."
+            )
+            raise ValueError(msg)
+        return self
```
Daankrol: It seems that the API accepts an invalid training job before the worker actually fails. validate_compile_support is still only declared on TrainingJobSpec. The API endpoint POST /api/jobs:train validates against TrainJobPayloadBase and persists the job before TrainingJobSpec is constructed. An RLDX-1 job submitted with compile_model: true returns HTTP 201 Created and is enqueued, failing only later in the background worker.

Would suggest moving or adding this check to TrainJobPayloadBase is still needed to reject invalid requests

library/src/physicalai/policies/rldx1/policy.py
```diff
@@ -192,6 +195,7 @@ def __init__(  # noqa: PLR0913
         # Precision / compilation
         use_bf16: bool = True,
         gradient_checkpointing: bool = True,
+        compile_model: bool = False,
```
Daankrol, line 198: would suggest to change line 313: 
        self.save_hyperparameters(ignore=["config"]) 

to
            self.save_hyperparameters(ignore=["config", "compile_model"])

this is inline with the other policies

library/tests/unit/export/test_manifest_callbacks.py
```diff
@@ -0,0 +1,99 @@
...
+import pytest
+import torch
+
+from physicalai.export.backends import ONNXExportParameters, OpenVINOExportParameters
+from physicalai.export.mixin_policy import ExportBackend, ExportablePolicyMixin
+from physicalai.inference.manifest import ComponentSpec, Manifest, ModelSpec
+from physicalai.policies.rldx1.export import Rldx1ExportMixin
+
+
+_CALLBACKS_SCHEMA_SKIP_REASON = "Runtime ModelSpec.callbacks is unavailable; waiting for the separate Runtime dependency update"
```
Daankrol, line 17: Tracking this with a GH Issue might be nice

Daankrol (commented):
Nice PR! I do see that this model does not use the PEFTMixin and thus only exposes full finetuning in Studio. Might be a good follow up PR.

application/ui/src/features/models/train-model-dialog/policies.ts
```diff
@@ -37,4 +37,10 @@ export const MODELS: ReadonlyArray<{
         description: 'Enhanced Pi0 with discrete state encoding and longer context',
         minVRAM: 40 * GB,
     },
+    {
+        id: 'rldx1',
+        name: 'RLDX-1',
+        description: 'General-purpose robot foundation model for dexterous manipulation',
```
AlexanderBarabanov, line 43: Please add license info here (we can revisit it later and improve the layout for all policies if needed)

```suggestion
        description: 'General-purpose robot foundation model for dexterous manipulation (RLWRLD Model License v1.0)',
```

application/backend/src/training/job.py
eugene123tw: Removed this part.

library/src/physicalai/policies/rldx1/policy.py
eugene123tw: Added

library/tests/unit/export/test_manifest_callbacks.py
eugene123tw: Removed as callback support is added to runtime.

AlexanderBarabanov (approved):

PR #1140, opened 2026-09-15
feat: Extend training options

RHeckerIntel:
Implements #895 

The Training setup was getting longer with more options, whilst not having enough parameters available. This adds multi-step wizard for training a model. 

### Camera mapping

SmolVLA has a strong preference for overview first, second the wrist camera and then auxiliary. The second screen in the wizard allows you to select the camera mapping. This is passed onto the trainer and the resulting model will assume the correct names for inference. This is only enable for SmolVLA for now.

### Batch size and auto

Batch size is kept as power of two slider. Secondly auto batch size algorithm is disabled by default. this algorithm can kill the process (in case of memory issues) or crash the system entirely.

### Export specific backends

Last step of the wizard allows you to select the backends you want to export. No quantization support. That'll be next later.

#### Screenshots

image

image

image

image


Daankrol:
Nice, I would prefer to merge https://github.com/open-edge-platform/physical-ai-studio/pull/1083 first

ArendJanKramer (approved):

PR #1143, opened 2026-09-15
fix(act): restore reference optimizer defaults and fix single-camera normalization

### LINKED ISSUE #998: [Bug]: ACT training defaults cause policy collapse to constant output
### Description

ACT policies trained with the native physicalai ACT implementation collapse to a constant, observation-invariant action chunk: on a real SO-101, the deployed policy commands the same pose regardless of camera input. Root cause is three compounding defaults in library/src/physicalai/policies/act/, verified in source and reproduced with a controlled A/B. A minimal patch restoring reference values is attached and validated end-to-end (training → OpenVINO export → real-arm rollout).

### Steps to Reproduce

headless training via run_training_job():

Environment: PAIS library checkout, uv sync --extra cu128 (CUDA build; the collapse also occurs in fp32). Minimal repro script, reconciled against the invocation that produced the failing model:

Dataset (private on HF, can share if helpful)

LeRobot v3 SO-101 teleoperation dataset: 35 episodes, 4,189 frames total, recorded at 30 FPS — averaging ~120 frames (~4 seconds) per episode. Note this means the default chunk_size=100 spans roughly 85% of a typical episode. Single pick-and-hold manipulation task with one task string. Three camera streams per frame at 1536×1920 (float32, CHW). 6-DoF state and action in the SO-101 normalized unit convention. Limited object-position diversity (single-position regime).

Dataset health was independently verified: states, actions, and images are correct and time-varying when loaded through the PAIS reformatting path. The same dataset trains a working policy once the attached patch is applied, so the dataset is not a cause of the failure. Its short-episode regime does make the collapse fastest and most total (heavy chunk padding, strong mean-trajectory attractor); the underlying mechanism is dataset-independent, but severity on longer-episode datasets may differ.

```python
"""Minimal reproduction of the ACT policy collapse via the shared training runner.

Run with the Studio backend src importable:
    PYTHONPATH=<pais>/application/backend/src python repro_train.py

Everything not set below is a code default -- TrainingJobSpec (extra="forbid")
exposes no field that could override the defective hyperparameters
(optimizer_lr, backbone LR, grad clip).
"""

from pathlib import Path

from training import TrainingJobSpec, run_training_job

WORK = Path.home() / "pais_act_repro"
DATASET = WORK / "dataset"
OUTPUT = WORK / "model"
CACHE = WORK / ".model.training-cache"

# Point DATASET at a LeRobot v3 SO-101 dataset root (see characteristics in the
# Environment section; ours is available to maintainers privately on request).
# e.g.: huggingface_hub.snapshot_download("<your-so101-dataset>",
#           repo_type="dataset", local_dir=DATASET)

spec = TrainingJobSpec(
    policy="act",
    policy_source="physicalai",   # native implementation (NOT "lerobot")
    max_epochs=25,                # 3 epochs already shows the divergence via probes
    batch_size=4,                 # also reproduces at 8
    num_workers="auto",
    val_split=0.1,
    precision="32-true",          # collapse occurs in full fp32; not a numerics issue
    compile_model=False,
    auto_scale_batch_size=False,
    device_type="cuda",
    device_index=0,
)

run_training_job(
    spec,
    dataset_root=DATASET,
    output_dir=OUTPUT,
    cache_dir=CACHE,
    report=lambda progress, message, extra: print(f"[{progress:3d}%] {message} {extra}"),
    should_stop=lambda: False,
    resume_from=None,
)

print(f"Checkpoint: {OUTPUT / 'model.ckpt'} -- run the verification probe against it.")
```
### Expected Behavior

Outputs vary with frame / servo inputs

### Actual Behavior

Outputs identical across frames to ~4e-6 (raw units) — the policy is observation-invariant.
Validation L1 plateaus at ~0.394 vs a computed constant-predictor (dataset-mean action) baseline of ~0.399.
Backbone forensics: pretrained ResNet conv weights drift 7–10% rel-L2; forward activations amplified ~1000x by layer4 vs a fresh ImageNet ResNet.
Deployed on hardware: 900 control steps logged with per-joint action min == max and total variation exactly 0.0.

### Component

Library (VLA framework)

### Environment

```markdown

```

### Logs / Error Output

```shell

```

### Additional Context

Root cause

ACT policies trained with the native physicalai ACT implementation collapse to a constant, observation-invariant action chunk — on a real SO-101, the deployed policy commands the same pose regardless of camera input. Three compounding defaults in library/src/physicalai/policies/act/ are responsible:

optimizer_lr = 1e-4 in config.py — 10x the reference value. LeRobot 0.6.0 uses optimizer_lr = 1e-5 (lerobot/policies/act/configuration_act.py).
No backbone learning-rate treatment: configure_optimizers in policy.py builds a single AdamW over all trainable parameters, so the pretrained ResNet-18 trains at the full 1e-4. Only BatchNorm statistics are frozen (FrozenBatchNorm2d); requires_grad is never touched. LeRobot carries a separate optimizer_lr_backbone = 1e-5 param group; the PAIS config has no such field, and TrainingJobSpec (extra="forbid") exposes no way to override any of this from any invocation path, including the Studio UI — the UI-resolved hparams.yaml contains these same values.
optimizer_grad_clip_norm = 10000 in config.py — clipping is implemented but the threshold is never reached, so it is effectively disabled. LeRobot default: 10.

Mechanism, confirmed by measurement: the unclipped 10x LR drives the pretrained backbone's conv weights to drift (7-10% relative L2 by 26k steps), the drift compounds through the residual stages into ~1000x activation amplification at layer4, and the transformer collapses to emitting the dataset-mean action chunk. Trained checkpoints produce outputs identical to ~4e-6 across widely separated dataset frames, and validation L1 pins at the constant-predictor baseline (always predicting the dataset-mean action). Deployed on hardware, 900 consecutive control steps logged per-joint action min == max with total variation exactly 0.0.

Fix

The attached act_training_fix.patch (applies with git apply -p1 at repo root) makes three changes: optimizer_lr 1e-4 -> 1e-5; a new optimizer_lr_backbone: 1e-5 field with a param-group split on ".backbone." in configure_optimizers; optimizer_grad_clip_norm 10000 -> 10. These restore the values the LeRobot reference implementation uses for the same architecture.

Re-run and validation

With the patch as the only variable, on the same machine and dataset (single RTX A6000, fp32, batch 8):

Controlled A/B at 3 epochs (~1.4k steps each): the unpatched control already shows backbone activation distortion (layer4 std at 0.15x a fresh ImageNet ResNet) and near-constant predicted chunks (0.24 raw units of within-chunk motion); the patched run stays healthy (backbone ~1.0x, within-chunk motion 30.9, cross-frame output sensitivity 90.5 vs the control's 25.9). A 3-epoch run (~15 min) is sufficient to observe the divergence.
200-epoch patched run (~97k steps, ~16 h): train L1 0.010, val 0.261 versus the ~0.399 constant-predictor baseline. Checkpoints saved every 25 epochs were probed individually: cross-frame output sensitivity stays at 74-87 raw units across the entire run with no decay toward the collapse floor.
fp32 OpenVINO export of the patched model matches the PyTorch checkpoint to 4 significant figures on the sensitivity metric; the export of the unpatched model reproduces its collapsed outputs digit-for-digit.
On-robot A/B with identical deployment configuration except the model directory: the unpatched policy holds one fixed pose for the full rollout (action variation 0.0 over 900 steps); the patched policy completes the pick task on the same arm, cameras, and calibration.

[act_training.patch](https://github.com/user-attachments/files/31199304/act_training.patch)

Daankrol:
This PR restores LeRobot reference optimizer parameters for the ACT policy and fixes single-camera visual feature normalization during training. It also defaults visual observations to ImageNet statistics and disables OpenVINO FP16 compression to prevent attention logit overflow on GPU runtimes.

Resolves #998

### Key fixes

1. **Backbone learning rate and clipping:** Restores `optimizer_lr = 1e-5`, adds a dedicated `1e-5` parameter group for the ResNet backbone in `configure_optimizers()`, and sets `optimizer_grad_clip_norm = 10.0`. Previously, training ran at `1e-4` with clipping set to 10000, causing backbone weights to drift and policy outputs to collapse to near-constant actions.
2. **ImageNet normalization default:** Added `use_imagenet_stats: bool = True` in `ACTConfig`. 3-channel visual observations now normalize against ImageNet mean and std instead of small empirical dataset variance, preventing ambient lighting differences from being magnified by 50x and matching the pretrained ResNet-18 input distribution.
3. **Single-camera normalization:** In `ACTModel`, single-camera visual features use the key `images`. Observation dictionaries provide keys like `images.wrist`. Normalization matching now handles both `IMAGES` prefix variants so inputs are not left unnormalized during training.
4. **OpenVINO export stability:** Disabled `compress_to_fp16` for OpenVINO export. ACT transformer attention logits reach magnitudes around 46, which overflow FP16 limits (65,504) on GPU plugins and produce corrupted action chunks.

### Probe measurements (100 synthetic steps)

| Metric | Before patch | After patch |
| :--- | :--- | :--- |
| Optimizer parameter groups | 1 group (lr=1e-4, 153 params) | 2 groups (non-backbone lr=1e-5: 133 params, backbone lr=1e-5: 20 params) |
| Single-camera image normalization | Bypassed (output unchanged) | Applied (standardized with ImageNet stats) |
| ResNet conv1 relative L2 drift | 0.4648% | 0.0611% (7.6x lower drift) |
| Cross-frame action sensitivity | 0.001584 (collapsed) | 0.168331 (106x higher visual responsiveness) |
| Within-chunk trajectory motion | 0.008046 (near static pose) | 0.518730 (64x dynamic action trajectory) |

### Changes

- `library/src/physicalai/policies/act/config.py`: restore `optimizer_lr = 1e-5`, `optimizer_grad_clip_norm = 10.0`, add `optimizer_lr_backbone = 1e-5` and `use_imagenet_stats = True`.
- `library/src/physicalai/policies/act/policy.py`: separate backbone optimizer group, apply ImageNet stats when enabled, and set `compress_to_fp16 = False` for OpenVINO export.
- `library/src/physicalai/policies/utils/normalization.py`: support single-camera `IMAGES` prefixes in normalizer matching.
- `library/configs/physicalai/act.yaml` and `act_xpu.yaml`: align config `optimizer_lr` with reference default.
- `library/tests/unit/policies/test_act.py`: add unit tests for optimizer groups, single-camera normalization, and ImageNet stats overrides.


library/src/physicalai/policies/act/config.py
```diff
@@ -113,8 +113,9 @@ class ACTConfig(Config):
     dropout: float = 0.1
     kl_weight: float = 10.0
 
-    optimizer_lr: float = 1e-4
+    optimizer_lr: float = 1e-5
+    optimizer_lr_backbone: float = 1e-5
     optimizer_weight_decay: float = 1e-4
-    optimizer_grad_clip_norm: float = 10000
+    optimizer_grad_clip_norm: float = 10.0
```
sovrasov, line 126: The original ACT in lerobot doesn't apply clipping. Let's leave it as is now (fixed to 10) and see if that reduces convergence speed later

library/src/physicalai/policies/act/model.py
```diff
@@ -134,7 +134,8 @@ def __init__(  # noqa: PLR0913
         elif len(visual_observation_features) > 1:
             for vf in visual_observation_features:
                 if vf.name is not None:
-                    input_features_filtered[vf.name] = vf
+                    cam_name = vf.name.removeprefix("observation.").removeprefix(f"{IMAGES}.").removeprefix("image.")
```
sovrasov: Is that also related to a single camera normalization bypass?

sovrasov (commented):
Thanks!

library/src/physicalai/policies/act/config.py
Daankrol: No it actually does, also confirmed by a test run in Lerobot 

- In lerobot/src/lerobot/optim/optimizers.py, AdamWConfig sets grad_clip_norm: float = 10.0 by default.
- In lerobot/src/lerobot/policies/act/configuration_act.py, get_optimizer_preset() returns AdamWConfig, which inherits grad_clip_norm = 10.0.
- In lerobot/src/lerobot/scripts/lerobot_train.py (lines 125–126), the training step explicitly executes:
if grad_clip_norm > 0:
    grad_norm = accelerator.clip_grad_norm_(policy.parameters(), grad_clip_norm)
- in the train_config.json:
"optimizer": {
    "type": "adamw",
    "lr": 1e-05,
    "grad_clip_norm": 10.0
}

Daankrol: But if you mean, lets keep it at 10.0 then i do agree with you.

library/src/physicalai/policies/act/model.py
Daankrol: No, line 137 is in the multi-camera branch (len > 1). The single-camera bypass was resolved in normalization.py.

sovrasov (dismissed):

samet-akcay (dismissed):
lgtm, thanks for addressing this

Daankrol:
Did some real life data tests but even with this "fix" the model is still not working well. Lets not merge untill we have actually solved this issue.

library/src/physicalai/policies/act/model.py
sovrasov: When we use datasets prepared by the studio, the `vf.name` is already clean from prefixes.
You can check the cleaning process in `library/src/physicalai/data/lerobot/dataset.py:L164`. It also covers lerobot's-baked dataset

library/src/physicalai/policies/act/policy.py
```diff
@@ -72,7 +75,11 @@ def _remap_lerobot_act_state_dict(
             image_prefix = "normalize_inputs.buffer_observation_images_"
             if key.startswith(image_prefix):
                 suffix = key.split(".")[-1]
-                image_key = f"buffer_{IMAGES}" if single_camera else key[len("normalize_inputs.") : -len(f".{suffix}")]
+                if single_camera:
```
sovrasov: I see that one affects only checkpoints loaded from lerobot: no intersections with PAS trainin

sovrasov (approved):

PR #1154, opened 2026-09-15
fix: parse HF_TOKEN from application/backend/.env

### LINKED ISSUE #1153: fix: HF_TOKEN in application/backend/.env has no effect
### Description

Setting `HF_TOKEN` in `application/backend/.env` (as documented for native installs) has no effect: the Hugging Face access check in the UI still reports no token configured.

`resolve_hf_token()` (`application/backend/src/services/training_backends/local.py`) falls back to `os.environ.get("HF_TOKEN", "")` when no token is set via the Settings page. That reads the real process environment only. Pydantic-settings' dotenv source parses `.env` internally and feeds values straight into the `Settings` model - it never exports them into `os.environ`, and nothing else in the backend calls `load_dotenv()`. So a value that only exists in `.env` is invisible to that fallback.

This works for the Docker deployment only because docker-compose injects `HF_TOKEN` as a real container environment variable, not because of the `.env`-file mechanism itself.

### Steps to Reproduce

1. Run the backend natively (`uv run --extra xpu physicalai-studio serve`).
2. Add `HF_TOKEN=<token>` to `application/backend/.env`.
3. Open a policy (e.g. SmolVLA) in the UI.

**Expected:** the Hugging Face access warning clears.
**Actual:** the UI still reports "no token configured", even with a valid token in `.env`.

### Workaround

Set the token via Settings > General > Hugging Face in the UI instead; that path persists to `settings.json` and is read correctly.

nisyhaal:
## Description

`resolve_hf_token` fell back to `os.environ.get("HF_TOKEN")`, which only sees a real process environment variable - never a value that only exists in `application/backend/.env`, since pydantic-settings' dotenv source feeds `.env` values into the `Settings` model directly without exporting them into `os.environ`. As a result, setting `HF_TOKEN` in `.env` (as the native install docs describe) had no effect.

Added a top-level `Settings.hf_token` field (alias `HF_TOKEN`) so pydantic-settings resolves it from either a real env var or `.env` the same way it already resolves fields like `STORAGE_DIR`, and `resolve_hf_token` now reads that field instead of `os.environ` directly. The settings-page token (`huggingface.hf_token`) still wins whenever both are set - no change to that precedence.

Added tests covering: settings-page token takes precedence, `.env`/env-var fallback is used when unset, and both unset resolve to `None`. Also added coverage for the field itself (real env var, `.env` file, empty-string treated as unset).

## Related Issues

Fixes #1153


AlexanderBarabanov:
@nisyhaal Could you please take a look on [prek failure](https://github.com/open-edge-platform/physical-ai-studio/actions/runs/34968411586/job/105643212594?pr=1154)? Thanks!

AlbertvanHouten:
Thanks for opening this PR. At this moment we do not support setting HF_TOKEN through a .env file anymore and require setting it either through the UI or directly in the settings.json file which the UI reads. You are right to point out the issues in our documentation though, I have opened another PR to fix the docs: https://github.com/open-edge-platform/physical-ai-studio/pull/1184.

PR #1161, opened 2026-09-16
fix: upgrade twine and trivy-action

AlexanderBarabanov:
This is follow-up PR to fix issue similar to https://github.com/open-edge-platform/physical-ai-studio/pull/1141
In addition, Trivy action is bumped to mitigate https://github.com/open-edge-platform/physical-ai-studio/actions/runs/34554240180/job/103123368874


ivanzati (approved):
Thank you, ALexander!

PR #1168, opened 2026-09-17
refactor: Add image build target without CUDA files prohibited for re-distribution

ivanzati:
## Description

- Add image build target without CUDA files prohibited for re-distribution
- Generalize image build workflow to run in a fork


AlexanderBarabanov (approved):
Thank you, Ivan!

